PR #19, opened 2015-04-30
Expose properties as constants and make case consistent with StatusBarIOS

amccloud:


lwansbrough:
This is probably a good change, but because the proposed changes are breaking I think I'll leave them for a major release. I'll merge when the time comes for that. Thanks a lot!

PR #41, opened 2015-05-30
Add flash mode setting

guodong000:
Thanks for creating this project~


qingfeng:
:beer:

lwansbrough:
Thanks for this, there is another PR for similar functionality but I feel as though this fits better with the current API.

PR #47, opened 2015-06-12
Use dummy image in simulator

jtremback:


lwansbrough:
Great work, thanks for this.

PR #48, opened 2015-06-14
added video support for cameraRoll and disk targets

machard:
I am really not familiar with Objective C
Don't hesitate to indicate me changes i need to make.

Currently only support cameraRoll and disk targets

### DIFF
--- a/Camera.ios.js
+++ b/Camera.ios.js
@@ -182,6 +182,10 @@ var Camera = React.createClass({
     }
 
     NativeModules.CameraManager.capture(options, cb);
+  },
+
+  stopCapture() {
+    NativeModules.CameraManager.stopCapture();
   }
 
 });
--- a/RCTCameraManager.h
+++ b/RCTCameraManager.h
@@ -45,6 +45,7 @@ typedef NS_ENUM(NSInteger, RCTCameraFlashMode) {
 @property (nonatomic) AVCaptureSession *session;
 @property (nonatomic) AVCaptureDeviceInput *captureDeviceInput;
 @property (nonatomic) AVCaptureStillImageOutput *stillImageOutput;
+@property (nonatomic) AVCaptureMovieFileOutput *movieFileOutput;
 @property (nonatomic) AVCaptureMetadataOutput *metadataOutput;
 @property (nonatomic) id runtimeErrorHandlingObserver;
 @property (nonatomic) NSInteger presetCamera;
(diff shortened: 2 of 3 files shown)


lwansbrough:
Wow awesome thank you. This is a big change so I'm going to spend some time reviewing this - hopefully we can get it merged asap.

darkrishabh:
Just a request.. If there can be a kind of video editor with some set values or custom values for limiting the size of the video upload. 
eg - you can select a video from camera roll but you can edit it to be of not more than 1 minute. something like that.

I haven't written single line in objective-c but if you need any help in terms of JS and React I am up for it.

Also if it is helpful to you in anyway, I wrote a wrapper for DB Models in React native. would love your feedback. view my repo

eduedix:
@darkrishabh I think what you requested should be definitely a separate component and not included in the react-native-camera.

machard:
Regarding the quality of the capture, a setting could be added to the capture session via `setSessionPreset`. But it seems to be related to both video and picture, it would be better in an other PR

lwansbrough:
@darkrishabh @eduedix Yes that's out of scope for this module, though perhaps a recording event could be passed back periodically from the camera so the progress of the capture can be monitored. It might be possible to get the current bytes captured, and total recording time - I haven't looked into that though.

darkrishabh:
hmm.. I would surely love the video recording. Will try to work on a video editing thing till then.

@machard  it will be great if you can tell me how I will be able to implement the quality part also set recording limit by hard coding. I am not too good with objective C to understand and implement myself. I would really appreciate your help.

rt2zz:
Instead of `capture({mode: 'video'})` I would rather see a seperate `captureVideo` method (or something similar).

Reasoning being: still capture and video capture can happen simulatenously, and also `stopCapturing `only makes sense if the video capture has already started.

eduedix:
+1 @rt2zz

lwansbrough:
The current RC is now apart of master. This isn't on npm yet which gives you guys time to suggest changed.

I'm not sure I want to add `captureVideo` in place of `capture({ mode: 'video' })`. The point of implementing the `capture` method was to have a unified interface when video capture was made available. You can still call `capture` multiple times, by the way (or at least, in theory you can - I haven't tried it.)

Also, currently the module now prompts for microphone access immediately which is obviously a huge problem, so I'll have to figure out how I can initialize the camera module without starting the recording sessions. I think someone else filed a bug regarding the temperature of the phone when the camera module has been imported but not initialized - this likely has to do with the capture sessions being initialized upon module initialization.

PR #70, opened 2015-07-19
In RCTCameraManager.m, encapsulated many of the "dispatch_async" bloc…

achang28:
In RCTCameraManager.m, I encapsulated many of the "dispatch_async" blocks within another dispatch_async block, along with setting the dispatch queue priority to high.

Made a minor tweak within Camera.ios.js -- wondering if there is a difference with this change over the previous version.

Personally I noticed that the number of "15-second UI locks" significantly reduced from EVERY time Camera is rendered vs. 1 out of every 4 times.  I am running React Native 0.7.1 on Iphone 5s, IOS 7.13

### DIFF
--- a/Camera.ios.js
+++ b/Camera.ios.js
@@ -82,10 +82,12 @@ var Camera = React.createClass({
   },
 
   componentWillMount() {
-    NativeModules.CameraManager.checkDeviceAuthorizationStatus((function(err, isAuthorized) {
-      this.state.isAuthorized = isAuthorized;
-      this.setState(this.state);
-    }).bind(this));
+    NativeModules.CameraManager.checkDeviceAuthorizationStatus((err, isAuthorized) => {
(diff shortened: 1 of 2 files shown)


lwansbrough:
Thanks for this, I'll try it out in the morning.

lwansbrough:
`dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0)` - is this a serial queue? I ask be cause if we encapsulate calls to a serial queue inside this queue, we need to make sure that this one is serial to ensure events happen in order.

lwansbrough:
According to the information I can find in the Apple docs + on SO, this queue is concurrent, which will create race conditions.

lwansbrough:
Ok so since we resolved your issue with the alternative way I'm going to close this. Thanks for your help!

Issue #64, opened 2015-07-14
What type of bar codes can be read?


nraboy:
I'm sorry if this is an inappropriate place to ask, but what types of bar codes can be read with this component?  Maybe you can add them to the README?

Thanks,

jtremback:
I've never used the barcode functionality, but I saw this in the source:

```
           @"BarCodeType": @{
               @"upce": AVMetadataObjectTypeUPCECode,
               @"code39": AVMetadataObjectTypeCode39Code,
               @"code39mod43": AVMetadataObjectTypeCode39Mod43Code,
               @"ean13": AVMetadataObjectTypeEAN13Code,
               @"ean8":  AVMetadataObjectTypeEAN8Code,
               @"code93": AVMetadataObjectTypeCode93Code,
               @"code138": AVMetadataObjectTypeCode128Code,
               @"pdf417": AVMetadataObjectTypePDF417Code,
               @"qr": AVMetadataObjectTypeQRCode,
               @"aztec": AVMetadataObjectTypeAztecCode
               },
```

nraboy:
Awesome.  Thanks for finding this for me @jtremback :-)

lwansbrough:
@nraboy Yeah that's the list, also the scan event will return the type of barcode recognized.

liukepeng:
I want to  implement  scan QR_Code with React Native for Android.Can you help me?

lwansbrough:
@liukepeng Android is not yet supported unfortunately.

liukepeng:
Thank you very much!

PR #75, opened 2015-07-26
Adding barcode types to Readme

JedWatson:
I had the same question that was asked + answered in #64, so I thought it might be worth adding the list of supported barcode types to the Readme for other users.


lwansbrough:
Thanks!

PR #77, opened 2015-07-27
Tap-to-focus and Pinch-to-zoom

unknownexception:
I'm making an app which heavily use `react-native-camera` awesome module. And I'd like to make some contributions. This is initial approach to add two features: Tap to focus and Pinch to zoom. For now, it isn't customizable and I would be happy if you look at this, review and we could discuss some API changes and that should be configurable.
1. For now a small rechtangle has a fixed width (80px) and fixed animation (1 second blinking) and implemented totally on the native side. I think it should be up to user, so `<Camera focusedComponent={myComponent} />`? Or just `<Camera onFocusChanged={}`?
2. Focusing functionality is also has customizable constants. Maybe I should reuse `focusWithMode()` but allow user to set focus zone. 
3. Zoom could be calibrated with some constants too, I don't know if this fits for everybody or needs to be customized too. 
4. Should these features be enabled by default, but not always enabled? `<Camera tapToFocus={false} pinchToZoom={false}/>`?

### DIFF
--- a/Camera.ios.js
+++ b/Camera.ios.js
@@ -51,7 +51,10 @@ var Camera = React.createClass({
     torchMode: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.number
-    ])
+    ]),
+    defaultTouchToFocus: PropTypes.bool,
+    onFocusChanged: PropTypes.func,
+    onZoomChanged: PropTypes.func
   },
 
   mixins: [NativeMethodsMixin],
@@ -205,7 +208,10 @@ var RCTCamera = createReactNativeComponentClass({
     type: true,
     orientation: true,
     flashMode: true,
-    torchMode: true
+    torchMode: true,
+    onFocusChanged: true,
+    onZoomChanged: true,
+    defaultTouchToFocus: true
   }),
   uiViewClassName: 'RCTCamera',
 });
--- a/Camera.ios.js
+++ b/Camera.ios.js
@@ -52,7 +52,7 @@ var Camera = React.createClass({
       PropTypes.string,
       PropTypes.number
     ]),
-    defaultTouchToFocus: PropTypes.bool,
+    defaultOnFocusComponent: PropTypes.bool,
     onFocusChanged: PropTypes.func,
     onZoomChanged: PropTypes.func
   },
@@ -211,7 +211,7 @@ var RCTCamera = createReactNativeComponentClass({
     torchMode: true,
     onFocusChanged: true,
     onZoomChanged: true,
-    defaultTouchToFocus: true
+    defaultOnFocusComponent: true
   }),
   uiViewClassName: 'RCTCamera',
 });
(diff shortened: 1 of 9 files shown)


lwansbrough:
This is great, thanks so much! To answer your questions:

1. Por que no los dos? Might be useful to be able to listen to the focus event as well as being able to provide a focus component.
2. I'm indifferent - if you're developing these features I'll leave it to your discretion. I think you'll know what works best in practice if you're actually using these features in your app.
3. IMO the more customization the better. It always sucks when a module is perfect for someone except feature X that isn't exactly working how they need. As long as we can abstract most of the complexity away from the API.
4. For backwards compatibility I think they should be disabled by default.

unknownexception:
Sorry for delay, I've been preparing the app to AppStore. Thanks again for this module!
```javascript
  <Camera
        ref="cam"
        style={styles.container}
        type={this.state.cameraType}
        onFocusChanged={(e) => {console.log('focusChanged', e.nativeEvent}} 
        defaultTouchToFocus={true}
        onZoomChanged={() => {console.log('zoomChanged', e.nativeEvent}}
      ></Camera>
```
API could looks unclear. Everything is disabled by defaut, but providing handlers would turn these features on. `defaultTouchToFocus` is responsible to show native feedback if onFocusChanged is enabled, you can disable that. You get coordinates of touch in `e.nativeEvent`, so you can write your own component for this purpose.

I don't sure if I should add explicit flags for these events or just check handlers.

unknownexception:
I'm using it at http://hash.ai and found a couple of other bugs. Would be awesome to get feedback on this pull request.

morukutsu:
@unknownexception I've been using it a bit and it works really well for now, great work!
To be compatible with react-native 0.10+, minor changes are needed in RCTCameraManager.m.

``` javascript
- (NSDictionary *)customDirectEventTypes
{
    return @[
      @"focusChanged",
      @"zoomChanged",
    ];
}
```

Currently, I'm only using the tap-to-focus feature so I only tested that one.

lwansbrough:
I can merge these changes in if the above fix is applied upstream. Thanks again for the work you've put into this!

unknownexception:
Upgraded. Please, review `readme` additions.

morukutsu:
@lwansbrough I think you should consider merging this PR. Especially since you've made several changes in the main branch, it could complicate future merging efforts. The sooner, the better :+1:

lwansbrough:
I've been slow on this one, sorry. I had mentally shelved it because I was under the impression it was waiting on further changes. I'll add this asap.

lwansbrough:
Hey guys, please install and test this out to make sure it's working. I haven't published on npm yet - if you're unaware you can install from Github, IIRC: `npm install lwansbrough/react-native-camera#master`

lwansbrough:
Thanks again for the work you put in on this @unknownexception!

unknownexception:
@lwansbrough Thank you for awesome plugin! I'm using it to make my [app](http://hash.ai/) and it works great. I surely will replace current fork with the `origin/master` and test it on all devices.
I have some others changes for camera, though, which I'd like to contribute back, need to extract it into pieces.

eduedix:
@unknownexception I like your app, it is a shame that it is not open source and I can't mention on its own page :P

PR #82, opened 2015-08-27
Use proper check to see if running on simulator

datwelk:
# What
Fixes a busted `isSimulator` check in Xcode 7. Without this fix `isSimulator` will always return `NO` when running in Xcode 7. As a consequence the project will crash with the message `AVCaptureSession can't startRunning between calls to beginConfiguration / commitConfiguration`.

@lwansbrough

### DIFF
--- a/RCTCameraManager.m
+++ b/RCTCameraManager.m
@@ -213,13 +213,11 @@ RCT_EXPORT_METHOD(stopCapture) {
   }
 }
 
-- (BOOL)isSimulator {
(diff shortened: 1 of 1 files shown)


lwansbrough:
Thanks for your contribution!

RGBz:
It doesn't look like this was published to npm yet.

lwansbrough:
Oh right, I'll push a new version later tonight.

PR #85, opened 2015-09-05
move beginConfiguration down to prevent not commiting

jim-lake:
Crash caused by beginConfiguration without commitConfiguration.  Move down the beginConfiguration to be after error checking to ensure we always call commit after begin.


lwansbrough:
Hi Jim, can you merge the latest PR from upstream into your branch so that it can be auto-merged here?

PR #91, opened 2015-09-22
Update Header Search File Paths

qaiken:
Update Header Search File Paths to Reflect New Directories


lwansbrough:
Thanks, managed to do this in the XCode project and forgot about the readme. Cheers!

PR #100, opened 2015-10-07
Suggestion about memory target

machard:
Hi,

This is my proposition about the memory target.
Currently it returns a base64 string though the bridge, which is not really something good to do.

In react native there already is an in-memory store for images, RTCImageStoreManager.
So my change suggests to store the image in it and return the specific store uri created by it. These uri can be used in any of the built in react image stuff.
People who want the base64 can always do so as RTCImageStoreManager exposes getBase64ForTag.

My use case is that I want to implement a series of treatments (resizes, metadata, watermark..) to the image before saving the resulting images to disk. I want to be able to command the transformations in JS for more flexibility.
## Warning, below this line is just an extreme conclusion i had after making this change.

This also led me to think that this camera module should only return in memory images and let the other stuff (resize, metata/gps edition, file/camera roll saving...) to be handled by others modules.
That's what react native ImageEditingManager does for example.

### DIFF
--- a/ios/RCTCamera.xcodeproj/project.pbxproj
+++ b/ios/RCTCamera.xcodeproj/project.pbxproj
@@ -184,6 +184,7 @@
 					/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include,
 					"$(SRCROOT)/../../../React/**",
 					"$(SRCROOT)/../../react-native/React/**",
+					"$(SRCROOT)/../../react-native/Libraries/**",
 				);
 				IPHONEOS_DEPLOYMENT_TARGET = 8.2;
 				MTL_ENABLE_DEBUG_INFO = YES;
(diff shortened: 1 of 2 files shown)


lwansbrough:
I haven't looked at the code much yet but I really like how this sounds. I'll review it later. Thanks @machard!

machard:
At the end i am not so sure that this is a good idea as RCTImageStoreManager stores them as UIImage : metadata are lost

lwansbrough:
@machard What if we send back the URI with the metadata? Better or worse than the current implementation?

machard:
I'll make a pull request to react-native to change how RCTImageStoreManager stores images.
it seems easily doable

lwansbrough:
@machard Legend. Keep me updated.

machard:
It's been merged !

lwansbrough:
@machard Awesome so does that mean this is all ready to be pulled in then?

machard:
It seems it will be part of 0.16. but a slight change will be needed (storeImage => storeImageData)

lwansbrough:
@machard Noted.

rt2zz:
@machard @lwansbrough amazingly this still merges cleanly 7 months later :o 

I think this change makes sense but correct me if I am wrong it appears ImageStore is iOS only right now. In this case I propose the following:

1. rename target `memory` -> `base64`
2. merge this solution as target: `imageStore` and document it as iOS only

Also we will need to rename storeImage -> storeImageData as @machard pointed out. Thoughts?

Minishlink:
@lwansbrough @rt2zz @machard Any updates on this?

rt2zz:
no update, I think we all agree this makes sense but would need the updates mentioned here https://github.com/lwansbrough/react-native-camera/pull/100#issuecomment-204851866 before we could get it in.

PR #109, opened 2015-10-22
use the AVMetadataObjectTypeDataMatrixCode only if the system support it...

yinuoba:
use the AVMetadataObjectTypeDataMatrixCode only if the system support this API.
and then we can use the react-native-camera in iOS7.
otherwise it will compile error if the system version less than iOS8.


lwansbrough:
@yinuoba Thank you. I'm preparing a bug fix release that should be out in a day or two if I can find some time. :)

PR #113, opened 2015-11-02
Added a step in the 'Getting Started' section

zckly:
Added the location in which Header Search Paths can be found to make it easier for people that are unfamiliar with XCode to find Header Search Paths.


lwansbrough:
Thanks!

PR #114, opened 2015-11-02
implemented method getFOV for ios

WellDone2094:
implemented method in RCTCameraManager for IOS to get the FOV, the method accept a callback that will be called with an object as argument, the object has this format
```
{
   backCamera: double,
   frontCamera: double
}
```


lwansbrough:
Great, thank you for this. I may make some minor changes before merging this in, nice feature add though.

peveloper:
Very cool :+1:

Issue #80, opened 2015-08-26
Camera freezes when opened 2 times via navigator


hrlarsen:
Hi

Currently have an app where you are meant to take several photos. Currently I have a class handling the camera capturing (CameraCapture.js), and this is opened using the IOS navigator with a push from a "capture images" class (imagine a number of empty thumbails that needs to be filled by capturing photos). When the CameraCapture is pushed the first time everything is fine and I can capture a picture etc. But when I then try to open it again it freezes. 

The freeze does not happen when I pop the "capture images" class and open it again. So it's like the camera is not properly destroyed or something when the component is unmounted, and only destructs when the class holding the CameraCapture class is unmounted.

hrlarsen:
I found that this happens because CameraCapture variable ```javascript var CameraCapture = require('./CameraCapture') ``` creates an instance of the CameraCapture object, and this cannot handle showing the <Camera> more than once.

achang28:
I hate to say it, but I, too am still experiencing freezes much like @hrlarsen is experiencing (even though 0.3.3 reduced the original problem). I wonder if there is some sort of "cleanup" or additional mounting process that should be done in the obj-c world, in addition to what has already been built in react-native-camera?

lwansbrough:
@hrlarsen Interesting. I'll look more into this sometime soon.

zubru:
+1 I also have problem with this. When I go back from view with camera and once again go into it, camera freezes

jforaker:
+1 
Does anyone have a solution for this?

jforaker:
SOLVED:

http://stackoverflow.com/a/33048580/1516309

@zubru 
@lwansbrough 
@achang28 
@hrlarsen

lwansbrough:
Thanks Jakeed1 of StackOverflow for this :) I'll implement sometime tonight hopefully. Tomorrow latest.

jforaker:
That is me ;)

lwansbrough:
@jforaker Oops hahah, of course. Thanks!

lwansbrough:
Can you guys give it a shot? `0.3.7`

gpbl:
I'm using 0.3.7 but the camera still freezes as i go back to the view using it

lwansbrough:
@gpbl Can you confirm that you completely reinstalled the camera module? It's possible the binaries haven't updated. It's also possible that this fix wasn't the proper fix. I haven't tested it myself.

gpbl:
[...]

PR #115, opened 2015-11-02
[bugfix] RCTViewManager view method must always return a new instance

morukutsu:
Fix for https://github.com/lwansbrough/react-native-camera/issues/107

Manager "view" method must always allocate and return a new instance.

### DIFF
--- a/ios/RCTCameraManager.m
+++ b/ios/RCTCameraManager.m
@@ -16,11 +16,7 @@ RCT_EXPORT_MODULE();
 
 - (UIView *)view
 {
-    if(!self.camera){
-        self.camera = [[RCTCamera alloc] initWithManager:self bridge:self.bridge];
-        return self.camera;
-    }
-    return self.camera;
+    return [[RCTCamera alloc] initWithManager:self bridge:self.bridge];
 }
 
 RCT_EXPORT_VIEW_PROPERTY(aspect, NSInteger);
@@ -284,7 +280,6 @@ RCT_EXPORT_METHOD(stopCapture) {
 #endif
 	
   dispatch_async(self.sessionQueue, ^{
-    self.camera = nil;
     [self.previewLayer removeFromSuperlayer];
     [self.session stopRunning];
     for(AVCaptureInput *input in self.session.inputs) {


lwansbrough:
I'm interested in accepting this PR, however I'm concerned it will conflict with changes made here, which were implemented to solve a different issue: https://github.com/lwansbrough/react-native-camera/commit/8cc61edef2c018b81e1c52f13c7d261fe6a35a63

morukutsu:
It took me quite some time to figure out what was the issue. For the explanation, see this comment: https://github.com/facebook/react-native/blob/c5b990f65f0791041378a61cd503ccf686b00bb1/React/Views/RCTViewManager.h#L35
So I'm pretty positive this is the right way to fix it. 

The problem is the bugfix for #80 is not the right one. It fixes the bug but does a regression. I agree you should not merge this now since it could reintroduce the other bug.

I will reproduce it and work on a proper fix for #80. I use this module a lot so I don't mind spending some time to support it ;)

morukutsu:
See: https://github.com/morukutsu/rn-cam-tests/blob/master/index.ios.js
With this example, I cannot reproduce #80 with react-native 0.14 & this PR patch.

@hrlarsen @zubru @jforaker gpbl Hello, could someone check the test case I provided is similar to what you were doing? Thanks!

morukutsu:
The test attempt to validate that mounting and unmounting the Camera component (manually by removing it from the vdom or by removing the current scene from the navigator) will not freeze.

realjamie:
@morukutsu , I tested your code and it works without crashing.

sgonyea:
This works great.  I'd love to see this get a merge / release!

lwansbrough:
@morukutsu Thank you for your contribution.

FrankSalad:
@lwansbrough  #80 still occurs - if the view hierarchy isn't reset by the navigator. If a component in the back stack declares a <Camera> and the component in the foreground declares a <Camera>  RCTCamera infinitely loops in `layoutSubviews`

morukutsu:
Hey @FrankSalad, based on https://github.com/morukutsu/rn-cam-tests/blob/master/index.ios.js.
Can you provide a repro for the bug?
Based on what you describe, the bug could be triggered by declaring a Camera in the route 0?

Thanks

thsig:
I'm encountering the same as @FrankSalad. Replacing
```objc
- (UIView *)view
{
    return [[RCTCamera alloc] initWithManager:self bridge:self.bridge];
}
```
with
```objc
- (UIView *)view
{
    if(!self.camera){
        self.camera = [[RCTCamera alloc] initWithManager:self bridge:self.bridge];
        return self.camera;
    }
    return self.camera;
}
```
in `RCTCameraManager.m` fixed this for me (from http://stackoverflow.com/questions/33046061/react-native-camera-freezes-app-when-opened-2-times-via-navigator/33048580#33048580).

Hopefully I'll have time to reproduce this using a sandbox repo sometime soon.

I'll be using this patch for now, but perhaps it should be merged into the main repo?

morukutsu:
You can try this fix but you will probably wake up https://github.com/lwansbrough/react-native-camera/issues/107. The way react-native seems to work suggest the single instance fix proposal is not the right one.

A proper repo from any of the bug reported would be handy ;)

thsig:
Ah, good point - that's not currently a problem for my app, but that means that a different fix is called for. Again, hope to find time to reproduce properly soon. Thanks!

PR #129, opened 2015-11-30
Added code to check audio permissions.

pjcabrera:
Added code to check audio permissions in ios/RCTCameraManager.m method checkDeviceAuthorizationStatus.

This fixes https://github.com/lwansbrough/react-native-camera/issues/124


danlopez:
:+1: - I work with @pjcabrera, and this PR allows us to check audio and video permissions concurrently.

lwansbrough:
Hmm... This is okay, but what if I don't want microphone access? It's better than what we have now so I think I'll accept it - but the API should probably change to `requestAccessToMediaType(mediaType, callback)`.

Thanks for your help!

PR #130, opened 2015-11-30
Fix crash when user pinches to zoom on devices that don't support it.

pjcabrera:
When zoomFactor was computed with a value greater than device.activeFormat.videoMaxZoomFactor, it caused a crash. Changed the pinch to zoom code to respect the value of device.activeFormat.videoMaxZoomFactor.

This fixes https://github.com/lwansbrough/react-native-camera/issues/107


lwansbrough:
Beautiful, thank you.

PR #131, opened 2015-12-02
Fix for back camera showing briefly when the type property is set to 'front'.

pjcabrera:
The first time an app is run that uses this component with the type set to 'front', the back camera shows briefly before the component obeys the type property and switches to the front camera. The presetCamera should be set to the camera type set in the props when the component is created.

### DIFF
--- a/ios/RCTCameraManager.m
+++ b/ios/RCTCameraManager.m
@@ -14,6 +14,12 @@
 
 RCT_EXPORT_MODULE();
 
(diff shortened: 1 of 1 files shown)


lwansbrough:
Nice, thanks!

pjcabrera:
@lwansbrough could you publish an updated npm module? It's been a couple months since my PRs were merged, but the latest npm module is 4 months old. I'd like to stop using my fork of this project as a package.js dependency. Thanks!

Issue #134, opened 2015-12-13
Check permissions to access Camera


bpeters:
If a user denies permissions for the camera and then tries to access the camera the app will crash. Is there a way to check for camera permissions from the Camera Api? I can then request permission with an alert like the default.

lwansbrough:
Apparently I forgot to expose the `checkDeviceAuthorizationStatus` via the JS API. You can do what you're asking via:

```javascript
NativeModules.CameraManager.checkDeviceAuthorizationStatus((function(err, isAuthorized) {}
```

I'd accept a PR if you want to expose that method on the `Camera` reference like others have been in the `index.ios.js` file.

tonmanayo:
Hi, I'd like to know if there is any way of checking Camera permissions. I'd like to being up a modal before asking the users on IOS for permissions

PR #139, opened 2015-12-22
Expose camera authorization

bradbumbalough:
In response to issue #134, I added the `checkDeviceAuthorizationStatus` method to the Camera reference API. Since it's not a value list, I opted to not add it to the `constants` var.


lwansbrough:
Thanks Brad, looks good.

bradbumbalough:
Thanks!

PR #140, opened 2015-12-29
Fix: Wrong image orientation when device orientation is locked

toriumi0118:
I fixed the bug that the image saved as portrait orientation when the device auto orientation lock mode.
Because UIDeviceOrientationDidChangeNotification is not worked in the locked mode.
(I attached the picture which show the locked mode.)

You need to import **CoreMotion.framework** to make it work.

### DIFF
--- a/ios/RCTCamera.h
+++ b/ios/RCTCamera.h
@@ -1,6 +1,7 @@
 #import <UIKit/UIKit.h>
 #import <AVFoundation/AVFoundation.h>
 #import "CameraFocusSquare.h"
+#import <CoreMotion/CoreMotion.h>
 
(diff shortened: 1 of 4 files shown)


ptomasroos:
Hey @toriumi0118 This seems like a great fix.

Can you rebase this on top of master so I can review it?

toriumi0118:
@ptomasroos 
yap, it's so long time ago and I throw away my app which this lib is used.
so I cannot check it, even if I could fix again or fix conflict.

So please close this.

rt2zz:
@toriumi0118 sorry we were not able to process this, and thank you!

PR #151, opened 2016-01-12
Add option to store captures in temp directory

christopherdro:
Add option to save to temp directory for images and videos.


rt2zz:
+1 this is probably a better option for most apps rather than saving to documents directory

lwansbrough:
Thanks a lot!

PR #159, opened 2016-01-24
Add timestamp to simulator still capture

sharathprabhal:
When working with the simulator, its not always obvious if new pics are taken because they are all blank. This PR adds a random color and timestamp to stills captutered on the simulator.


sharathprabhal:
@lwansbrough Fixes #162

lwansbrough:
I like it! Thanks.

PR #163, opened 2016-01-26
updated readme to include the import for android

ardowz:
Missing import line in Main Activity which prevents it from compiling


lwansbrough:
There's a lot more going on in this than what is described.

lwansbrough:
Please re-open this on a branch that wont include other files.

PR #166, opened 2016-01-27
[Android] Add option for capturing to temp directory

christopherdro:
/cc @rt2zz


rt2zz:
+1 tested & working

rt2zz:
cc @lwansbrough

Issue #168, opened 2016-01-27
Switch to Promise-based API


corbt:
React Native is standardizing on promises over callbacks for async operations ([discussion](https://github.com/facebook/react-native/issues/4971), implementations: https://github.com/facebook/react-native/pull/5336, https://github.com/facebook/react-native/pull/5504, https://github.com/facebook/react-native/pull/5512). I think it would be good for this library to follow suit, as it will be in line with platform standards and also easier to use. Example of what the new API would look like dropped into the README example:

```js
// Raw promise
_takePicture() {
  this.refs.cam.capture().then(
    data => console.log(data),
    error => console.log(error)
  );
}
```

```js
// Async/Await
async _takePicture() {
  try {
    const data = await this.refs.cam.capture();
    console.log(data);
  } catch (e) {
    console.log(e);
  }
}
```

This doesn't have to be a breaking change (at least immediately) because we can check for the presence of a callback argument and fallback to that if passed, but I think that promises should be the default.

If this sounds good I'm happy to put together a PR.

lwansbrough:
There are a lot of changes we want to make to the API, so I'm not against breaking the API for a major release version. async/await capture would be nice.

PR #170, opened 2016-01-28
Switch from callbacks to promises

corbt:
This PR implements #168. I've tested it on iOS with both video and photo captures and it works great. Haven't had a chance to test on Android, but the change is quite minor (especially since Android doesn't implement video yet) so it shouldn't cause any trouble.


lwansbrough:
Thanks so much for this - I think it's a great change. 1.0.0 here we come!

PR #187, opened 2016-02-08
Fix promise returns

corbt:
Currently all of the methods that return a promise on iOS are broken in 0.20-rc. This is because the RCTPromiseBlock syntax has been changed to require an array to be returned.

DO NOT MERGE this PR until React Native 0.20 has landed, because it will cause a breaking change in apps on <=0.19.


corbt:
Turns out this was a bug on the RN side that has now been fixed, so this PR is no longer necessary. https://github.com/facebook/react-native/commit/c9a1956c4fd9cfb6bf883c835b501d097181e1a4

lwansbrough:
Thanks anyway!

PR #191, opened 2016-02-10
Add type for capture options

juggleru:


rt2zz:
looks like this was added at some point https://github.com/lwansbrough/react-native-camera/blob/master/Camera.js#L159

:+1: thanks

PR #201, opened 2016-02-15
fix constants in defaultProps

rt2zz:
I am not 100% sure what the desired constants organization is here, but presumably `static constants` was updated to the new scheme while `defaultProps` was not.

Also standardizes no spaces on empty lines (before it was inconsistent)

### DIFF
--- a/Camera.js
+++ b/Camera.js
@@ -32,12 +32,12 @@ function convertStringProps(props) {
   if (typeof props.type === 'string') {
     newProps.type = Camera.constants.Type[props.type];
   }
-  
+
   return newProps;
 }
 
 export default class Camera extends Component {
-  
+
   static constants = {
     Aspect: CameraManager.Aspect,
     BarCodeType: CameraManager.BarCodeType,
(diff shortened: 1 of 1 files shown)


rt2zz:
Also threw in a bug fix for the remove listener logic. Because the listener was being set in an async function it was guaranteed to be defined by the time componentWillUnmount is called. This fixes that by defining the listener as the first line of `componentWillMount`

lwansbrough:
@rt2zz The defaultProps are actually using the constants defined above, so there shouldn't be an issue there. Feel free to merge this if you think your approach is better.

rt2zz:
I was getting a undefined error, which I assume is because Camera class attempts to define it's static properties before it has access to it's own (earlier defined) static properties. 

Maybe there was something specific about the way I am importing Camera - I take it from your description that you are not getting an undefined error?

lwansbrough:
No need to debug if there's a reasonable fix :)

rt2zz:
:+1:

PR #203, opened 2016-02-17
Add jpegQuality property to set Jpeg quality of captured picture (Android)

juggleru:
Set Jpeg quality of captured picture using Camera.Parameters.setJpegQuality(int quality)

### DIFF
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraViewManager.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraViewManager.java
@@ -39,7 +39,7 @@ public class RCTCameraViewManager extends ViewGroupManager<RCTCameraView> {
 
     @ReactProp(name = "jpegQuality")
     public void setJpegQuality(RCTCameraView view, int jpegQuality) {
-        view.setTorchMode(jpegQuality);
+        view.setJpegQuality(jpegQuality);
     }
 
     @ReactProp(name = "torchMode")
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraViewManager.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraViewManager.java
@@ -37,6 +37,11 @@ public class RCTCameraViewManager extends ViewGroupManager<RCTCameraView> {
         view.setCameraType(type);
     }
 
+    @ReactProp(name = "jpegQuality")
+    public void setJpegQuality(RCTCameraView view, int jpegQuality) {
+        view.setTorchMode(jpegQuality);
+    }
+
     @ReactProp(name = "torchMode")
     public void setTorchMode(RCTCameraView view, int torchMode) {
         view.setTorchMode(torchMode);
--- a/Camera.js
+++ b/Camera.js
@@ -77,6 +77,7 @@ export default class Camera extends Component {
       PropTypes.string,
       PropTypes.number
     ]),
+    jpegQuality: PropTypes.number,
     torchMode: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.number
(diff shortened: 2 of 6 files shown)


rt2zz:
This looks quite useful. Have you considered an iOS implementation as well?

android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraViewManager.java
```diff
@@ -37,6 +37,11 @@ public void setType(RCTCameraView view, int type) {
         view.setCameraType(type);
     }
 
+    @ReactProp(name = "jpegQuality")
+    public void setJpegQuality(RCTCameraView view, int jpegQuality) {
+        view.setTorchMode(jpegQuality);
```
rt2zz: This appears to be a mistake... 

As a side note, is it necessary to proxy all of these methods through all three java classes? I am not sure what the rationale was behind the original implementation.

juggleru: I just copied original style. I'm not sure that I want to rewrite whole component.
About iOS implementation - I'm not working with iOS at this time.

rt2zz: Sorry I meant the `view.setTorchMode(jpegQuality)` appears to be a mistake. presumably it should be `view.setJpegQaulity`

juggleru: Oh! I see, thanks a lot! )

rt2zz:
So I am thinking we can use this code, but expose the api of `captureQuality`. since ios only gives us low med high, we need to decide between either string constants or integers...

lwansbrough:
@rt2zz Agreed. `captureQuality` would be best - and would `quality` as a property of the `capture(options)` be another thing we should add?

As for strings or ints... we could expose constants for both platforms and allow integers as well - then just "round" the integer to the nearest (ceiling, better) quality.

manask88:
I want to see if I can help here in order to unify both iOS https://github.com/lwansbrough/react-native-camera/pull/207  and Android's API implementations. .. so i'd like to clarify a few things.
in iOS, we are only able to pass three values to define quality:
high, medium and low
in Android, it seems like it can take numbers from 1 to 100, where 100 is the best quality.

In order to unify this is what I think you guys suggested and also makes sense to me.
If we use the iOS API implementation as a guide, then, we could have these:
high -> 100
medium -> 50
low -> 1

In order to not lose the more fined grained quality the Android API has, we could have a prop named captureQualityAndroid which can take numbers from 1 to 100, and will take precedence over captureQuality if both have values.

lwansbrough:
@manask88 Thanks for offering help. I would prefer if the `captureQuality` could just take a string or an int I think. Maybe we could throw if `Platform.OS === 'ios' && typeof captureQuality === 'string'`. I'd also like to say that a quality of `1` would be *pretty* awful. Maybe 100, 70, 50?

pjcabrera:
@lwansbrough The numbers you suggest sound reasonable to me. Throwing an exception on iOS if the prop value is `typeof number` sounds reasonable too.

manask88:
I was planning to modify the API for the Android PR but then, I realized that It's probably better and easier for @juggleru to make the modifications since this is his PR and I won't be able to test on Android.
In order to unify both API in the easiest way, I think it will be simpler if you just change the prop name from `jpegQuality ` to `captureQuality` and then add a validation like https://github.com/apprennet/react-native-camera/blob/setSessionPreset/Camera.js#L37. So, in your case, you would do 
```js
 if (Platform.OS === 'android' && typeof props.captureQuality === 'number') {
    newProps.captureQuality = props.captureQuality;
  }
```

Please, let me know what you think @juggleru or if you have better ideas on how to tackle this.

manask88:
I have made the necessary changes to this PR to be aligned with the iOS one. Please, check it out here https://github.com/lwansbrough/react-native-camera/pull/225

PR #204, opened 2016-02-18
fix auto-focus mode on android

rpopovici:
This PR should fix auto focus problem with most front facing cameras on android. Basically, we check first to see if auto focus mode is available before setting it.
Also this PR should fix capture rotation for front facing cameras

### DIFF
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCamera.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCamera.java
@@ -131,16 +131,20 @@ public class RCTCamera {
         }
 
         CameraInfoWrapper cameraInfo = _cameraInfos.get(type);
-        int rotation = cameraInfo.info.orientation;
+        int displayRotation;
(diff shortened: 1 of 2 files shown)


rt2zz:
Just tested this out, looks good and is working :+1:

android/src/main/java/com/lwansbrough/RCTCamera/RCTCamera.java
```diff
@@ -131,16 +131,20 @@ private void adjustPreviewLayout(int type) {
...
+            displayRotation = rotation;
         }
         cameraInfo.rotation = rotation;
         // TODO: take in account the _orientation prop
 
-        camera.setDisplayOrientation(cameraInfo.rotation);
+        camera.setDisplayOrientation(displayRotation);
 
         Camera.Parameters parameters = camera.getParameters();
         parameters.setRotation(cameraInfo.rotation);
```
rt2zz: do you know why we have to reset the rotation after capture?

PR #208, opened 2016-02-22
Fixes infinite loop in RCTCamera.layoutSubviews

FrankSalad:
When someone activates multiple `<Camera>` components, for example, if they double tap a navigator button that opens a View with a `<Camera>` component, RCTCamera.m will call `[super layoutSubviews];` in an infinite loop. I believe it has to do with the shared `self.manager.previewLayer`. When one RCTCamera instance sets the `frame` on the same shared previewLayer, the other RCTCamera instance has its `layoutSubviews` called, it sets the frame, and the first one has `layoutSubviews` and they just keep handing off to each other infinitely. 

This patch has each RCTCamera instance allocate and manage its own preview layer. The disadvantage is that this previewLayer wont respond to `RCTCameraManger.changeAspect` and `RCTCameraManager.changeOrientation`. Open to other solutions.

### DIFF
--- a/ios/RCTCamera.h
+++ b/ios/RCTCamera.h
@@ -9,6 +9,7 @@
 @property (nonatomic) RCTCameraManager *manager;
 @property (nonatomic) RCTBridge *bridge;
 @property (nonatomic) RCTCameraFocusSquare *camFocus;
+@property (nonatomic) AVCaptureVideoPreviewLayer *previewLayer;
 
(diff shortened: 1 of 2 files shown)


rt2zz:
@FrankSalad sorry for taking so long to review this and thanks. it looks like this (or at least a similar change) landed at some point in the recent past: https://github.com/lwansbrough/react-native-camera/blob/master/ios/RCTCameraManager.m#L151

Feel free to rebase and reopen if the issue is not sufficiently resolved.

PR #211, opened 2016-02-24
set capture quality on android and iOS

rpopovici:
Allow the camera to use the highest available resolution for both preview and capture on both iOS and android.

### DIFF
--- a/Camera.js
+++ b/Camera.js
@@ -33,6 +33,10 @@ function convertStringProps(props) {
     newProps.type = Camera.constants.Type[props.type];
   }
 
+  if (typeof props.captureQuality === 'string') {
+    newProps.captureQuality = Camera.constants.CaptureQuality[props.captureQuality];
+  }
+
   return newProps;
 }
 
@@ -44,6 +48,7 @@ export default class Camera extends Component {
     Type: CameraManager.Type,
     CaptureMode: CameraManager.CaptureMode,
     CaptureTarget: CameraManager.CaptureTarget,
+    CaptureQuality: CameraManager.CaptureQuality,
     Orientation: CameraManager.Orientation,
     FlashMode: CameraManager.FlashMode,
     TorchMode: CameraManager.TorchMode
@@ -57,6 +62,10 @@ export default class Camera extends Component {
     ]),
     captureAudio: PropTypes.bool,
     captureMode: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.number
+    ]),
+    captureQuality: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.number
     ]),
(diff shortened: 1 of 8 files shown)


rpopovici:
On android the getBestPreviewSize() and getBestPictureSize() methods are more or less copy paste from https://github.com/Tredsite/react-native-camera, credits go to Tredsite.

rt2zz:
interesting, do you know why tredsite is working independently on a repo with no issues allowed? It would be good to collaborate. 

As for this PR, the implementation looks good but I am concerned about it interacting with configurable picture quality. This will need to be coordinated with the two other quality PR's.

rpopovici:
@rt2zz I have no idea why they choose to work independently, but Tredsite is not the only one doing this. Also Reaction-Framework https://github.com/Reaction-Framework/react-native-camera and timmh https://github.com/timmh/react-native-camera have their own implementations.

Anyway, I agree with you that we should have these quality PRs coordinated. From what I have seen, there is a big difference between iOS and Android when it comes to setting the resulution. On iOS we have presets and on android we have actual resolutions. And to make it more complicated, on iOS some presets are available only in picture mode and the other only in video. On android some devices like the Samsungs provide an inverted lists of resolutions :D and these resolutions differ for picture/video/preview. Also, finding the available resolutions requires to have the camera initialized, so adding a static method to get all this info probably will not work before an actual instantiation of the native module.   
My proposal here would be to have a 'configure' callback as a prop in Camera module. And this prop can be called at init time with available resolutions/presets(this can be a simple array of strings like: '640x480', '1980x1080' .. for android, and 'low', 'medium' .. for iOS ) and this callback should return back the selected valued. In this way we can overcome all the existing limitations. Of course, this can be extended also to other props like autofocus, torch, etc

alexkehayias:
When I tried this patch in a project images are very dark when captured on iPhone6+. Looks like it needs to configured earlier in the session instead of on `captureStill`?

rpopovici:
@alexkehayias you are right. I experienced similar behavior.

agrass:
Hi @rpopovici, thanks for this work. you already started with your proposal solution to configure the image quality? What do you think to have a standard presets for all? To make it more simple, but inside each one do the conversion. For example if we decide to have low, medium and high then inside Android may be we can do something like this: height = max resolution available, medium = half of max resolution available?

rpopovici:
@agrass I don't know if simplifying is a good solution because there are too many differences between the platforms. My proposal would be to give the user a list with all available options and let him choose from those. For example on iOS, picture mode, you have: low, medium, high and very high(AVCaptureSessionPresetPhoto). In video mode you don't get to have the very high option. On android you get a list with available resolutions and let the user choose the one he sees fit. On android you can provide also a preview resolution and so on.. If I have time this wkend, I will try to implement something.

rpopovici:
@agrass it's done!

rt2zz:
awesome stuff. I just tested and confirmed working on android. Can someone confirm iOS working?

agrass:
Great solution! Thank you @rpopovici

rt2zz:
confirmed working on iOS as well +1

@lwansbrough I believe this is ready to merge

lwansbrough:
Feel free to merge @rt2zz!

rt2zz:
:+1: :facepunch:

antoncode007:
It appears that the maximum IOS resolution is 1080 x 1920. This is with the "captureQuality={Camera.constants.CaptureQuality.high}" setting. Is that the expected behavior?

PR #212, opened 2016-02-24
Add new mirrorImage prop (iOS only right now).

dhrrgn:
This prop will mirror the resulting image if `true`.

**Note:** This is iOS only at the moment because I do not have an Android device to test an Android implementation on at the moment.

### DIFF
--- a/Camera.js
+++ b/Camera.js
@@ -82,6 +82,7 @@ export default class Camera extends Component {
     onBarCodeRead: PropTypes.func,
     onFocusChanged: PropTypes.func,
     onZoomChanged: PropTypes.func,
+    mirrorImage: PropTypes.bool,
     orientation: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.number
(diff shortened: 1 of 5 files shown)


rt2zz:
This looks good implementation wise. However I think a better api might be: `mirrorImage` which will mirror the image regardless of if it is front or back camera.

dhrrgn:
@rt2zz Good call.  I updated the Pull Request.

ustasb:
@rt2zz What's preventing this from being merged?

rt2zz:
Originally the plan was to keep *all* image transform / processing to other modules. IMO mirror image is a potential exception to this rule since it is such a common use case and the code to support it is trivial. That said it would be great to get additional community feedback, and also for @lwansbrough to weight in.

dhrrgn:
@rt2zz This needs to be in the library IMO.  As I have implemented it, it simply applies a transform to the image in-memory.  If it were left to a 3rd party module, the image would be saved to disk, then the path sent over the bridge, then it would need sent back over the bridge to the 3rd-party module, read from disk, transformed, re-written to disk, and sent back over the bridge.  This is far from ideal.

lwansbrough:
Makes sense to me! I wouldn't particularly qualify this as image manipulation. :+1:

rt2zz:
:+1:

PR #213, opened 2016-02-24
Fix zoomFactor getting reset to 1.0 when it goes above the max.

dhrrgn:
The zoomFactor should get set the max, not 1.0, when it goes above the max, otherwise, when you go past the max, it resets the zoom to 1.0 and just keeps re-zooming in while spreading your fingers.


lwansbrough:
Thank you!

PR #217, opened 2016-03-02
Add an example app.

danhawkes:
To simplify testing on various devices.


charpeni:
Closed in favor of the #304

PR #225, opened 2016-03-14
change prop name from JpegQuality to captureQuality

manask88:
This is based on https://github.com/IntelliQru/react-native-camera work.
All what I did is to change the change prop name from JpegQuality to captureQuality in order to align the Android PR https://github.com/lwansbrough/react-native-camera/pull/203 to https://github.com/lwansbrough/react-native-camera/pull/207 . I will just need somebody with the Android env installed to test these changes out

### DIFF
--- a/Camera.js
+++ b/Camera.js
@@ -33,6 +33,10 @@ function convertStringProps(props) {
     newProps.type = Camera.constants.Type[props.type];
   }
 
+  if (Platform.OS === 'android' && typeof props.captureQuality === 'number') {
+   newProps.captureQuality = props.captureQuality;
+ }
+
   return newProps;
 }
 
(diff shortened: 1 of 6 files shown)


rpopovici:
Guys, you got this all wrong. JPEG quality has nothing to do with snapshot resolution. On iOS, in PR https://github.com/lwansbrough/react-native-camera/pull/207/files you are setting the snapshot resolution by changing the sessionPreset value. On Android you are changing the JPEG compression quality which has nothing to do with snapshot resolution. 
For more info take a look at https://github.com/lwansbrough/react-native-camera/pull/211

rt2zz:
@rpopovici thanks for clarifying. Afaict there is no way to directly set jpeg quality on iOS, is that correct? If this is the case perhaps we should take the android PR as is  in #203. Thoughts?

rpopovici:
@rt2zz you should be able to set the JPEG quality for the snapshot, but I don't know exactly how. UIImageJPEGRepresentation(image, compressionQuality) allows you to set the compression quality but don't know if we can use this in this context

rt2zz:
Currently we use, `AVCaptureStillImageOutput jpegStillImageNSDataRepresentation` which lacks a quality param. Also it seems to me captureQuality is sufficient.

PR #228, opened 2016-03-16
transform barcode metadata points to screen coordinates

br3tt:
The `onBarCodeRead` event provides bounds-- but they are in a not-so-useful points format. This will convert them into _real_ screen coordinates.


rt2zz:
:+1: is there any use case to also have the untransformed values?

PR #230, opened 2016-03-17
Add audio and video authorization check separately

niborb:


rt2zz:
Couple of comments:
1) We probably need to think more about how these permissions checks work. e.g. if captureAudio is false when Camera mounts but changes to true, we need some extra logic to handle that.
2) I think the naming is a bit confusing Audio/Camera/Device

Nevertheless this is crucial to get in as requesting audio permission is obtrusive in its current state. :+1:

Issue #227, opened 2016-03-16
Saved image is incorrectly rotated - picture orientation does not match the preview orientation.


cdimascio:
### Steps to reproduce
1. Create react native app using sample code provided in README.md
2. Run app on a Nexus 9 (or any device whose default orientation is landscape not portrait)
3. Hold device in portrait mode and Take a picture. (Picture is now saved)
4. Open picture. The picture is rotated incorrectly. It's sideways
### Expected behaviour

Saved Picture should retain the same orientation as when taken.
### Actual behaviour

Saved Pictured orientation is incorrectly rotated
### Environment
- **Node.js version**:  5.2.0
- **React Native version**: 0.21
- **React Native platform + platform version**: iOS 9.0, Android 5.0, etc Android 6
### react-native-camera

**Version**: npm version or "master"

KalebPortillo:
+1

rpopovici:
This is happening because on most SAMSUNGs and some other devices the orientation information is saved in EXIF metadata tag. You have to use this information if you want to properly visualize this picture. Take a look here for more info http://stackoverflow.com/questions/28298956/captured-image-getting-rotated-on-nexus-9

rpopovici:
Guys I made a PR which is suppose to fix the locked orientation problem. This might fix the Nexus 9 issue too https://github.com/lwansbrough/react-native-camera/pull/232

ljones87:
+1 any further resolution on this? Trying to work with react-native-camera on google pixel 7.1.2. Also utilizing react-native-orientation-lock in landscape mode..

vagnerSl:
+1

ljones87:
Short term fix: we used the react-native-camera, orientation={Camera.constants.Orientation.landscapeLeft}, in conjunction with react-native-image-resizer: ex. ImageResizer.createResizedImage(data.path, 640, 480, 'JPEG', 100, 0)), the last number, 0, specifying the rotation, as well as react-native-orientation: Orientation.lockToLandscape().

42void:
+1

victorbadila:
+1

facuescobar:
+1

mylestan:
Ran into this issue was well. The use of:

> react-native-image-resizer: ex. ImageResizer.createResizedImage(data.path, 640, 480, 'JPEG', 100, 0))

As suggested by @ljones87 worked for me.

ospfranco:
[...]

PR #232, opened 2016-03-23
fix snapshot orientation when device orientation is locked

rpopovici:
As you probably noticed, when the device has locked orientation or app is locked to a specific orientation, the snapshot rotation is broken. This happens because the orientation events are fired only when the orientation is unlocked. In this PR I am using the sensor information to get the actual device orientation. The hardware events are fired only once when capture method is called, in order to avoid battery drain or event queue overloading. I tested on on both iOS and droid and it works great. On iOS I took the PR done by @toriumi0118 and refactored it to fit the single event idea.

### DIFF
--- a/ios/RCTCameraManager.m
+++ b/ios/RCTCameraManager.m
@@ -264,9 +264,6 @@ RCT_EXPORT_METHOD(changeFlashMode:(NSInteger)flashMode) {
 
 RCT_EXPORT_METHOD(changeOrientation:(NSInteger)orientation) {
   [self setOrientation:orientation];
-  if (self.previewLayer.connection.isVideoOrientationSupported) {
-    self.previewLayer.connection.videoOrientation = orientation;
-  }
 }
 
 RCT_EXPORT_METHOD(changeMirrorImage:(BOOL)mirrorImage) {
@@ -386,7 +383,6 @@ RCT_EXPORT_METHOD(hasFlash:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRej
       });
     }]];
 
-    [self.previewLayer.connection setVideoOrientation:self.orientation];
     [self.session startRunning];
   });
 }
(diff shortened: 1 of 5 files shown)


KalebPortillo:
Thanks @rpopovici

rpopovici:
did some code clean up on iOS

ios/RCTCameraManager.m
```diff
@@ -539,6 +549,7 @@ - (void)captureStill:(NSInteger)target options:(NSDictionary *)options resolve:(
           reject(RCTErrorUnspecified, nil, RCTErrorWithMessage(error.description));
         }
       }];
+      }];
```
corbt: Should this line be duplicated? It looks like an error.

corbt: Ah I see now that it's closing the `getDeviceOrientationWithBlock`, maybe it would be better to indent this whole block to match?

corbt:
I've tested this on iOS and can confirm that it works for still images. @rpopovici would you mind adding support for video as well? Here's a diff that works: https://github.com/corbt/react-native-camera/commit/d3d1ad5715017975568131179f853a5d3dddbaab (note that I only actually add 2 lines, it just shows that a bunch were changed because the indentation level is increased);

ios/RCTSensorOrientationChecker.m
```diff
@@ -0,0 +1,105 @@
...
+
+- (UIInterfaceOrientation)getOrientationBy:(CMAcceleration)acceleration
+{
+    if(acceleration.x >= 0.75) {
+        return UIInterfaceOrientationLandscapeLeft;
+    }
+    if(acceleration.x <= -0.75) {
+        return UIInterfaceOrientationLandscapeRight;
+    }
+    if(acceleration.y >= -0.75) {
```
corbt: This should be `<=`

corbt:
Also, this should only activate if the orientation is set to `auto` (the default).

corbt:
Also btw I know that I've listed some issues/tweaks, but overall great job with this functionality! We've had some very confused users who were recording in the wrong orientation (with orientation-locked devices) and unhappy with the results. This is a great feature for this library to have.

rpopovici:
Thanks @corbt! I will try to add video support also, but right now I am really busy with other stuff

rt2zz:
To better understand this change, can you explain what will happen in the following scenarios:

When device orientation is locked in portrait and:  

- | Camera orientation prop | sensor detected orientation | result 
---- | ---- | ---- | ----
1 | "landscapeLeft" | landscape | ?
2 | "landscapeLeft" | portrait | ?
3 | "auto" | landscape | ?
4 | "auto" | portrait | ?

In otherwords, does this PR play nicely with the orientation prop?

rpopovici:
@rt2zz I am very busy right now with my project, but I will come back to you with an answer asap

rpopovici:
@rt2zz I think now is ready for merge. So, basically sensor orientation is used only in auto mode and only for snapshot rotation. Preview is not affected by this PR.

rt2zz:
👍  will have to review, and hopefully land this after https://github.com/lwansbrough/react-native-camera/pull/270 gets in

sodik82:
sorry for hijacking communication - I am using `rpopovici:master` since I am interested in this feature (as I lock my application orientation to portrait) but it seems there can't be create issues for `rpopovici:master`.

When I am taking normal (portrait) image, it is rotated (even it should not). I am using Samsung S4, however everything seems to work well in iPhone and (android) Lenovo phone.

rpopovici:
@sodik82 take a look at the answer I gave here https://github.com/lwansbrough/react-native-camera/issues/227
RN's <Image> component should be able to show correctly the saved image.

sodik82:
if I understand it correctly, you are suggesting that I should use exif info. However there is no exif info in the image obtained from `react-native-camera`. 

my workflow is following:
1. I take picture `camera.capture()` and retrieve URI (phone is in normal - portrait position)
2. display the image as thumbnail - in `Image` - however it is incorrectly rotated.
3. I upload the file to server where I try to rotate according to exif info - however no exif info is there. When I upload image from CameraRoll (taken by standard camera) the image is correctly rotated.

when I try to take picture while the phone is in landscape position (while orientation is still locked to portrait) - the result image is upside down.

corbt:
Once this lands I'll open a new PR with support for video rotation on iOS.

rpopovici:
@corbt Already did that. Thanks!

rpopovici:
@sodik82 what Samsung are you using?

sodik82:
Samsung S4 GT-I9505, Android 5.0.1

is there anything I can investigate?

fanwashere:
I saw the wrongly orientated captured image problem (issue #227) specifically on Samsung phones (Galaxy S6, S6 Edge). Hoping this PR will resolve it.

@rpopovici I pulled your fork to test it out, but RN errors on [this duplicate declaration of playSoundOnCapture prop](https://github.com/rpopovici/react-native-camera/blob/master/index.js#L183). Otherwise, it runs fine and I'll test it again on a Galaxy S6 tomorrow to see if it fixes the orientation problem.

rt2zz:
👍 awesome work

This will still need to undergo further testing before we can cut a release though.

PR #235, opened 2016-03-27
Add hasFlash support for Android

cdimascio:
The following changes add hasFlash() support to Android. It uses same semantics as hasFlash() for iOS.


rt2zz:
:+1:

PR #239, opened 2016-03-31
Faster video file processing

corbt:
This PR drastically speeds up the processing time for large video files, as well as leading to lower overall disk usage. It accomplishes this by moving video into their final destination instead of copying them (and relying on the system to GC the old temp file).

Previously a 2-minute HD recording would take ~10s on my phone to finish processing once I called `stopCapture`; with this change it takes less than a second.


rt2zz:
:+1:

PR #240, opened 2016-03-31
capture returns metadata

corbt:
This is the first step towards implementing https://github.com/lwansbrough/react-native-camera/issues/188. This PR has two parts:

1. I've changed all `capture` callback invocations to return an object of the form `{'path': ...}` instead of returning the path directly.
2. I've also implemented `size`, `width`, `height` and `duration` return values for video captures on iOS, because that's the most important metadata for my use case.

Part (1) is most important, because by switching to returning objects everywhere the API will be future-proofed. More capture metadata can easily be added later.

cc @rt2zz 

EDIT: had linked to wrong issue above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ You can access component methods by adding a `ref` (ie. `ref="camera"`) prop to 
 
 #### `capture([options]): Promise`
 
-Captures data from the camera. What is captured is based on the `captureMode` and `captureTarget` props. `captureMode` tells the camera whether you want a still image or video. `captureTarget` allows you to specify how you want the data to be captured and sent back to you. See `captureTarget` under Properties to see the available values. The promise will be fulfilled with the image data or file handle of the image on disk, depending on `target`.
+Captures data from the camera. What is captured is based on the `captureMode` and `captureTarget` props. `captureMode` tells the camera whether you want a still image or video. `captureTarget` allows you to specify how you want the data to be captured and sent back to you. See `captureTarget` under Properties to see the available values.
 
 Supported options:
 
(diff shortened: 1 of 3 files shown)


rt2zz:
nice! I think this functionality is important. I will give it a test tomorrow. We should also get @lwansbrough to take a look since we are talking about a significant api change (resolving objects instead of strings to capture promise)

rt2zz:
Because this is such a major breaking change and many people are developing against master, I decided to take this into [v1 branch](https://github.com/lwansbrough/react-native-camera/tree/v1)

We will have to think about a strategy to switchover. Probably should publish a legacy version pre-change in npm.

rt2zz:
Sorry I realize closing all these PR's is not the most elegant solution to merging, but trying to achieve some element of efficiency here. Hopefully we can get v1 merged back to master this weekend.

corbt:
@rt2zz closing is fine by me as long as it makes it into a release eventually. :) Helps me keep track of my open PRs better too.

sscaff1:
@corbt Anyway to get height and width on the capture too?

corbt:
@sscaff1 yes that would be pretty easy to add, I only didn't because I haven't needed it for my own app. But I'm sure the maintainers would accept a PR for that functionality.

sscaff1:
@corbt That would be extremely useful for an app that I'm developing.

corbt:
@sscaff1 I'm probably not going to implement this myself in the short term anyway because my todo list is long enough as it is. :) But if you or anyone else wants to look into the code and make a PR for it I'm sure it would be accepted.

PR #245, opened 2016-04-03
[captureMode] conver string captureMode prop, audio permission handling

rt2zz:
1. convert string captureMode props
2. only request audio permission if captureMode = video


rt2zz:
took this into v1

PR #253, opened 2016-04-06
play shutter click sound on capture (android)

cdimascio:
title says it all.
Note that the MediaActionSound API used in this implementation was added for API level 16 which is the min SDK for react-native-camera.

### DIFF
--- a/Camera.js
+++ b/Camera.js
@@ -88,6 +88,7 @@ export default class Camera extends Component {
       PropTypes.string,
       PropTypes.number
     ]),
+    playSoundOnCapture: PropTypes.bool,
     torchMode: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.number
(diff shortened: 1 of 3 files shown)


rt2zz:
This looks great. In order to merge though we would need to put it behind a prop/option. Also would need an parallel functionality on the iOS side.

rt2zz:
one idea on the api is to add a prop `playSoundOnCapture={true/false}`. Then later if we choose to support more sounds we can convert to constants. Alternatively we could leave sound completely out of scope and consumers can choose to implement something like react-native-sound.

cdimascio:
Thanks for the feedback. I've implemented your idea to add playSoundOnCapture as a property. I have set the default as true. This seemed the proper choice as react-native-camera iOS plays a sound on capture. As for sound being in/out of scope. IMO, it should be part of react-native-camera, particularly since precedence has already been set via the IOS implementation. iOS always plays a sound on capture.

rt2zz:
:+1: lgtm

PR #254, opened 2016-04-09
Search only defined barcodes to improve performance

Sommerfeld:
We highly recommend to merge this to massively improve performance and not look for barcodes if u just want to take images or something. In our case we are just searching for qr codes, which is done way faster now.

This is just done in iOS and should also be implemented for Android. 

Someone may also want to reflect this in ReadMe.
`barCodeTypes={[Camera.constants.BarCodeType.qr]}`

Keep up the great work!:)

### DIFF
--- a/Camera.js
+++ b/Camera.js
@@ -84,6 +84,7 @@ export default class Camera extends Component {
     onFocusChanged: PropTypes.func,
     onZoomChanged: PropTypes.func,
     mirrorImage: PropTypes.bool,
+    barCodeTypes: PropTypes.array,
     orientation: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.number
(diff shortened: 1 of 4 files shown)


philipheinser:
+1

rt2zz:
yes! this looks great and is much needed.

We need to consider how to support "all" barcodes. Perhaps `barCodeTypes` defaults to all if onBarCodeRead is defined.

philipheinser:
@rt2zz we could define a key "all" thats adds available types. but not as default this could lead to the same misuse that we had before. for example when someone is using this only to capture an image the device would use the default and check for all barcodes.

rt2zz:
@philipheinser my suggestion is that barCodeTypes defaults to null (or []) **unless** `onBardcodeRead` is defined. In other words, no barcode scanning should happen unless it is being listened for.

I am wary of overloading the propType with both strings and arrays for `barCodeTypes`

Sommerfeld:
It's kinda late in Germany. You are absolutely correct. We read over `onBardcodeRead`.  We will overwork our pull request asap.

rt2zz:
Thanks! no worries about the defaults, I took this into [v1 branch](https://github.com/lwansbrough/react-native-camera/tree/v1) and implemented the defaulting behavior. 👍

PR #259, opened 2016-04-15
rename Camera->index

corbt:
This PR slightly simplifies the library's Javascript layout. It also improves Flow support for the library (Flow doesn't handle `*.[platform].js` files right now https://github.com/facebook/flow/issues/945).


rt2zz:
added to v1 :)

PR #260, opened 2016-04-15
Increase cohesion of iOS property methods

corbt:
Currently, there are number of props that are declared in `RCTCameraManager`, called on an `RCTCamera` instance and then immediately call back into methods on the `RCTCameraManager`. This leads to quite a confusing control flow. It also makes it more difficult to add or change existing props, because changes have to be kept in sync between three places.

This PR refactors a number of these properties to be handled entirely in the `RCTCameraManager` without crossing into `RCTCamera`. No behavior is affected, just the location that various methods are called. As a bonus, this change leads to -47 LOC.

### DIFF
--- a/Camera.js
+++ b/Camera.js
@@ -122,7 +122,7 @@ export default class Camera extends Component {
     playSoundOnCapture: true,
     torchMode: CameraManager.TorchMode.off,
     mirrorImage: false,
-    barCodeTypes: Object.keys(CameraManager.BarCodeType),
+    barCodeTypes: Object.values(CameraManager.BarCodeType),
   };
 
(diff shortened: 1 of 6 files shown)


rt2zz:
added to v1, PR'ed to master: https://github.com/lwansbrough/react-native-camera/pull/270

corbt:
@rt2zz I added a commit (https://github.com/corbt/react-native-camera/commit/3d0880de8da6c904b58c2092651e3a27afbc70bd) in this branch to remove the method definitions from `RCTCameraManager.h` but it's not showing up in the PR, maybe because it's closed?

rt2zz:
👍

PR #261, opened 2016-04-15
Torch fixes

corbt:
This PR fixes two issues with torch behavior on iOS.

1. Fixes https://github.com/lwansbrough/react-native-camera/issues/123 (can now initialize the component with the torch turned on).
2. Fixes an issue where the torch is on but then turns itself off as soon as a video recording starts.

Note that this PR seems really big, but that's just because it's based on https://github.com/lwansbrough/react-native-camera/pull/260. The actual changes related to this PR are in the second of the two commits on this branch, so reviewers can just look at that.

### DIFF
--- a/Camera.js
+++ b/Camera.js
@@ -122,7 +122,7 @@ export default class Camera extends Component {
     playSoundOnCapture: true,
     torchMode: CameraManager.TorchMode.off,
     mirrorImage: false,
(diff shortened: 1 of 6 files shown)


rt2zz:
I pulled this into v1 and PR'ed back to master here: https://github.com/lwansbrough/react-native-camera/pull/270 

👍

Issue #263, opened 2016-04-18
Android Error: still running in background when component is not showing anymore


agrass:
### Expected behaviour

Camera stop running when change the route view
### Actual behaviour

Camera still running and logging this like crazy: 

```
D/QCamera2HWI(  285): bool qcamera::QCamera2HardwareInterface::TsMakeupProcess(mm_camera_buf_def_t*, qcamera::QCameraStream*, unsigned char*, TSRect&) pch_makeup_enable = null
D/mm-camera-intf(  285): mm_stream_qbuf: VIDIOC_QBUF buf_index 2,stream type 1,frame_idx 305,queued cnt 5
E/mm-camera(  326): isp_thread_proc_cmd: HW update still on going
D/mm-camera-intf(  285): mm_stream_read_msm_frame: VIDIOC_DQBUF buf_index 1, frame_idx 306, stream type 7, queued cnt 7
D/mm-camera-intf(  285): mm_stream_qbuf: VIDIOC_QBUF buf_index 1,stream type 7,frame_idx 306,queued cnt 8
```

Then when I re render the component of camera it fail and don't show nothing. 
### Environment

Android with RN 0.19 or RN 0.24 
### react-native-camera

master

ospfranco:
Hey! this repository was recently transferred to the react-native-community, as part of our ongoing cleaning process we are closing old tickets (there is just too many of them!), please feel free to try the latest versions or the master branch to see if your problem has been solved.

If you are still experiencing this problem feel free to re-open this or fill a new ticket.

PR #264, opened 2016-04-18
Add releaseCamera Method in Android

agrass:
To fix problem of multiples cameras running and crashing in Android devices, issue #263 

Example:
Before changing the route to prevent re render and the instance of another camera when come back, I do:

``` 
this.camera.releaseCamera({target: Camera.constants.CaptureTarget.disk});
```


Camera.js
```diff
@@ -186,6 +186,22 @@ export default class Camera extends Component {
...
+      target: props.captureTarget,
+      quality: props.captureQuality,
+      type: props.type,
+      title: '',
+      description: '',
+      ...options
+    };
+    CameraManager.releaseCamera(options);
+  }
+
```
vespakoen: The releaseCamera method expects "type" to be in the options array, you pass in a lot of other stuff which seems unnecessary?

agrass: You are right, I'm going to remove the other options

corbt:
I haven't tried this library with Android, but could the component be configured to automatically release the camera when it's unmounted? That seems more ideal.

agrass:
You're right, @corbt . I closed this PR, because it's to fix something specific with `react-native-router-flux` or `NavigationExperimental`. I don't know if it's the best workaround but one fix in case of `react-native-router-flux` is unmounting the first camera before re open another one (using type='reset') when changing scene.

PR #266, opened 2016-04-22
separate React from ReactNative

corbt:
In React Native 0.25, React Native stops wrapping shared React functionality (createClass, PropTypes etc.) and expects end users to import these from the react package instead. This PR implements that behavior.

### DIFF
--- a/Camera.js
+++ b/Camera.js
@@ -1,5 +1,5 @@
-import React, {
-  Component,
+import React, { Component } from 'react';
+import {
   NativeAppEventEmitter,
   NativeModules,
   Platform,


rt2zz:
nice. Is there some way to do this backwards compatible? Or do we need to just reply on peer dependencies and helpful readmes?

corbt:
@rt2zz I think this is backwards compatible for any version of RN after React was moved to `peerDependencies` (https://github.com/facebook/react-native/commit/9f01f9677096117da8be721913a09db870505c2f), so going back to 0.21. Haven't actually tested it though.

rt2zz:
👍   Took this into branch v1: https://github.com/lwansbrough/react-native-camera/tree/v1

I will PR master with v1 later today. The big issue being the change in response shape from capture.

PR #270, opened 2016-04-23
v1 Updates

rt2zz:
I originally split out new dev into v1 because of the major breaking change to the response shape of capture. However this has made taking in PR's more difficult and has resulted in less people testing the new changes.

I recommend we take this branch back into master and continue dev there. People who are installing for master will likely feel some migration pain, but that should be expected when installing from master.


rt2zz:
included in this branch are:

- torch fixes
- improved barcode scanning performance (does not scan if no onBarcodeRead handler)
- capture returns object instead of string
- only check audio permission when required

rt2zz:
@lwansbrough it would be great if you can take a peak at this and 👍  or 👎

lwansbrough:
@rt2zz Looks good. I guess this is a WIP? Looks like a lot of the methods referenced in RCTCameraManager.h have been removed?

rt2zz:
@lwansbrough ah, good catch RCTCameraManager.h needs updating. The removed methods come from this commit: https://github.com/lwansbrough/react-native-camera/commit/7fc183c2e7fb2baeb4960ba69bc98bde4fa82e61#diff-21ad20d50067b33a89c9ea2e27c4624e by @corbt which replaced `set[Thing]` methods with RCT_CUSTOM_VIEW_PROPERTY implementations.

It looks like if this is the direction we want to go further cleanup is required, to remove said methods from the .h file and also get rid of the `RCT_EXPORT_METHOD(changeOrientation...`. That said I am not sure why `changeOrientation` was ever exported so I may be missing a use case.

Perhaps @corbt can weigh in.

corbt:
@rt2zz I'm guessing `changeOrientation` was exported just as a shorthand to access it from `setOrientation` in `RCTCamera.m` without having to declare it in `RCTCameraManager.h`. We could refactor it to just be a normal instance method on the `RCTCameraManager`, or alternatively move it entirely to the `RCTCameraManager` as a custom view property like I've done with others in https://github.com/lwansbrough/react-native-camera/pull/260, but that change wasn't as trivial as the other ones I made in that PR, so I decided to leave it out for now.

I've added a commit to https://github.com/lwansbrough/react-native-camera/pull/260 to remove the method declarations in `RCTCameraManager.h` that branch removed. Sorry about that!

rt2zz:
@corbt ah got it. I will cherry-pick that fix and update this PR tonight

rt2zz:
👍  I cherry-picked the CameraManager.h cleanup. 

I doubt this PR is bug free, but at this point we need as many eyes on it as possible. I would recommend merge.

lwansbrough:
If you think it's good to merge then go for it!

PR #271, opened 2016-04-23
V1 orientation

rt2zz:
this is a rebased version of https://github.com/lwansbrough/react-native-camera/pull/232 compliments of @rpopovici

I am inclined to take it in but I hesitate due to the large scale of the changes. @corbt can you weigh in on this?

### DIFF
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCamera.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCamera.java
@@ -224,6 +224,32 @@ public class RCTCamera {
         }
     }
 
+    public void adjustCameraRotationToDeviceOrientation(int type, int deviceOrientation)
+    {
+        Camera camera = _cameras.get(type);
+        if (null == camera) {
+            return;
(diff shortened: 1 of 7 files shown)


corbt:
Right now if I understand this right it activates unconditionally. I don't think it should activate if the user has explicitly set an orientation.

On the other hand, the documentation for `orientation` says:

> The orientation property allows you to specify the current orientation of the phone to ensure the viewfinder is "the right way up."

So maybe orientation is only related to the viewfinder, and this is unrelated?

rpopovici:
@rt2zz this is OK, but I don't think is ready for merging because is not taking in consideration the orientation parameter. I am using the real device orientation all the time, and I think this should apply only to auto mode. Beside this @corbt did some changes to add support for video on iOS. I am not seeing these changes here.

corbt:
For what it's worth, I'm using this PR in my own fork right now with support for video on iOS as well.

rt2zz:
got it, closing for now then we can continue discussion in 232

PR #288, opened 2016-05-16
Orientation only specifies recording output

corbt:
@rpopovici recently added functionality to automatically set the output file's orientation based on the device's sensor data, which is great. However, there are still some cases where you might need to explicitly set the `orientation` prop of the camera. For example, when the device is held near-horizontal the accelerometer can't know whether you were intending to record something in landscape or portrait.

The `orientation` prop is meant to deal with this kind of case, but unfortunately it is too powerful! Currently `orientation` sets the orientation of both the preview display and the output file. This makes it impossible to eg. record landscape videos if the user's UI is rotation-locked to portrait.

This PR simplifies the `orientation` code and updates it to **only determine the output file's rotation**. It's still important that the preview orientation stays in sync as well of course, so I've moved the preview rotation code to the camera's initialization, so the preview will always be in sync with the device's rotation.


rt2zz:
This makes a lot of sense to me 👍 

cc/ @rpopovici

rpopovici:
@rt2zz Yes, it makes sense to me too, to do this, but we have to do it on android also :D and I am really busy right now..

PR #292, opened 2016-05-23
always commit camera confiugration on capture session initialization

richardstephens:
Hi there. 

On attempting to use this component, I have found that the app crashes when leaving the camera view.  Examination of the logs showed that the crash was because configuration changes had not been committed to the session.  This fixes the problem for me.

Thanks.
-Richard


charpeni:
Fixed by #312.

PR #297, opened 2016-05-24
React, and Component should now be imported from 'react'.

darrylblake:
This fixes:
"Unhandled JS Exception: Super expression must either be null or a function, not undefined"


rt2zz:
👍

PR #300, opened 2016-05-27
Add methods for retrieving the authorization status on iOS without requesting permission 

dzautner:
Currently, the only way to get a media type's current authorization status on iOS is to call  `check${mediaType}AuthorizationStatus`.

While it works without an issue most of the time, if the user never set the app's permissions before on that specific media type, the call will pop up a permission alert. 

If one wish to react differently if the user have or have not seen the alert (priming comes to mind), he currently has no way to do it, as the simple call to check the auth status will result in the user being prompt with the permission alert. 

To address this issues I have introduced two native methods on iOS:

```
getVideoAuthorizationStatus
getAudioAuthorizationStatus
```

Both methods return a promise that is resolved with the media type's authorization status code. 

The iOS status codes are just an integer from 0-3, so for convenience I have also added constants that hold the meaning of said status codes to the exported constant map.

The PR also removes a few trailing white spaces found in RCTCameraManager.m.

### DIFF
--- a/index.js
+++ b/index.js
@@ -124,6 +124,9 @@ export default class Camera extends Component {
     barCodeTypes: Object.values(CameraManager.BarCodeType),
   };
 
+  static getVideoAuthorizationStatus = CameraManager.checkVideoAuthorizationStatus;
+  static getAudioAuthorizationStatus = CameraManager.checkAudioAuthorizationStatus;
+
   static checkDeviceAuthorizationStatus = CameraManager.checkDeviceAuthorizationStatus;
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -124,6 +130,9 @@ export default class Camera extends Component {
     barCodeTypes: Object.values(CameraManager.BarCodeType),
   };
 
+  static getVideoAuthorizationStatus = CameraManager.checkVideoAuthorizationStatus;
```
rt2zz: typo? I think this should be CameraManager.getVideoAuthorizationStatus

charpeni:
Any news on this @dzautner ?

PR #305, opened 2016-05-30
Export constants

charpeni:
`Camera.constants` are now accessible with :

```js
import { constants as CameraConstants } from 'react-native-camera';
```


rt2zz:
lgtm, will give it a couple of days for anyone else to weight in first...

PR #312, opened 2016-06-05
add commitConfiguration to stopSession

stantoncbradley:
fixes `'NSGenericException', reason: '*** AVCaptureSession can't stopRunning between calls to beginConfiguration / commitConfiguration'` that happens when unmounting component


stantoncbradley:
I was getting that error when transitioning to a different scene from the camera using react-native-router-flux.  this line fixed it for me

charpeni:
Thank you!

PR #318, opened 2016-06-10
bind this to _onBarCodeRead proxy

robbevan:
barcode read fails currently because 'this' is undefined

### DIFF
--- a/index.js
+++ b/index.js
@@ -142,7 +142,7 @@ export default class Camera extends Component {
   }
 
   async componentWillMount() {
-    this.cameraBarCodeReadListener = NativeAppEventEmitter.addListener('CameraBarCodeRead', this._onBarCodeRead);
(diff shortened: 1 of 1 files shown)


charpeni:
I think this method should be binded in the constructor.

chirag04:
huh? it's binding correctly using the arrow function. I don't think this change is required

robbevan:
Barcode read definitely *doesn't* work for me, fails with: undefined is not an object (evaluating '_this.props.onBarCodeRead')

chirag04:
what's your babelrc look like?

robbevan:
Ah, looks like passPerPreset was the culprit. Thanks.

chirag04:
if you using passPerPreset then you are def going to be hit by this: https://phabricator.babeljs.io/T7274
you might want to downgrade babel to 6.6

PR #319, opened 2016-06-14
Use dispatch_async when capturing still images (iOS)

tume:
Added one more dispatch_async call when capturing still images. If I understood correctly the code is already wrapped in one dispatch_async but the actual completion handler is called outside of it.

Noticed this when capturing photos with "photo" quality which are taking some time....and rendering progressbar without this patch does not really work because it will just freeze the whole ui.

Not sure if there is some problems with this approach but at least with my limited testing it seems to work and ui remains responsive enough for the progressbar.

Also the actual time that it takes to capture a photo seems to have increased slightly from the previous version I was using (don't have any real data to back this up, just eyeballing it) but that is probably another issue.


abhisuri97:
Could you write up a brief description of how to use it in the module? I am interested in using this for streaming to a URL.

tume:
Don't really understand @abhisuri97 how this change relates to streaming to a URL.

But would really like to get comments if this is reasonable approach or there is something else I could do? Really affects our app when trying to render something while waiting to take a picture.

Issue #76, opened 2015-07-26
Orientation sometimes fails to initialise correctly


JedWatson:
I'm seeing an intermittent error when the camera is initialised, where the orientation is incorrectly detected.

Using an iPad Mini 2 w/ iOS 8.4, starting in Landscape mode, it sometimes starts in Portrait mode. There doesn't seem to be a reliable way to detect the current orientation in React Native itself (?) otherwise I'd set the `orientation` prop, but my app can be used in either `landscapeLeft` or `landscapeRight` (locked in XCode)

Not sure if this is related to my project never allowing portrait orientation, that may mess with the detection?

Otherwise it looks like this may be a timing issue, it works properly about 80% of the time.

If I reorient the device, the camera fixes itself and stays fixed, until the next time the camera is initialised.

jtremback:
You're talking about the orientation of the viewfinder, right?

JedWatson:
@jtremback yes, that's correct.

lwansbrough:
Is this still an issue?

JedWatson:
It's still an issue, and actually a pretty major one for us, I just haven't been able to get someone with enough iOS / ObjectiveC experience to look into how to fix it yet :grimacing:

Any chance you could take a look? I can set up a project that replicates the issue if that will help.

lwansbrough:
I might have some time, no guarantees though. Pretty busy myself. AFAIK React Native doesn't have great support for horizontal display mode at this point. Haven't checked into that in a while though.

corbt:
Any workarounds @JedWatson? I'm running into this as well, which is a blocking issue because our app only records in landscape mode. I'm finding that consistently when the phone is already in landscape mode and I render the `Camera` the preview is in portrait orientation, until I wiggle the device a bit and it snaps to landscape. If I render the `Camera` when the phone is in portrait and then rotate the device, it behaves correctly.

Interestingly, if I set the `orientation` prop to any value except `auto`, including `landscapeLeft`, the preview is locked in the portrait orientation and never changes.

If I figure out a fix or workaround I'll post it here.

### Landscape phone with incorrect portrait preview
![img_0497](https://cloud.githubusercontent.com/assets/176426/12843065/c081cb8c-cbee-11e5-91c9-b1aa9b04a28e.jpg)

### Landscape phone with correct landscape preview
[...]

PR #333, opened 2016-07-02
Fix initial orientation setting, refs #76

nschoen:
Fixes a bug which is described in #76 and already had been fixed in the pull request #183 . It seems that the code from the pull requests has been removed by some merge.
I just copied and adjusted the solution of the user corbt and put it in the appropriate method.

Tested with my own app on an iPad Mini Retina device with iOS 9.0.2.

### DIFF
--- a/ios/RCTCamera.m
+++ b/ios/RCTCamera.m
@@ -64,7 +64,7 @@
 
 - (id)initWithManager:(RCTCameraManager*)manager bridge:(RCTBridge *)bridge
 {
-  
+
   if ((self = [super init])) {
     self.manager = manager;
     self.bridge = bridge;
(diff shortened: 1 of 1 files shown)


rt2zz:
would it make sense to always dispatch async the video orientation? also it seems that dispatch async still does not guarantee video orientation is supported so we should probably move if the if statement inside of the async block.

corbt:
I wrote this solution originally but it's not a good one because there's still a race condition. This only works if the set-orientation block gets into the async sessionQueue after startSession.

A better fix is to set the orientation at the end of the startSession block, as I do here: https://github.com/lwansbrough/react-native-camera/pull/288/files#diff-67aecadd1220a0eebd27fd0e61cd2b91R408

nschoen:
@corbt True, still running in this situation. Your solution sounds better. I tested it, works good for me since I only need the iOS version. I close this PR an checkout your branch until it's merged.

PR #354, opened 2016-08-01
Ensure that preview orientation is correct on launch

gethinwebster:
Currently, with orientation="auto", the orientation is not correctly set on first render. By moving the orientation change into the sessionQueue, we ensure that the previewLayer is correctly initialised before trying to change its orientation.


rt2zz:
👍

PR #360, opened 2016-08-08
fix examlpe jsCodeLocation

anativ:
Currently the example is working on IOS
The app is looking for "http://192.168.1.15" the default should be 127.0.0.1


rt2zz:
👍  :D

PR #368, opened 2016-08-13
Set the device orientation before instantiating RCTCamera

corbt:
Currently, there is a race condition in Android where if the camera is open while the device is in landscape mode, and the preview starts before the device orientation is registered, the preview appears stretched out vertically.

This PR ensures that the camera is always instantiated with the correct orientation, fixing the race condition.


jakegardner:
@corbt thanks for this!

PR #370, opened 2016-08-16
Android Barcode scanning

fbacker:
Added barcode scanner with zxing for Android.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -32,4 +32,5 @@ repositories {
 
 dependencies {
   compile "com.facebook.react:react-native:0.19.+"
+  compile 'com.google.zxing:core:3.2.1'
 }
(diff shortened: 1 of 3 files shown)


chirag04:
cc @rt2zz

jqn:
@fbacker Hi, I was wondering if this makes onBarCodeRead work on Android.

fbacker:
Yes it will.
zxing is slower than iOS scanner so using showViewFinder={true} will make the scanner only search inside the rectangle. 

Example for searching datamatrix for ios & android

    <Camera
      type={this.state.cameraType}
      captureAudio={false}
      captureTarget={this.cameraTypeParsed}
      captureQuality={this.state.quality}
      style={styles.camera}
  
      torchMode={this.torchModeParsed}
      showViewFinder={true}
      onBarCodeRead={this.barcodeReceived.bind(this)}
      barCodeTypes={['org.iso.DataMatrix,DATA_MATRIX']}
    />

Mostly force pushed this code to make barcode scanner on android it work in my project. 
An overhaul should be made so that everything worked exactly the same on Android and iOS.

Barcode types on android
MAXICODE
CODABAR
AZTEC
RSS_EXPANDED
ITF
QR_CODE
UPC_EAN_EXTENSION
CODE_39
CODE_128
EAN_8
UPC_E
CODE_93
PDF_417
UPC_A
DATA_MATRIX
RSS_14
EAN_13

PR #375, opened 2016-08-20
Update README.md

AzizAK:
Update README.md with the New Promises of ES7


rt2zz:
If you switch to async await you need to remove the .then and .catch

ptomasroos:
This is pretty outdated and any one wanting to use async await will probably figure out the syntax!

PR #380, opened 2016-08-24
prevent freezing when opened 2 times via navigator

matejkriz:
see http://stackoverflow.com/questions/33046061/


charpeni:
I can't test on iOS right now, if someone can tell us if it works.

Issue #367, opened 2016-08-11
Camera does not appear to work with Android 6 (Marshmellow)


cyphire:
### Steps to reproduce
1.  Emulator works under android 4
2. Emulator and actual android 6 device does NOT WORK, crashes on the render with <Camera>
3.

### Expected behaviour
Camera should open up in my app.

### Actual behaviour
Camera crashes the application.  Works in Genymotion 4.0 Android emulator, crashes in any 6.0 emulator or LG G4 (running android 6).

### Environment
- **Node.js version**:3.9.3
- **React Native version**: 0.31.0
- **React Native platform + platform version**: Android 6.0

### react-native-camera
**Version**: npm version or "master": Master

Also - IN any version the following occurs but may not be related:

My camera no longer works as well.

React Native 0.31.0
Android 6 does NOT WORK either in the emulator or on an actual device.  The <Camera> crashes the app with no explaination.

Android 4 (Emulator) still DOES work.

This is perplexing.

Also this code snippit says that the module is undefined:

Error: TypeError: undefined is not a function (evaluating '_reactNativeCamera2.default.checkDeviceAuthorizationStatus()')

```
  test_permissions() {
	try {

	  console.log(Camera.checkDeviceAuthorizationStatus());
	} catch(e) {
	  console.log('Error: ' + e);
	}
  }
```

anotherstarburst:
Getting the same error:


`java.lang.ExceptionInInitializerError
	at com.lwansbrough.RCTCamera.RCTCameraView.setActualDeviceOrientation(RCTCameraView.java:112)
	at com.lwansbrough.RCTCamera.RCTCameraView.<init>(RCTCameraView.java:28)
	at com.lwansbrough.RCTCamera.RCTCameraViewManager.createViewInstance(RCTCameraViewManager.java:17)
	at com.lwansbrough.RCTCamera.RCTCameraViewManager.createViewInstance(RCTCameraViewManager.java:7)
	at com.facebook.react.uimanager.ViewManager.createView(ViewManager.java:44)
	at com.facebook.react.uimanager.NativeViewHierarchyManager.createView(NativeViewHierarchyManager.java:211)
	at com.facebook.react.uimanager.UIViewOperationQueue$CreateViewOperation.execute(UIViewOperationQueue.java:148)
	at com.facebook.react.uimanager.UIViewOperationQueue$DispatchUIFrameCallback.dispatchPendingNonBatchedOperations(UIViewOperationQueue.java:890)
[...]

PR #381, opened 2016-08-24
Fixes issue #367

chadwilken:
Fixes issue #367 by checking if `System.console()` is null before attempting to use it to print. This is recommended by the Android docs.


cyphire:
Hey Chad, will that fix the issue with the camera not working under 6?

chadwilken:
It will fill the crash that you guys are seeing. It won't however make it
use the camera2 api. However as long as marshmallow isn't your build target
it will still work with the old camera api.

cyphire:
Cool I'll try that thanks so much!

corbt:
Why are we using `System.console` at all? Wouldn't Android's [`Log.e`](https://developer.android.com/reference/android/util/Log.html#e(java.lang.String, java.lang.String, java.lang.Throwable)) be more appropriate here?

PR #387, opened 2016-08-30
Fix iOS images always saved as Portrait

jakegardner:
The determination of device orientation using accelerometer values was not complete. Other issue reporters and our team found that the the accelerometer values fell outside the defined bounds, coincidentally always landing on portrait.

This PR updates the calculation and removes the subsequent image rotation during save (the intent of which I assume was to rectify the incorrect value from SensorOrientationChecker).

### DIFF
--- a/ios/RCTSensorOrientationChecker.m
+++ b/ios/RCTSensorOrientationChecker.m
@@ -71,18 +71,20 @@
 
 - (UIInterfaceOrientation)getOrientationBy:(CMAcceleration)acceleration
 {
-    if(acceleration.x >= 0.75) {
-        return UIInterfaceOrientationLandscapeLeft;
-    }
-    if(acceleration.x <= -0.75) {
+    float x = -acceleration.x;
+    float y = acceleration.y;
+    float angle = atan2(y, x);
+
+    if (angle >= -2.25 && angle <= -0.75) {
+         return UIInterfaceOrientationPortrait;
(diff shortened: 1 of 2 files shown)


ios/RCTSensorOrientationChecker.m
```diff
@@ -93,10 +95,11 @@ - (AVCaptureVideoOrientation)convertToAVCaptureVideoOrientation:(UIInterfaceOrie
             return AVCaptureVideoOrientationPortrait;
         case UIInterfaceOrientationPortraitUpsideDown:
             return AVCaptureVideoOrientationPortraitUpsideDown;
+        // Landscape UI and device orientation are opposite
         case UIInterfaceOrientationLandscapeLeft:
-            return AVCaptureVideoOrientationLandscapeLeft;
-        case UIInterfaceOrientationLandscapeRight:
             return AVCaptureVideoOrientationLandscapeRight;
+        case UIInterfaceOrientationLandscapeRight:
+            return AVCaptureVideoOrientationLandscapeLeft;
```
rt2zz: Why is *LandcapeRight mapped to *LandscapeLeft?

jakegardner: While testing orientations, I found that UIInterface and AVCaptureVideo did not match up Landscape values. With *LandscapeLeft = *LandscapeLeft, the saved image was upside down.

PR #393, opened 2016-09-01
Adjusted rotation should be that of camera, not display

corbt:
This minor fix causes videos recorded with the front camera in portrait orientation to have the correct orientation. Tested on Moto G and Nexus 9.


tadas-subonis:
When is this getting released?

tadas-subonis:
Guys?

PR #399, opened 2016-09-08
Mirror image for Android

gaguirre:
This PR will add Android support for `mirrorImage` option.
The image will be mirrored only if this options is set to `true` **AND** the picture was taken with the front camera.

As other properties this one can be set on the component declaration
```jsx
<Camera
  ref={cam => this.camera = cam}
  mirrorImage={true}
/>
```
or be passed as an option to the capture method
```javascript
this.camera.capture({ mirrorImage: true })
```

### DIFF
--- a/Example/Example.js
+++ b/Example/Example.js
@@ -189,6 +189,7 @@ export default class Example extends React.Component {
           type={this.state.camera.type}
           flashMode={this.state.camera.flashMode}
           defaultTouchToFocus
+          mirrorImage={false}
         />
         <View style={[styles.overlay, styles.topOverlay]}>
           <TouchableOpacity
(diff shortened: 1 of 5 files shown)


gaguirre:
This is complementary to iOS implementation added in #212

rt2zz:
@gaguirre thanks! why only mirror when using the front camera? Would be a cleaner api if it works for both cameras.

GeoffreyPlitt:
+1

gaguirre:
@rt2zz I don't know why the back camera picture mirroring results in a black image.
Also I thought mirroring would be useful only for front camera pictures, but as you said before it would be better to be consistent with iOS.

Anyway, I could debug it and see if I can make it work for back camera too. Give me one day or two :)

gaguirre:
Done! I think it was an issue with the emulator camera that wasn't active for the back one.

rt2zz:
Awesome!

PR #401, opened 2016-09-09
Added runtime check for supported bar code types

micksabox:
During my testing, I could not can data matrices. So I set a breakpoint in `RCTCameraManager` during the metadata capture, and realized the data matrix was not being properly set.

So I changed it to do a runtime check, and now it works.

### DIFF
--- a/ios/RCTCameraManager.m
+++ b/ios/RCTCameraManager.m
@@ -39,33 +39,41 @@ RCT_EXPORT_MODULE();
 
 - (NSDictionary *)constantsToExport
 {
+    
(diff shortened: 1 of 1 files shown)


charpeni:
Is that better than #412?

BeatMaster:
Experienced the same issue with the data matrix, and manually integrating the change fixed the issue. Would be really useful if this could be merged into the top-level. Would be happy to help but not sure what steps are required?

ptomasroos:
Can someone rebase this? Seems useful to pull in.

micksabox:
I believe #586 is the rebase for this

ptomasroos:
Thanks @micksabox Then lets close this!

PR #406, opened 2016-09-13
Update Readme

derdav3:
for everyone that links it manually  - see: https://github.com/lwansbrough/react-native-camera/issues/164


jacobSingh:
Hahah... I went to edit this file since I spent an hour on this and then saw it was changed and checked for ghosts in my machine.

Well, looks like we both found this.

Another piece of instructions, after installing, I think the emulator needs a restart.  Is this your expeirence?

derdav3:
😃  this also killed a good hour for me.. it's a good indicator if a pull request is required. :)

I had my test-device connected and just rebuild the project - don't know about the emulator..

PR #413, opened 2016-09-15
Android Barcode Scanner

plougsgaard:
I'm aware there are already attempts at solving this problem. However I couldn't get them working properly, so here's.. well.. another attempt!

## What

- Adds barcode scanner functionality to Android using the zxing library
- Only runs the scanner if an `onBarCodeRead` function is supplied
- Is really good at scanning if a small number of hints is supplied
 - See the `barCodeTypes` props
- Uses a simple lock to avoid flooding the runtime with scanner tasks
- Supports all iOS codes except `[code138, code39mod43, itf14]`
- Additionally supports `[codabar, code128, maxicode, rss14, rssexpanded, upca, upceanextension]`

## Test

Tested on a Nexus 5.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -32,4 +32,5 @@ repositories {
 
 dependencies {
   compile "com.facebook.react:react-native:0.19.+"
+  compile "com.google.zxing:core:3.2.1"
 }
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCamera.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCamera.java
@@ -17,6 +17,7 @@ public class RCTCamera {
     private final HashMap<Integer, CameraInfoWrapper> _cameraInfos;
     private final HashMap<Integer, Integer> _cameraTypeToIndex;
     private final Map<Number, Camera> _cameras;
+    private boolean _barcodeScannerEnabled = false;
     private int _orientation = -1;
     private int _actualDeviceOrientation = 0;
     private int _adjustedDeviceOrientation = 0;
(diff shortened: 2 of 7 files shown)


vinayr:
👍 👍 This is better than other PRs. Tested on 4.4.2 and 5.1
PS: setting `barCodeTypes` breaks on iOS. It's an existing issue, nothing to do with this PR.

plougsgaard:
@vinayr thanks for testing it, glad it works!

By the way, my test was on a Nexus 5 with Android 6.0.1.

chirag04:
cc @rt2zz

rt2zz:
LGTM, I am not in a position to test, but two 👍  sounds about right, lets do this ;)

Scimonster:
I just tested both this and react-native-barcodescanner on a few devices. Here are the results:

Success and failure are in terms of recognizing the barcode.

Device | rn-camera | rn-barcodescanner
----------|---------------|---------------------------
Samsung S7 Edge, Android 6.0.1 | Success | Success*
Samsung Galaxy Trend Plus, Android 4.2.2 | Fail | Success
[Otium S5][1], Android 4.4.2 | Fail | Success*
Nexus emulator (camera mapped to computer webcam), Android 4.4.2 | Fail | Fail

* Barcode scans correctly but it is very laggy and part of the screen actually remains black after navigating away from the camera.

Is there any way of improving these results from my side? `captureQuality` was tested both at `high` and `photo`. Or perhaps it's a problem with your code, that it either isn't able to parse a fuzzy picture, or it isn't properly focusing the camera.

I would like to use this library due it being much less laggy than the other one, but only if the most important thing i need it for actually works!

[1]: http://www.otiumobile.com/SmartPhones/S5_d6d17fab-b1e7-ec34-47e2-001035688ee2.html

chirag04:
@Scimonster If it's something you care about, I would say its best to send a PR.

Scimonster:
@chirag04 I can't; i a) don't know where the problem lies and b) don't do Java.

sibelius:
@Scimonster 

a) try to debug and find where the problem is
b) learn java

44326334:
@plougsgaard  I used QRcode base on `react-native-camera` . and the camera callback function `onBarCodeRead` result called fast , almost 10ms once. I readed your merged to used `barcodeScannerEnabled` to control . but I did not how to used it . please help me , thx!

PR #420, opened 2016-09-23
Default captureAudio to false

dijonkitchen:
Since default captureMode is a still picture, not video. 

This also works around some new iOS 10 permissions requiring notifying the user with a reason for using microphone (which is not good for the user experience if they're just taking a picture). Reference: https://github.com/lwansbrough/react-native-camera/issues/386#issuecomment-249007206

If you need video and sound, then explicitly pass it through as props in your Camera component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ The `aspect` property allows you to define how your viewfinder renders the camer
 
 #### `iOS` `captureAudio`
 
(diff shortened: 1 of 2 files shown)


dijonkitchen:
Updated readme!

charpeni:
Thank you, it's greatly appreciated.

PR #427, opened 2016-09-26
Document other iOS authorization status methods

ronaldheft:
The API already exposes specific methods for checking camera and microphone access individually. These should be documented. Especially since calling `checkDeviceAuthorizationStatus` requests the microphone permission, which is undesirable in photo-only use cases.


charpeni:
Thank you!

PR #436, opened 2016-09-29
Race conditions initializing flashMode

### LINKED ISSUE #433: Initializing with flashMode ignores value and uses the default
Using the latest build from master, if I initialize the camera with a set flash mode, it's ignore and the default value is used (off).

`<Camera flashMode="on" />`

I traced it back to `setFlashMode` in RCTCamera and there is check if the camera is null. When debugging, the camera is indeed null on initialization.

If I set the flashMode after a set delay (enough time for the camera to initialize), it indeed works.

ronaldheft:
Fixes #433

I found race conditions initializing `flashMode` on both iOS and Android. The issue can occur when a the `type` parameter is also specified, as the camera reinitialized. The parameters get set on the old instance of the camera and never get reinitialized on the new camera instance.

### DIFF
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCamera.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCamera.java
@@ -31,7 +31,7 @@ public class RCTCamera {
     }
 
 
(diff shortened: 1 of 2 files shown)


charpeni:
Thank you.

PR #441, opened 2016-10-01
Fix crash on second invocation of camera when running on iOS 10 simulator

ibussieres:
Fixes https://github.com/lwansbrough/react-native-camera/issues/426

Camera quality has no effect on simulator in the first place, so no problem ignoring it.


JoschaP:
thank u @ibussieres 👍

charpeni:
Thank you.

PR #449, opened 2016-10-06
Added CaptureQuality.preview

locly:
On Android because the available preview qualities can be different to the capture qualities you can end up in the situation where preview images are completely different sizes to captured ones (See [here](https://stackoverflow.com/questions/26958110/android-taken-photo-is-bigger-than-the-preview) for an example).

During normal take-a-picture use cases having a larger or different aspect ratio images isn't an issue. There are however cases where you want the taken image to match more closely to the preview image. An example of this is if you are providing some overlayed content on the image in the preview and then want to recompose that content onto the captured image. For example using the device to take a picture inside a photo frame...

This pull request adds the `CaptureQuality.preview` constant to Android to handle this use case. It does a fuzzy search across the available capture sizes in case there is a scenario where a preview size is not available in capture mode.

In our tests when taking pictures on iOS with this use case we haven't had any issue with image dimensions being wildly different so we've only added the PR for Android


jasongaare:
This is exactly what I am looking for and have been frustrated trying to solve. The sooner we can merge this, the better!

charpeni:
Can you fix conflicts?

locly:
Fixed the merge conflicts - should be good to go. Thanks!

charpeni:
Thank you!

PR #455, opened 2016-10-10
Fix .gitignore and add .npmignore

AndrewJack:
- remove Example/.gitignore
- add rules to root .gitignore 
- exclude Example from npm upload


charpeni:
Thank you!

PR #469, opened 2016-11-04
Add 1080p, 720p, and 480p capture qualities.

abrahambotros:
Add support for common specific resolutions (1080p, 720p, 480p). This allows selection of a specific resolution preset, in addition to the existing low/medium/high choices (which provide convenience but not quite as much control). Setting of capture/recording parameters are delegated to the native platforms according to their own setting profiles for each of these 3 resolution presets.


abrahambotros:
Looks like this will hopefully help with a few different issues: https://github.com/lwansbrough/react-native-camera/issues/447, https://github.com/lwansbrough/react-native-camera/issues/222, and https://github.com/lwansbrough/react-native-camera/issues/153.

Admittedly, I've only tested video on both Android and iOS (mainly using 480p and 720p in particular), and have not tested still-pictures (unable to get the Example app up and running at the moment I think due to out-of-date dependencies, etc.?). Any additional verification and testing by others here would be very much appreciated!

abrahambotros:
Added additional support for 540p in latest commits, too, with the following note (in the commit message and also the README):

> Note that Android does not currently offer a QUALITY_540P setting. We currently default to the base 480p settings, and then modify only the video size on top of this. Fix in future to apply all appropriate settings for 540p, not only raw video size.

Open to any better suggestions of what to do in the 540p case for Android, though.

abrahambotros:
As a side note, I played a bit with adding 540p also, but Android does not have a preset profile for 540p, and my device (Samsung Galaxy S7) doesn't support the 960x540 video resolution anyway. Decided to leave this out for now, but feel free to voice opinions. Thanks!

GeoffreyPlitt:
The 480p setting seems wrong to me. Though historically 480p referred to SD aspect ratios (i.e. 640x480), it's never used that way these days, especially with mobile phones whose screens are usually very close to HD aspects ratio of 16:9. The correct 480p setting for phones is 853x480, which maintains 480 lines but with an aspect ratio of 16x9.

See https://en.wikipedia.org/wiki/480p

abrahambotros:
@GeoffreyPlitt, thanks for catching this and pointing that out! Commits https://github.com/lwansbrough/react-native-camera/pull/469/commits/ee5f47e517a3ff86c74e25058f180af9b3df7c7a, https://github.com/lwansbrough/react-native-camera/pull/469/commits/a2da2ba629ca32db154d0d1c400a81068be86b19, and https://github.com/lwansbrough/react-native-camera/pull/469/commits/49d55d67dc754239b6bb1d73ec8e6f0534bbff09 should address this, setting the "ideal" 480p resolution to 853x480 on Android, and letting each device fall back to whatever it supports in that zone (on a related note, the 480p preset on Android is technically supposed to be 720x480 by default; see https://developer.android.com/reference/android/media/CamcorderProfile.html#QUALITY_480P).

On iOS, I'm leaving the 480p preset to map to 640x480 as specified in the docs (https://developer.apple.com/reference/avfoundation/avcapturesession/1669314-video_input_presets); I added a note to the README alerting people that this isn't close to the HD aspect ratio.

Let me know what you think.

GeoffreyPlitt:
Love it! Looks great, thanks, will test tomm.

larryranches:
@abrahambotros I just tested your last commit on both Android and ios. Everything is working great! Are you planning to more changes? If not, I think this PR should be ready to go 👍

abrahambotros:
@larryranches, thanks a bunch! Great to hear. Not planning on adding any more changes, so any maintainers/collaborators should feel free to merge as they see fit!

abrahambotros:
And @charpeni, thanks again for adding the proper label on this PR too!

charpeni:
Thank you, it's greatly appreciated.

abrahambotros:
Thanks again for handling the merge and closing the related issues @charpeni!

Issue #461, opened 2016-10-22
Deadlock when rotating app


johnryan:
### Steps to reproduce

Create a camera view and rotate the device several times.
### Expected behaviour

The preview should rotate with the app.
### Actual behaviour

Eventually the preview layer gets locked in RCTCamera.m -layoutSubviews and the main thread becomes locked. I've attached a screenshot of the stack trace. I was able to solve the problem by changing layoutSubviews but it seems like a hack:

```
- (void)layoutSubviews
{
  [super layoutSubviews];
    dispatch_async(dispatch_get_main_queue(), ^{
        self.manager.previewLayer.frame = self.bounds;
        [self setBackgroundColor:[UIColor blackColor]];
        [self.layer insertSublayer:self.manager.previewLayer atIndex:0];
    });

}
```

<img width="344" alt="screen shot 2016-10-22 at 4 20 16 pm" src="https://cloud.githubusercontent.com/assets/416564/19622248/70beb95a-9873-11e6-967f-5396f23a3f3d.png">
### Environment
- **Node.js version**: 6.8.0
- **React Native version**: 0.35
- **React Native platform + platform version**: iOS 9.3.5
### react-native-camera

**Version**: master

Anuj-786:
I think you should have to use orientation i.e 
Values: `Camera.constants.Orientation.auto or "auto" (default), Camera.constants.Orientation.landscapeLeft or "landscapeLeft", Camera.constants.Orientation.landscapeRight or "landscapeRight", Camera.constants.Orientation.portrait or "portrait", Camera.constants.Orientation.portraitUpsideDown or "portraitUpsideDown"`

johnryan:
I haven't changed any of the values so it is set on the default (auto)  and as such should rotate with the rest of the app. It works sometimes but eventually locks up and freezes the app.

GeoffreyPlitt:
I have a stacktrace which may help:

java.lang.RuntimeException·getParameters failed (empty parameters)
Raw
Camera.java:-2android.hardware.Camera.native_getParameters	
Camera.java:1996android.hardware.Camera.getParameters	
RCTCamera.java:332com.lwansbrough.RCTCamera.RCTCamera.adjustPreviewLayout	
RCTCamera.java:169com.lwansbrough.RCTCamera.RCTCamera.setActualDeviceOrientation	
RCTCameraView.java:122com.lwansbrough.RCTCamera.RCTCameraView.setActualDeviceOrientation	
[...]

PR #470, opened 2016-11-04
Fix issue #461

egpast:
Hi. I don't have much experience with iOS code, but this issue is affecting a project I'm involved with that depends on this module, and I believe I have a fix that works and makes sense.

I think the issue described in #461 was caused by the fix for #354, which moved preview orientation updates off of the main thread. [Apple documentation suggests that most UI updates are thread-unsafe and should be run on the main queue](http://stackoverflow.com/a/6087096), and this issue seems to confirm that - orientation deadlock always occurs between the `previewLayer` updates in `layoutSubviews` and `changePreviewOrientation`, as shown below. This deadlock also stops if the #354 fix is reverted.

Deadlock thread details

@johnryan's comment introducing issue #461 contained a workaround moving the `layoutSubviews` `previewLayer` update to the main queue, but it didn't work for me (on an iPhone 6 running iOS 10), and I think that makes sense - the main queue and the manager's `sessionQueue` need not run on the same thread, so they can still deadlock. However, moving **both** `previewLayer` updates to the main queue serializes them properly. I did also try just moving the `layoutSubviews` update to the `sessionQueue`, but this resulted in unacceptably slow resizing, often lagging multiple seconds behind device reorientation.

In this commit, I've kept the `changePreviewOrientation` `previewLayer` update enqueued in the `sessionQueue` as well, so #354 should be unaffected by these changes. The changes have been tested on the following phones so far:

- iPhone 6, iOS 10.0.2
- iPhone 6S, iOS 9.3.5
- iPhone 6S Plus, iOS 10.1.1

### DIFF
--- a/ios/RCTCamera.m
+++ b/ios/RCTCamera.m
@@ -182,9 +182,11 @@
 - (void)changePreviewOrientation:(NSInteger)orientation
 {
     dispatch_async(self.manager.sessionQueue, ^{
-        if (self.manager.previewLayer.connection.isVideoOrientationSupported) {
-            self.manager.previewLayer.connection.videoOrientation = orientation;
-        }
+        dispatch_async(dispatch_get_main_queue(), ^{
+            if (self.manager.previewLayer.connection.isVideoOrientationSupported) {
+                self.manager.previewLayer.connection.videoOrientation = orientation;
+            }
+        });
     });
 }
 


rt2zz:
IF someone can weigh in on this, that would be greatly appreciated. I am all for solving the underlying bug, but having a double dispatch_async strikes me as odd. Thoughts?

johnryan:
@rt2zz do you know why the sessionQueue exists in the first place? Looking at the code it looks like its there just to synchronize the setting of the properties. If sessionQueue was replaced with async dispatches to the main queue it would effectively create a serial queue and then only one dispatch_async would be required.

egpast:
@rt2zz, I think the necessity of the `sessionQueue` dispatch would be something to ask @gethinwebster about; I avoided changing that at all simply because it wasn't clear to me from the #354 PR whether the timing issue discussed there was resolved by moving to _a_ queue or to _that_ queue specifically. As far as I know, I never saw the described issue on my test devices, even when I briefly rolled that commit back, so I didn't feel confident reasoning about an alternative implementation on my own.

Your comment got me thinking about all the dispatches in this PR, though, and I think the one in `layoutSubviews` might be superfluous. My stack trace above suggests that that function is already running on the main thread, so putting it on the main queue is likely just delaying it.

I don't have access to my test iPhones at the moment, but I should be able to check the above idea - and maybe get a better handle on what #354 was fixing - tomorrow, or Monday at the latest.

egpast:
It looks like the `sessionQueue` dispatch **is** necessary, but the `layoutSubviews` dispatch is not, as expected. I've updated the commit to reflect this.

I probably should have tried a little harder to replicate #354 before, since it's really not that unintuitive - without the fix, the camera always starts in portrait orientation, regardless of device orientation. This still happens if you dispatch `changePreviewOrientation` on the main queue, but the `sessionQueue` dispatch in the actual fix stops it. So I'd say the double dispatch is necessary unless there's some other, better way to sufficiently delay `changePreviewOrientation`.

PR #479, opened 2016-11-17
Decreasing video recoding size for low quality option

filipecosta01:
Setting the audio and video encoding instead of using the default ones seems to decrease the size of the video record file.
Tested some cases and using Motorola G2 with low quality option, video record size was 4.7MB (10 seconds length). After improvements, the video size has like 190KB.

As for some people is important to upload the video and a low quality is really necessary, maybe it would be useful for community.

Waiting for some feedback, if you really think this would be a nice to have.


android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
```diff
@@ -501,13 +501,43 @@ private void releaseMediaRecorder() {
...
+
     private byte[] mirrorImage(byte[] data) {
         ByteArrayInputStream inputStream = new ByteArrayInputStream(data);
         Bitmap photo = BitmapFactory.decodeStream(inputStream);
 
         Matrix m = new Matrix();
         m.preScale(-1, 1);
-        Bitmap mirroredImage = Bitmap.createBitmap(photo, 0, 0, photo.getWidth(), photo.getHeight(), m, false);
+
+        Bitmap mirroredImage = Bitmap.createBitmap(photo, 0, 0, photo.getWidth(), photo.getHeight(), m, true);
```
rt2zz: what does this false -> true change do?

rt2zz:
Can you explain why the image rotation changes are needed? rotation has been a pain point in the past and I want to make sure any related changes are thought through and maintainable.

filipecosta01:
Think I should run more tests for the image rotation, cause I'm using the tablets front-face camera and for this user case sometimes the correct image is created in landscape instead of portrait (as I said, tablets front-face camera only).

PR #480, opened 2016-11-18
Update example

rt2zz:
@charpeni I took the liberty to merge your PR https://github.com/lwansbrough/react-native-camera/pull/422 with master and resolve the conflict. I think this is gtg (if a bit out of date at this point :) 

I couldnt figure out how to PR it back to your repo, let me know if we should merge or we can port this change back to your branch first.


charpeni:
Should be fine like this :)

PR #488, opened 2016-11-26
Include microphone permission key in Info.plist

antonshevchenko:
Mention microphone permission (and what to add to `Info.plist) for iOS 10 and up.


charpeni:
Thank you!

Issue #475, opened 2016-11-14
android image rotation is wrong (mirrorImage=true, orientation=portait)


madsleejensen:
### Steps to reproduce
1. Take a photo with Front camera, with `mirrorImage=true` `orientation=portait`

### Expected behaviour
The image should be captured as the preview view displays.

### Actual behaviour
The image is rotated 90deg. The expected behavior would be the same rotation as `mirrorImage=false` but flipped horizontal.

### Environment
- react-native v0.34.0
- **React Native platform + platform version**: Android 6.x (Samsung) 

### react-native-camera
**Version**: master

jarradh:
We're experiencing the same behaviour

ospfranco:
Hey! this repository was recently transferred to the react-native-community, as part of our ongoing cleaning process we are closing old tickets (there is just too many of them!), please feel free to try the latest versions or the master branch to see if your problem has been solved.

If you are still experiencing this problem feel free to re-open this or fill a new ticket.

Issue #309, opened 2016-06-02
Android photo taken in portrait gets rotated when saved to camera roll


daavidjobe:
### Steps to reproduce

1.Capture an image on a Samsung device in portrait mode and save it to camera roll
### Expected behaviour

The image should be saved as portrait
### Actual behaviour

The image is saved as landscape

This could be fixed by rotating the image if device orientation is 0 and device model is samsung but it´s a bad solution. I don´t know if the issue exists on all samsung devices or just the ones I have tested it on. Also, rotating a bitmap makes the module slow
### Environment
- **Node.js version**:
- **React Native version**: 0.26.2
- **React Native platform + platform version**: Android 5.1
### react-native-camera

**Version**: master

abalhier:
Same issue here on a Samasung Galaxy S5. Same RN version.
I don't have another device to try this on though :/

daavidjobe:
Forgot to mention what models I´ve tested it on. I´m experiencing this on Galaxy S4 and S6. I think this ultimately comes down to a hardware issue. I don´t think rotating the images on these devices is the right solution since it is very slow and the maintainers would not accept a PR of that kind.

nicastelo:
Same issue on LG H340 Leon

daavidjobe:
I found a solution that suit my needs at least. I changed the captureTarget to temp and then saving it in my own Java code. This way i can show the temporary image while I´m rotating and what not. The bad part is that I have to maintain some sort of blacklist of models this problem effects.

abalhier:
What I ended up doing was to retrieve camera orientation and expose it via a React module like this :
```
    static {
        for (int i = 0; i < Camera.getNumberOfCameras(); i++) {
            Camera.CameraInfo info = new Camera.CameraInfo();
            Camera.getCameraInfo(i, info);
            if (info.facing == Camera.CameraInfo.CAMERA_FACING_FRONT) {
                FRONT_CAMERA_ORIENTATION = info.orientation;
            } else if (info.facing == Camera.CameraInfo.CAMERA_FACING_BACK) {
                BACK_CAMERA_ORIENTATION = info.orientation;
            }
        }
    }

    @Nullable
    @Override
    public Map<String, Object> getConstants() {
        return ImmutableMap.<String, Object>of(
                "BACK_CAMERA_ORIENTATION", BACK_CAMERA_ORIENTATION,
                "FRONT_CAMERA_ORIENTATION", FRONT_CAMERA_ORIENTATION
        );
[...]

PR #493, opened 2016-11-29
Android orientation fix (#309, #475)

alwx:
This PR fixes the problem with orientation on some devices. This problem is described in #309, and the reason of it is that some devices save images without any rotation, simply modifying the EXIF metadata instead.

It is fine when you need to save your image to gallery (because gallery app knows how to read EXIF and display images), but it is definitely not okay if you save your image to memory and/or want to reuse somehow (for example, we convert it to base64 after capturing).

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -33,4 +33,5 @@ repositories {
 dependencies {
   compile "com.facebook.react:react-native:0.19.+"
   compile "com.google.zxing:core:3.2.1"
+  compile "com.drewnoakes:metadata-extractor:2.9.1"
 }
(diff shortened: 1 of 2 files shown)


markos85:
Hi, thanks for the help. Can we have some quick review and merge of this one ? :)

charpeni:
@alwx Any breaking changes ?

I can't test this right now, someone can tell me if everything works as expected ?

alwx:
@charpeni no breaking changes :) I've tested it on different devices using AWS Device Farm — everything works. But, of course, one more confirmation of its stability and correctness won't be superfluous :)

charpeni:
Thank you! :)

radko93:
It crashes on Galaxy S7 (Android 6.0). It seems that there's no `ExifIFD0Directory` in the metadata.

npomfret:
@gmahomarf are you able to test my PR?  It may help with this https://github.com/lwansbrough/react-native-camera/pull/682.

satyendrakannaujiya:
I also have the same issue

n1ru4l:
@satyendrakannaujiya RCTCamera is deprecated, please try RNCamera.

shaktalles:
Having same issue.

n1ru4l:
@shaktalles Please open a new issue with the issue template

PR #497, opened 2016-11-30
Update README.md

aneudy1702:
added a missing piece of information on getting ios working based on https://github.com/lwansbrough/react-native-camera/issues/3.

The whole installation does not work without this piece


charpeni:
It is still relevant in the latest version?

sibelius:
I don't think so

charpeni:
I'll close this pull request as we think it isn't relevant in the latest version. If we're wrong, please let use know.

PR #504, opened 2016-12-07
Update RCTCameraManager.m

jakelaw311:
Camera crashes on iPhone when invoked a second time. This change seems to fix it.


charpeni:
The fix you removed should fix this issue. #380 f4bf30e17b78c89d834625514929956af9d0336f

@matejkriz Any thoughts?

matejkriz:
@jakelaw311 the camera crashes every time for you with that condition? which iPhone and iOS?

I'm pretty sure, that my fix is not good, but I don't know enough Objective-C to make it properly.

sibelius:
is this still happening?

PR #520, opened 2016-12-16
[Android] Added orientation hint calculation for video.

gmahomarf:
This should allow video to be recorded in any orientation, regardless of screen rotation being enabled or disabled.


prasann278:
i have cloned full project but it is not working for me plz help me 

need to access camera on a button click in react native android

PR #523, opened 2016-12-19
[iOS] set camera pointer to nil when stopping session in emulator

callmephilip:
We are seeing a weird behavior when using the camera component + navigation running in iOS emulator (similar to what is described in https://github.com/lwansbrough/react-native-camera/issues/512).

This PR addresses this by performing a consistent clean up even when running on the emulator


mankins:
I was just about to submit the same PR. I've verified this fixes https://github.com/lwansbrough/react-native-camera/issues/452 ... thanks!

PR #531, opened 2016-12-23
Fix Android crashes

Fweeper:
This fixes two Android crashes.

The first crash is caused by a race condition in barcode scanner mode.  When `onPreviewFrame()` is called, which happens multiple times per second, it fires off an async task which calls `getParameters()` on the Camera object.  There is always a delay between `onPreviewFrame()` and the async task; usually it's a millisecond or less, but I've seen it go as high as 800 milliseconds.  If the Camera object is released in that delay window (e.g. if the user pauses the app), the call to `getParameters()` will throw an exception that crashes the app.  The fix for this is to simply catch and ignore that exception.

The second crash is similar, being caused by a call to `getParameters()` on a released, or possible `null`, Camera object.  This one happens if the Camera has been released but the viewfinder is still active; for example, when switching between front and back cameras.  If the user then initiates a touch event in the viewfinder, an exception will be thrown that crashes the app.  Again, the fix is to catch and ignore that exception.

(The change in `startCamera()` was not actually required for either fix, but was made simply for completeness and consistency.)

### DIFF
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraViewFinder.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraViewFinder.java
@@ -43,6 +43,16 @@ class RCTCameraViewFinder extends TextureView implements TextureView.SurfaceText
     // reader instance for the barcode scanner
     private final MultiFormatReader _multiFormatReader = new MultiFormatReader();
 
+    private static Camera.Parameters getCameraParameters(Camera camera) {
+        try {
+            return camera != null ? camera.getParameters() : null;
+        }
+        catch (RuntimeException e) {
+            // The camera has been released
+            return null;
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


burgalon:
This seems to fix the crash, but breaks the barcode scanning

Fweeper:
@burgalon, what specifically is breaking for you?  We've been using this commit in our RN app, which relies heavily on barcode scanning, for several weeks with no issues.

freiserg:
Yes, we have same crash with getParameters.

burgalon:
@Fweeper for some reason the barcode scan callback didn't trigger

Fweeper:
@burgalon could you supply a code example in which the barcode scanning does not work?  Here is an example that is working just fine for me:

```
<RNCamera
	style={{ flex: 1 }}
	barCodeTypes={[ 'org.iso.DataMatrix' ]}
	onBarCodeRead={ (e) => console.log(e.data) }
/>
```

PR #548, opened 2017-01-06
Add bounds (origin, size) data to Android barcode event

jmfirth:
This patch provides boundary data to Android module barcode read event, as is already provided by the iOS module.  I doubt the boundary data is perfectly accurate between platform modules but is useful enough for many contexts.


melihmucuk:
@jmfirth I think calculation should be left to right and top to bottom. This is reversed. Forexample, If I scan barcode on right side of the screen, x becomes 10. But it should be width - 10 basically.

PR #559, opened 2017-01-18
resolve promise in onScanCompleted callback; fixes #558

### LINKED ISSUE #558: Capture result on android not consistent with CameraRoll
### Steps to reproduce
1. Invoke `capture()` method with a `cameraRoll` target. 

### Expected behaviour
The `path` returned should match the `uri` of the objects returned by RN's `CameraRoll.getPhotos()` API.  Should be able to retrieve captured photo object with matching `uri` using `CameraRoll.getPhotos()` immediately after the promise returned by `capture()` is resolved.  This works as expected on iOS.

### Actual behaviour
The `path` in the `capture()` result is the external storage URI, not the URI exposed through the `CameraRoll`, and thus cannot be used to cross reference those objects and retrieve the one captured.  Also, there is an indeterminate delay after the `capture()` promise is resolved before the object appears in the `CameraRoll`.

### Environment
- **React Native version**: 0.40.0
- **React Native platform + platform version**: Android 6.0.1

### react-native-camera
**Version**: `next` branch

I have a fix for this that I will submit a PR for shortly.

danieldkim:
Fixes #558.

### DIFF
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
@@ -425,7 +425,7 @@ public class RCTCameraModule extends ReactContextBaseJavaModule
         f.setReadable(true, false); // so mediaplayer can play it
         f.setWritable(true, false); // so can clean it up
 
-        WritableMap response = new WritableNativeMap();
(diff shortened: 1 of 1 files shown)


ptomasroos:
Will look into this during the week. Thanks

PR #563, opened 2017-01-23
Fix barcode type 'code128' typo

Hybrid-Force:
Fix a typo in barcode types. The type should be `code128` instead of `code138`.


rt2zz:
tyty

PR #564, opened 2017-01-23
Update barcode read event listener when `onBarCodeRead` prop updates

Hybrid-Force:
Previously, if the initial `onBarCodeRead` prop is undefined, then it will never be updated even when new `onBarCodeRead` prop is provided.

After the fix, this event listener will be updated when new `onBarCodeRead` prop is received.

### DIFF
--- a/index.js
+++ b/index.js
@@ -171,7 +171,7 @@ export default class Camera extends Component {
 
   componentWillReceiveProps(newProps) {
     const { onBarCodeRead } = this.props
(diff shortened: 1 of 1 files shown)


andyhite:
👍 to this. I ran into the same problem when using the barcode scanner with a function supplied by `mapDispatchToProps`.

ptomasroos:
Thanks @Hybrid-Force !

PR #577, opened 2017-01-30
fix header search paths

chirag04:
seems like one header path was not updated. Cleaned up the xcode config as well.

cc @rt2zz


rt2zz:
❕

chirag04:
awesome. @rt2zz can you also bump the version on npm?

PR #578, opened 2017-02-02
iOS: minor focus and zoom changes

abrahambotros:
RCTCameraManager focusAtThePoint improvements
- Dynamicaly get size from bounds of camera view, instead of full screen, since
  it might not always be the case that the Camera react component is taking up
  the entire device screen.
- Also add setting exposure to the same point of interest we are focusing on.

Example
- Add onFocusChanged and onZoomChanged empty callback functions by default to
  Example app, allowing tap-to-focus and pinch-to-zoom to be readily
  experienced/experimented with.
- Updated react/react-native dependencies to match root package.json.

TODO/Other remarks
- Tap-to-focus seems to still not work perfectly... From logging, it always seems
  to get the right location in the view (meaning its getting the right location
  from the user touch and transforming it to the {0, 1} range appropriately), and
  does indeed engage the focus process, but it seems it sometimes refocuses on the
  center-ish region of what's in the camera instead of the location that was actually
  pressed. I thought this might be related to the subjectAreaDidChange getting called,
  which in turn sets the focus mode to continuous auto-focus at the view center, but
  from my experimenting, this method never actually gets called. I wasn't able to
  figure out if there's somewhere else in the library that's forcing continuous auto-focus,
  or if there's just some bug in our current focus procedure within focusAtThePoint.

### DIFF
--- a/Example/Example.js
+++ b/Example/Example.js
@@ -182,6 +182,8 @@ export default class Example extends React.Component {
           captureTarget={this.state.camera.captureTarget}
           type={this.state.camera.type}
           flashMode={this.state.camera.flashMode}
+          onFocusChanged={() => {}}
+          onZoomChanged={() => {}}
           defaultTouchToFocus
           mirrorImage={false}
         />
--- a/Example/package.json
+++ b/Example/package.json
@@ -6,8 +6,8 @@
     "start": "node node_modules/react-native/local-cli/cli.js start"
   },
   "dependencies": {
-    "react": "~15.3.0",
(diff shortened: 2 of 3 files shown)


flybayer:
These changes are working well for me. Thanks!

abrahambotros:
Resetting `Example/package.json` React/React-Native dependency versions to `master` settings, delegating that change to https://github.com/lwansbrough/react-native-camera/pull/679.

And thanks @flybayer for the testing and support! Much appreciated!

PR #582, opened 2017-02-02
Android: mute on recording start/stop when playSoundOnCapture=false

abrahambotros:
- Immediately before recording start/stop, mute all relevant system audio streams,
  saving their previous volumes. Only applies when playSoundOnCapture=false.
- Immediately after recording start/stop, restore all relevant system audio streams
  to their saved volumes. Only applies when playSoundOnCapture=false.


abrahambotros:
Note that this has trivial merge conflicts with https://github.com/lwansbrough/react-native-camera/pull/575; if/when the first of these two gets merged, I'll resolve conflicts so the second can merge without issue.

PR #583, opened 2017-02-03
Android: mute on recording start/stop when playSoundOnCapture=false

abrahambotros:
- Immediately before recording start/stop, mute all relevant system audio streams;
  after a small delay, unmute them.

Note that this has trivial merge conflicts with https://github.com/lwansbrough/react-native-camera/pull/575; if/when the first of these two gets merged, I'll resolve conflicts so the second can merge without issue.


abrahambotros:
Additionally, I was thinking, maybe we should just mute the system audio streams throughout the entire duration of the video recording? So mute, then start capture, then capture, then end capture, then unmute. This way, no system audio gets into the video audio, and in general the user isn't disrupted by sounds while recording; the downside is that this means the user won't be alerted audibly by system sounds during (which might be ok, I think most Android phones do this on the native camera?). Any thoughts and input welcome!

PR #595, opened 2017-02-13
Fix for rotation on front facing images with mirror set to true

Thomas101:
This fixes a use case where mirroring is enabled and the front camera is being used. Previously the image was not rotated correctly because the mirror call strips the exif data out.

This PR moves the `fixOrientation` call above the `mirrorImage` call to ensure the exif data is available to the rotation call

### DIFF
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
@@ -663,14 +663,14 @@ public class RCTCameraModule extends ReactContextBaseJavaModule
             @Override
             public void onPictureTaken(byte[] data, Camera camera) {
 
+                data = fixOrientation(data);
+
                 if (shouldMirror) {
                     data = mirrorImage(data);
                     if (data == null) {
                         promise.reject("Error mirroring image");
                     }
                 }
-
-                data = fixOrientation(data);
 
                 camera.stopPreview();
                 camera.startPreview();


ptomasroos:
LGTM
@rt2zz ?

mikaello:
I got the rotation bug on front facing camera with `mirror=true` on my Samsung S4 Active, this PR solves the problem for me on that device.

I also got an LG G4, front facing image with mirror works both before and after this fix.

Thomas101:
It looks like the work that @npomfret has done on 9fd9d1f8b181eb1ba162c19f52d43e8ea492e285 takes this into account and is no longer an issue. Will close this off. Awesome, thanks!

npomfret:
Hmm, i definitely didn't intentionally fix that.  what change in behaviour are you seeing?

Thomas101:
Previously the image was not rotated correctly because the exif data was discarded when mirroring. The fix I suggested was very much a tape-over-the-cracks fix by changing the ordering of execution to avoid this. In your MutableImage class you're keeping the original image data and pulling the exif data from this for manipulation as and when needed. Much more elegant and resilient in my opinion, so a giant thanks :-)

npomfret:
Ah i see. Good stuff.

I wonder if the rotation correct stuff is required now that the exif data is preserved?

Thomas101:
I would argue so. I don't think ReactNative takes into account exif data when displaying. If it was to be removed great to keep it as an option!

npomfret:
We could pass in a `{forceRotationCorrection: true}` (which defaults to false) option or something like that for android users.  Maybe remove it completely at a later date?

npomfret:
Actually, maybe we could consider removing it now.  I can't think of a reason to keep it.

PR #601, opened 2017-02-15
Encourage use of newest version (0.6)

mikemonteith:


rt2zz:
👍 we can probably drop this note altogether soon

PR #607, opened 2017-02-20
Added compatibility with Exponent

syousif94:
I made this compatible with a detached a Exponent project via `npm install` and `react-native link`

Feel free to close and ignore if there's a better approach.


ptomasroos:
I'm not using exponent and will have no ability to test stuff through exponent. Plus changing the header search paths, is that really something needed for exponent support? Not my 5 cents, but feels quite low level requirement

PR #615, opened 2017-02-27
Small, but important changes

NazarYablonskiy:
- Removed
```
    <uses-permission android:name="android.permission.RECORD_AUDIO" />
    <uses-permission android:name="android.permission.RECORD_VIDEO" />
    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
```

- Updated example
- Updated README.md

##### Explanation
My company uses this module only for scanning of barcodes. Some users blame us in espionage (at least, I think so (according to user feedback)) because the app requires an access to a microphone.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "type": "git",
     "url": "https://github.com/lwansbrough/react-native-camera.git"
   },
-  "version": "0.6.0",
+  "version": "0.6.1",
   "description": "A Camera component for React Native. Also reads barcodes.",
   "author": "Lochlan Wansbrough <lochie@live.com> (http://lwansbrough.com)",
   "peerDependencies": {
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "type": "git",
     "url": "https://github.com/lwansbrough/react-native-camera.git"
   },
-  "version": "0.6.1",
+  "version": "0.6.0",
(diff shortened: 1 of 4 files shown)


ptomasroos:
This is a quite breaking change. Personally I don't have a problem with changing this behaviour. 
But I'll let someone more provide feedback

chris-at-translate:
In theory remove all permissions and then ask for the permissions when they are needed is the better way to go.

ptomasroos:
@chris-at-translate that only works for devices with runtime permissions.

package.json
```diff
@@ -4,7 +4,7 @@
     "type": "git",
     "url": "https://github.com/lwansbrough/react-native-camera.git"
   },
-  "version": "0.6.0",
+  "version": "0.6.1",
```
rt2zz: remove this change, we will cut a new release separately

rt2zz:
I like these changes, and agreed while it is a fundamentally breaking change, now is the time to make it. Can you add a section near the top of the docs explaining the change and why we made it.

NazarYablonskiy:
@rt2zz done

ptomasroos:
LGTM!

PR #619, opened 2017-03-04
Fix crash when not enough disk storage to save media

flybayer:
I had my app crash when storage was full because `assetURL == nil` in this case.

### DIFF
--- a/ios/RCTCameraManager.m
+++ b/ios/RCTCameraManager.m
@@ -831,7 +831,14 @@ didFinishRecordingToOutputFileAtURL:(NSURL *)outputFileURL
                                     if (error) {
                                       self.videoReject(RCTErrorUnspecified, nil, RCTErrorWithMessage(error.description));
                                       return;
+                                    } else if (assetURL == nil) {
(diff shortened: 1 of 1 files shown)


ptomasroos:
Nice find, but can we be certain that its the only case when assetUrl comes back as nil?

flybayer:
@ptomasroos from my research, I can't say for sure that it won't be nil in any other case, but large majority of reports are because of low storage.

npomfret:
Could we do the same for images?

bryanberger:
reproduced this today.
```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSDictionaryM setObject:forKey:]: object cannot be nil (key: path)'
*** First throw call stack:
```

sibelius:
@bryanberger is this error because of this PR?

or this PR fixes your error?

bryanberger:
@sibelius this PR fixes the error I had.

sibelius:
Great to hear that, let’s merge this

Tks for the help

PR #620, opened 2017-03-05
let flexbox resize components

duhaime:
This PR introduces a small change to fix a small styling bug in the current README.

The README currently uses `Dimensions.get('window')` to set the height/width of the camera preview. This looks great in portrait mode, but if one rotates a physical device to landscape mode, the preview only occupies half the device screen.

By allowing flexbox to resize the view, this PR makes it so that the preview occupies the full width of the device screen.


rt2zz:
ty

PR #627, opened 2017-03-09
feat(ios): add support for `preview` to match viewport cropping

Ignigena:
Currently `Camera.constants.CaptureQuality.preview` is only available on Android. This is an attempt to bring similar functionality to iOS when a non-standard viewport aspect ratio is being used.

The functionality differs slightly from the Android implementation. Whereas the Android implementation appears to choose the native camera preset closest to the viewport size this implementation simply crops the resulting image after capture to match the viewport aspect ratio.

Due to this difference, the capture always happens using `AVCaptureSessionPresetPhoto` which is great for photos but likely not ideal for videos.

In any case this is at least a starting point towards having similar functionality on iOS. Open for review and consideration.

### DIFF
--- a/ios/RCTCameraManager.h
+++ b/ios/RCTCameraManager.h
@@ -16,7 +16,8 @@ typedef NS_ENUM(NSInteger, RCTCameraCaptureSessionPreset) {
   RCTCameraCaptureSessionPresetPhoto = 3,
   RCTCameraCaptureSessionPreset480p = 4,
   RCTCameraCaptureSessionPreset720p = 5,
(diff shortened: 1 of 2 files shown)


ptomasroos:
Will take a look at this ASAP. Thanks for the effort

mirrorshades:
@ptomasroos Why was this never merged?

ptomasroos:
That's a very valid question. Think it was forgotten. Can you debase this @ignigena

Ignigena:
@ptomasroos Rebased against current master branch!

sibelius:
I think it looks good

PR #635, opened 2017-03-19
adding location meta data to iOS camera roll

npomfret:


npomfret:
I could figure out how the location metadata described in the readme was supposed to work. I've added this change which writes the GPS information supplied to the iOS camera roll.  The result is that iOS images are tagged with GPS data as if they had been taken with the native camera app.  

Here's a bit of logging from my sample app to show what I men.  After taking the image the app then uses the new image URI to find the corresponding image in the CameraRoll.  It then console.logs it:

screen shot 2017-03-19 at 12 48 02

npomfret:
I think I can get the same working with android (so the image saved in the camera roll has the location added).  But as it's my first PR on the this I'd like to focus on one at a time.

rt2zz:
❤️ @npomfret 

Is any downside to automatically including location metadata? I am thinking we probably need a prop to toggle collections of location metadata as this can be potentially sensitive.

jtremback:
I haven't used or looked at this code for 2 years, but I saw this on my notifications and I was just curious: Isn't this the same thing I added in this commit from 2015? 

https://github.com/lwansbrough/react-native-camera/commit/a4bb5bbfaa0eceec8a52289a0ccc007a5f393559?diff=unified

npomfret:
@jtremback - hi, i didn't see you code, and yes it looks very similar.  I don't understand why the location metadata wasn't being added if your code is already in place?

npomfret:
Clearly this has already been done, for iOS at least.  Must be a problem with the way i'm using the library.

thanks

jtremback:
Maybe there has been a regression that's causing it not to work?

npomfret:
No, it was my fault.  Didn't understand the instructions, think maybe an example would be useful.

However, it doesn't seem to work on android (as far as I can see).  I'll submit another PR with an android fix.

jtremback:
Yea when I contributed, React Native did not yet support Android.

koswarabilly:
@npomfret 
i have this code in my project

```
takePicture() {
    const options = {};
    options.location = this.state.location;
    this.camera.capture({ metadata: options })
      .then((data) => console.log(data))
      .catch(err => console.error(err));
  }
```
everytime i press capture the application stop working

Issue #617, opened 2017-03-02
Wrong path returned by this.refs.camera.capture() when capture mode is cameraRoll (Android)


kev-slater-mhr:
Node version: 7.5
React-native-cli: 2.0.1
React-native: 0.41.2
Platform: Android 5.0
React-native-camera: 0.6.0

When taking a photo and the capture mode is cameraRoll, the path returned from this.refs.camera.capture() is the disk location of the image (i.e. file:///storage/emulated/0/DCIM/IMG_20170302_091011.jpg) as opposed to the images location in the camera roll (i.e. content://media/external/images/media/16329).

I suspect that the problem is down to the fact that no callback is provided in the call to MediaScannerConnection.scanFile in RCTCameraModule.java to capture (no pun intended!) the correct resource reference. 
This problem becomes more evident if you try and chain the camera view with a camera roll view and the image just taken does not immediately appear in the gallery of images (i.e. MediaScannerConnection has yet to complete its scanning).

npomfret:
Hi, this should be fixed in https://github.com/lwansbrough/react-native-camera/pull/637, would you like to test it?  

The currently implementation isn't wrong, its just there are 2 ways to represent the new image in android.  This PR adds a 2nd field to the response called 'mediaUri' which I believe is the thing you want.

kev-slater-mhr:
Hi, Will do hopefully within the next week. Thanks for the fix.

kev-slater-mhr:
Just given it a quick check and it looks good to me. Thanks for the resolution! Any ideas when this will be merged into master?

npomfret:
I hope later today.  Just testing out the other changes

npomfret:
If you could comment on the PR that would be helpful.

npomfret:
Can we close this?

PR #637, opened 2017-03-20
make android handle metadata.location (and some other minor things...)

npomfret:
* Android implemention now uses the metadata.location - when asking the RN CameraRoll for an image taken with this library it will now return with location information attached (same as iOS).
 * removed a little duplication in the java code, cleaned up some logging
 * added a 'mediaUri' to the return type - this is the same as the uri that the RN CameraRoll returns (applicable in certain save modes only)
 * clarified how to use metadata in docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,9 @@ class BadInstagramCloneApp extends Component {
   }
 
   takePicture() {
-    this.camera.capture()
+    const options = {};
+    //options.location = ...
+    this.camera.capture({metadata: options})
       .then((data) => console.log(data))
       .catch(err => console.error(err));
   }
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
@@ -516,13 +516,15 @@ public class RCTCameraModule extends ReactContextBaseJavaModule
     }
 
     private static Bitmap toBitmap(byte[] data) {
+        ByteArrayInputStream inputStream = new ByteArrayInputStream(data);
         try {
-            ByteArrayInputStream inputStream = new ByteArrayInputStream(data);
-            Bitmap photo = BitmapFactory.decodeStream(inputStream);
-            inputStream.close();
-            return photo;
-        } catch (IOException e) {
-            throw new IllegalStateException("Will not happen", e);
+            return BitmapFactory.decodeStream(inputStream);
+        } finally {
+            try {
+                inputStream.close();
+            } catch (IOException e) {
+                Log.w("problem closing stream", e);//will not happen
+            }
         }
     }
 
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
@@ -674,10 +674,11 @@ public class RCTCameraModule extends ReactContextBaseJavaModule
 
                 camera.stopPreview();
                 camera.startPreview();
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -112,7 +112,9 @@ class BadInstagramCloneApp extends Component {
   }
 
   takePicture() {
-    this.camera.capture()
+    const options = {};
+    //options.location = ...
+    this.camera.capture({metadata: options})
```
rt2zz: what is the expected shape of metadata? what is the reason not to simply have metadata be a boolean?

npomfret: The existing _readme.md_ says

> metadata This is metadata to be added to the captured image.
> location This is the object returned from navigator.geolocation.getCurrentPosition() (React Native's geolocation polyfill). It will add GPS metadata to the image.

So, for example the shape if you want to add location info would be:

`this.camera.capture({metadata: {location: {lat:.., lon: ...}})`

This PR doesn't change that.

rt2zz: ty, I had not known about that option. Seems to be a clunky api :(

android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
```diff
@@ -751,6 +760,29 @@ public void onPictureTaken(byte[] data, Camera camera) {
...
+    private void writeLocationExifData(File cameraRollFile, ReadableMap options) {
+        if(!options.hasKey("metadata"))
+            return;
+
+        ReadableMap metadata = options.getMap("metadata");
+        if (!metadata.hasKey("location"))
+            return;
+
+        ReadableMap location = options.getMap("location");
+        if(!location.hasKey("coords"))
```
rt2zz: AFAICT this means the metadata option will need both location and coords in order to be able to retrieve either? Is this on purpose?

assuming I understand this correctly, imo there are two reasonable options here
- change the option from `metadata: object` to something like `exifGeo: boolean`
- console.error in JS if `location XOR coords`

Thoughts?

npomfret: The RN geo location returns a gps position that looks like this:

```
{
  coords: {lat..., lon...},
  timestamp: 1234454
}
```

I'm not sure what you mean about passing a boolean instead of a gps location.

This change is just bringing the android API in line with the iOS one.

rt2zz: 👍 ya I realize that, it was a bad api to start with

rt2zz:
@npomfret is this ready to go? my dislike of the iOS api not withstanding, it would be great to get this in asap. 👍

npomfret:
I'm happy with it so far, it only parses latitude and longitude values for now, so not quite as good as the iOS one (which does speed, heading etc).  But it's a big improvement.  When you get your images from the RN CameraRoll they come with a location field at least. 

Give me another day to test.

kev-slater-mhr:
I can confirm that #617 (Wrong path returned by this.refs.camera.capture() when capture mode is cameraRoll) is fixed by #637.

npomfret:
I made a typo before my last commit which broke it, have fixed it but it's not appearing in my PR for some reason.  Maybe because it's approved?  Shall I just create another one?

npomfret:
@rt2zz  I've also got some performance fixes for android to go in.  But let's do those as a separate PR.

rt2zz:
@npomfret approve would not affect your PR, you must not have pushed to the right branch, but ya a new PR is just as well.

npomfret:
https://github.com/lwansbrough/react-native-camera/pull/643/commits

PR #643, opened 2017-03-23
added gps meta to android (attempt 2)

npomfret:


npomfret:
ok, I don't what I did there but here it is again.

npomfret:
thanks, another one on its way...

PR #644, opened 2017-03-23
android performance improvements

npomfret:
* removed some of the re-parsing of the output image byte array
 * removed re-saving of output file
 * moved image processing onto an async task to allow camera to be used while processing is running

### DIFF
--- a/android/src/main/java/com/lwansbrough/RCTCamera/MutableImage.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/MutableImage.java
@@ -1,0 +1,285 @@
+package com.lwansbrough.RCTCamera;
+
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.graphics.Matrix;
+import android.media.ExifInterface;
+import android.util.Base64;
+import android.util.Log;
+
+import com.drew.imaging.ImageMetadataReader;
+import com.drew.imaging.ImageProcessingException;
+import com.drew.metadata.Directory;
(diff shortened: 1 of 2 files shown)


npomfret:
I've asked a couple of people who've complained about performance to test but no response so far.  I've tested on some pretty low spec android 4.1 and 5 devices and it seems fine so far.  The image processing step is still not as fast as I'd like it to be though.

And we're losing some of the image metadata because the android exif library doesn't understand very many tags.  Maybe I'll look into finding a replacement for that next.

rt2zz:
This is certainly a step in the right direction, lets get it on master which should get some more eyes on it.

molerat619:
Thanks for working on these improvements. Maybe I am testing the wrong things, but I can't see any improvements on my LG G2 Android 4.4.
I was on 0.6.0 and upgraded using `npm install react-native-camera@https://github.com/lwansbrough/react-native-camera.git --save` - I hope that was correct.

The problems that I am facing are the following:
- very slow transition from one scene to camera scene
- if I use interaction manager, transition is good, but camera still takes too long to load (compared to Whatsapp for example)
- after taking a picture, the transition to the next scene takes too long (starts too late)

I am sure, some issues are on my side or react-native-router-flux. I would still be very grateful if anybody could support me here. I have been testing several things for days now.

npomfret:
I don't know what's causing this for you.  I've tested this on some very old and low powered android phones and if feels good.  Perhaps you could rule out the camera API by not loading it and see if your scene transition times are still slow.

PR #646, opened 2017-03-24
Update Android SDK Build Tools version

lucasbento:


ptomasroos:
What's the reason behind this PR?

android/build.gradle
```diff
@@ -11,8 +11,8 @@ buildscript {
 apply plugin: 'com.android.library'
 
 android {
-  compileSdkVersion 23
-  buildToolsVersion "23.0.1"
+  compileSdkVersion 25
+  buildToolsVersion "25.0.1"
```
Minishlink: - 25.0.2 is the lastest
- compileSdkVersion should stay at 23, react-native-camera doesn't take advantage of features that needs API 24 or 25

Minishlink:
@ptomasroos Build Tools 25 are required in order to take advantage of Gradle 3.3, which has great performance improvements!

android/build.gradle
lucasbento: @Minishlink: hmm, what would be the downside of using `compileSdkVersion` as `25`?

Minishlink: @lucasbento It would require people to download SDK 25, which is not as lightweight as only updating build tools (several Go)

Minishlink: @lucasbento The moment react-native target API 25, I'm all for it. But as of today, close to everybody has API 23 (mostly because of the [getting started](https://facebook.github.io/react-native/docs/getting-started.html#3-install-the-android-6-0-marshmallow-sdk))

ptomasroos:
I don't see this as something to move on with. The entire community is adapting the build, target and compile's are base react-native more or less is using. To get performance improvements on Gradle 3, please be my guest to add a "postinstall" to your package.json which walks through every build.gradle and upgrades those dependencies. I've done it for other use cases myself.

Minishlink:
@lucasbento If you want to override the build tools of your dependencies, there's a trick ;) I've just written a blog post on this : https://blog.bam.tech/developper-news/boost-compilation-time-react-native-android-app-latest-android-build-tools

sibelius:
I've wrote a script to upgrade all deps

https://gist.github.com/sibelius/0a5403a2b5e9dd9ed9d98078a1523a18

PR #653, opened 2017-03-31
Update compileSdkVersion and buildToolsVersion for android to 25

LowieHuyghe:


rt2zz:
presumably this will this cause issues for people with build tool still on 23?

Is the whole ecosystem moving to 25?

ptomasroos:
No RN is still stuck on 23 and I don't expect any change here soon. 

You can always use a post install which upgrades these versions in build.gradle

PR #654, opened 2017-04-02
fix json stringify non string key error

srt0422:
Error occurs with CameraManager.getFOV().  I had to hard code string keys in the resolve parameter object to use "getFOV".

Please merge.  My project is dependent on this fix.

### DIFF
--- a/ios/RCTCameraManager.m
+++ b/ios/RCTCameraManager.m
@@ -407,8 +407,8 @@ RCT_EXPORT_METHOD(getFOV:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRejec
   }
 
   resolve(@{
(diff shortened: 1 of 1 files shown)


rt2zz:
Thanks for the PR. Why does this work for some but errors for you? Can you paste the error maybe there is something else wrong, I am not keen on changing the constants strategy unless we change it consistently for the entire project.

srt0422:
I understand.  I thought that might be a concern.

My guess is that it has to do with our react-native version.

I'm working with react-native version 0.42.3

My project is located here:
https://github.com/srt0422/location-based-ar-example

After a quick google search, it looks like it may be related to this:
http://stackoverflow.com/questions/42617842/react-native-error-rctjsonstringify-encountered-the-following-error-invalid-t

I'll try running it without the debugger and see if I still get the error.

PR #663, opened 2017-04-10
Fix example crashes

MNBuyskih:
Update react and react-native then java compiller told me:

> {...}\react-native-camera\Example\android\app\src\main\java\com\example\MainApplication.java:20: error: getUseDeveloperSupport() in <anonymous com.example.MainApplication$1> cannot override getUseDeveloperSupport() in ReactNativeHost
    protected boolean getUseDeveloperSupport() {
                      ^
  attempting to assign weaker access privileges; was public

Fix this and voila! It`s works!


ptomasroos:
This is because you upgraded the version in the example, hence its no error in the example by default.

Thanks for the contribution.

PR #674, opened 2017-04-21
[Fix] Image orientation bug when mirroring set to true

CiralaDevelopment:
### What was the issue?

- Taking pictures on a device that doesn't rotate the image, but rather saves rotation in EXIF, with mirroring turned on resulted in an image rotated the wrong way because the mirroring flips the image before it's rotated

### How was it fixed?

- Rotate the image before mirroring

### DIFF
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
@@ -561,18 +561,18 @@ public class RCTCameraModule extends ReactContextBaseJavaModule
      * concurrently which would blow the memory (esp on smaller devices), and slow things down.
      */
     private synchronized void processImage(MutableImage mutableImage, Boolean shouldMirror, ReadableMap options, Promise promise) {
+        try {
+            mutableImage.fixOrientation();
+        } catch (MutableImage.ImageMutationFailedException e) {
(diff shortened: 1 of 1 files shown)


mikaello:
I had the same issue on a Samsung S7 (picture is rotated when `mirrorImage={true}`), and this PR solves the problem.

rt2zz:
cc @npomfret @charpeni does this make sense to you guys?

npomfret:
I don't quite understand the issue here.  But I don't have a device to test on that exhibits the behaviour.  The change looks innocent enough, I'm happy to push it through.  Can we just get one more person to test it who's got the same issue?

CiralaDevelopment:
@npomfret Because the image is mirrored before it is rotated, and the rotation angle stays the same, it is rotating the image the wrong way. When mirroring either the rotation needs to happen before the mirror or the mirroring needs to mirror the rotation angle as well. (think of it like the rotation is happening on the backside of the image because of the mirroring instead of the frontside)

npomfret:
@CiralaDevelopment thanks for the feedback.  Are you happy that this change work as intended?

CiralaDevelopment:
@npomfret Yes

npomfret:
@CiralaDevelopment  Many thanks for this PR.

ptomasroos:
Great stuff! I noticed this on our production as well specially on Samsung devices.

PR #679, opened 2017-04-23
Set more precise React/React-Native package version dependencies for Example

abrahambotros:
- In general, try to avoid breaking changes caused by >=
  - In particular, such a breaking change can be seen by testing the current Example
    app on Android with the existing package.json's and RN 0.43+, which causes errors
    in MainApplication public/protected method settings.

### DIFF
--- a/Example/package.json
+++ b/Example/package.json
@@ -6,8 +6,8 @@
     "start": "node node_modules/react-native/local-cli/cli.js start"
   },
   "dependencies": {
-    "react": "~15.3.0",
-    "react-native": "^0.34.0",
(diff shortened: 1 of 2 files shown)


rt2zz:
I agree with fixing the example app, but for peer deps things get trickier, especially when people want to use rc's.

abrahambotros:
@rt2zz, ah jeez of course, temporarily lapse of judgment there on my part, big thanks for correcting! Reverted changes to the main `package.json`, leaving for Example's `package.json`, and updated PR title accordingly.

rt2zz:
many thanks!

PR #682, opened 2017-04-24
make fixOrientation an option

npomfret:
* added an option to fix orientation, which defaults to false
* when 'fixing' the orientation it now does nothing if the orientation in the exif data is already set to '1' (right way up)
* after fixing the orientation it writes the orientation value into the exif data (instead of deleting it in iOS, and leaving it unchanged in android)

How to test:

In summary I think that auto-fixing the orientation isn't required because the EXIF data is now preserved.  So I've added a property `fixOrientation` which defaults to false in this PR.  So if you don't set it (i.e. make no changes to your JS code) and take a photo at each of the 4 possible orientations I think they should come out correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,14 @@ If set to `true`, the device will not sleep while the camera preview is visible.
 
 If set to `true`, the image returned will be mirrored.
 
+#### `fixOrientation` (_deprecated_)
+
+If set to `true`, the image returned will be rotated to the _right way up_.  WARNING: It uses a significant amount of memory and my cause your application to crash if the device cannot provide enough RAM to perform the rotation.
+
+(_If you find that you need to use this option because your images are incorrectly oriented by default,
+could please submit a PR and include the make model of the device.  We believe that it's not 
+required functionality any more and would like to remove it._) 
+
 ## Component instance methods
 
 You can access component methods by adding a `ref` (ie. `ref="camera"`) prop to your `<Camera>` element, then you can use `this.refs.camera.capture(cb)`, etc. inside your component.
--- a/android/src/main/java/com/lwansbrough/RCTCamera/MutableImage.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/MutableImage.java
@@ -76,6 +76,7 @@ public class MutableImage {
         }
     }
 
+    //see http://www.impulseadventure.com/photo/exif-orientation.html
     private void rotate(int exifOrientation) throws ImageMutationFailedException {
         final Matrix bitmapMatrix = new Matrix();
         switch (exifOrientation) {
--- a/android/src/main/java/com/lwansbrough/RCTCamera/MutableImage.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/MutableImage.java
@@ -66,7 +66,10 @@ public class MutableImage {
                 return;
             } else if (exifIFD0Directory.containsTag(ExifIFD0Directory.TAG_ORIENTATION)) {
                 int exifOrientation = exifIFD0Directory.getInt(ExifIFD0Directory.TAG_ORIENTATION);
-                rotate(exifOrientation);
+                if(exifOrientation != 1) {
+                    rotate(exifOrientation);
+                    exifIFD0Directory.setInt(ExifIFD0Directory.TAG_ORIENTATION, 1);
+                }
             }
         } catch (ImageProcessingException | IOException | MetadataException e) {
             throw new ImageMutationFailedException("failed to fix orientation", e);
@@ -77,7 +80,7 @@ public class MutableImage {
         final Matrix bitmapMatrix = new Matrix();
         switch (exifOrientation) {
             case 1:
-                break;
+                return;//no rotation required
             case 2:
                 bitmapMatrix.postScale(-1, 1);
                 break;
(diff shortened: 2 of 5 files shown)


npomfret:
I've not been able to test this on android because my devices don't set the exif data, i just get right-way-up images.

I have tested on iOS and it seems fine.  Interestingly I noticed that only 2 of the 3 rotations are applied in the iOS code (there's no 90 degree rotation) which confused me.  I still can't figure out why - does anyone know why?

Also, the rotations in the iOS code are applied differently to the android code.  The exif values are interpreted differently:

```
if (metadataOrientation == 6) {
              rotatedCGImage = [self newCGImageRotatedByAngle:cgImage angle:270];
              rotated = true;
            } else if (metadataOrientation == 3) {
              rotatedCGImage = [self newCGImageRotatedByAngle:cgImage angle:180];
              rotated = true;
            }
```

And in android its:

```
switch (exifOrientation) {
            case 1:
                return;//no rotation required
            case 2:
                bitmapMatrix.postScale(-1, 1);
                break;
            case 3:
                bitmapMatrix.postRotate(180);
                break;
            case 4:
                bitmapMatrix.postRotate(180);
                bitmapMatrix.postScale(-1, 1);
                break;
            case 5:
                bitmapMatrix.postRotate(90);
                bitmapMatrix.postScale(-1, 1);
                break;
            case 6://this is 270 degrees in iOS!?
                bitmapMatrix.postRotate(90);
                break;
            case 7:
                bitmapMatrix.postRotate(270);
                bitmapMatrix.postScale(-1, 1);
                break;
            case 8:
                bitmapMatrix.postRotate(270);
                break;
            default:
                break;
        }
```

molerat619:
Thanks for your contribution! How can I test this? Never tried to get PR from GitHub running on my machine before. I will be happy to give you feedback for an Android 4.4 device.

npomfret:
ok, the easiest way probably is to run with my fork.  Edit your package.json and replace the react-native-camera url with the url of my fork which is git+https://github.com/npomfret/react-native-camera.git

Then run `npm install && react-native start`.  When you're done testing remember to revert the change.

molerat619:
I have tested your fix (up to commit 71a188d) and removed the largeHeap fix before I started actually testing. I have performed actions that would have caused a crash without the largeHeap and without your fix and this time, the app did not crash. There are no performance improvements though.

npomfret:
@molerat619 thanks for testing, thats good news that it doesn't crash.  Have you tried taking image at all 4 camera orientations?  Did they appear correctly?

molerat619:
Correct. Of course, upside-down is not possible, but the other 3 orientations worked no problem.

mikaello:
I have testet this PR now, but on my phone both the current master (d54eae9ea5c41d5a1cf955cc5c7ec2f88dff81ba) and your PR seems to behave the same, all orientations seemed to appear correctly.

**Top row:** default camera on my Samsung S7
**Middle row:** current master (`└── react-native-camera@0.7.0  (git://github.com/lwansbrough/react-native-camera.git#d54eae9ea5c41d5a1cf955cc5c7ec2f88dff81ba)`)
**Bottom row:** your PR (`└── react-native-camera@0.7.0  (git+https://github.com/npomfret/react-native-camera.git#85761123799bb4f37a302009a6c7967990aec8c5)`)

**First column:** Picture taken in normal portrait
**Second column:** Phone rotated 90 degrees to the right (from normal potrait)
**Third column:** Upside down portrait
**Fourth column:** Phone rotated 90 degrees to the left (from normail portrait)

Screenshot of the grid view in the default gallery app on Samsung S7:

rt2zz:
awesome, @npomfret is this ready?

npomfret:
There's a conflict, and it needs a document update.  I'll sort it today.

Still a bit nervous.  Would like to get it tested on more devices.  If it is ok, then in time I think we should remove the optional 'fixRotation' altogether.

ptomasroos:
I agree @npomfret this should not be a special prop, feels awkward

npomfret:
I think it's premature to remove it altogether, maybe in a future release.

Antoine-C:
Nice job @npomfret !

Is this PR ready to be merged in master ?

npomfret:
@rt2zz would you like to merge this - i think it's ready for master.

rt2zz:
@npomfret 👍  thank you!

QQizi:
@npomfret I'm developping an app who require some webview with a canvas in it.

When i capture a picture with react-native-camera and send it to the webview, the image always comes out rotated in landscape. This bug doesn't happen in simulator and doesn't happen with CameraRoll.

If i use fixOrientation : true, problem does not appear.
I'm on iPhone 6 IOS 10.3.2

So i guess fixOrientation is still needed

npomfret:
Thanks for the feedback.  Maybe some web browsers ignore the embedded metadata.  Which browser do you see the rotate image in?

QQizi:
@npomfret I'm using the react-native Webview component, not a web browser.

npomfret:
On what devices?

QQizi:
@npomfret  I'm on iPhone 6 IOS 10.3.2

skovhus:
@npomfret I've tested out #682 (0.10.0):
- you can now avoid the memory crash if not using the `fixOrientation` flag
- but the rotation is still wrong -90 for front facing camera the different Android devices I tested (Samsung Galaxy S7 / Android 7 and a Samsung Galaxy S4 Android 5)

mikaello:
In my app the front facing camera is important, and a lot of users are using Android, so I have figured out that version 0.8.0 works, but any later version has the -90 rotation bug.

skovhus:
0.8.0 had a auto rotation that caused memory issues on old android phones. 0.9+ made this optional using the fixOrientation flag.

PR #684, opened 2017-04-25
Fix for bounds in Android

MocanuMarius:
This needs to be reviewed and documented for Android because it behaves a little bit weird..but at least i want your attention on this to make this work please :)


MocanuMarius:
I willl come back with this pull with proper curated values if you want.. just ping me.

PR #685, opened 2017-04-28
Provide dual support for React Native pre and post 0.40.0

dvicory:
For various reasons, upgrading React Native is problematic for us right now. We'd like to continue to use the improvements made here so this conditionally branches the iOS imports which are currently the only breaking change.


rt2zz:
thats pretty cool, I did not know you could do that. Would appreciate if someone more objC literate than me can gives this a thumbs up before we merge.

dvicory:
We tested with a fresh 0.39.2 and 0.43.4 project. But would welcome a thumbs up from someone else.

However, we do not use (or tested) CocoaPods. However I also don't expect this change to have had any impact on whether or not it would work when used as a Pod. My gut feeling is that use as a Pod may not quite work currently anyway.

ptomasroos:
We're using podspec @dvicory so please don't tell people its not working or assume its not. I can def try out the PR in our app and report back on that topic.

npomfret:
Hi.. Thanks for this PR!

I'm sure it works as described, I've seen other projects use this technique, although I wasn't expecting to see a change to the `project.pbxproj` file.  What's that for?

My opinion is that we shouldn't support older versions of RN.  I appreciate that everyone isn't on the latest build.  But if the policy is to support multiples versions I can see the code getting very complicated.  I think going forward, because of the popularity of RN, very few users will be on a pre RN0.40 build, so we're complicating the code for a small minority of projects.  

I think the policy should be to encourage people of upgrade if they want the new features.  Of course users always have the option to manage a fork of their own if upgrading isn't feasible for them at a time when they're desperate for a feature.  We could even put a link to working forks for older versions in the docs.

..but that's just my opinion, feel free to ignore it!

dvicory:
@npomfret, the change to `project.pbxproj` is what is normally there in pre-0.40. It's needed so the compiler can find the files, e.g. `#import "RCTViewManager.h"`. Other projects that provide dual support kept their `HEADER_SEARCH_PATHS`.

I can see where you're coming from about supporting older versions of RN. I don't believe that right now it is an unreasonable ask. The breaking change only affected imports and in my opinion does not create any unnecessary complications. I do agree that if more substantial changes than simply imports were needed that it could become excessive.

Right now we are forked, and that works. But that means we have to be vigilant about bringing back upstream changes, which is a potential for error. And IMO, a waste of people's time. For us, getting new features isn't so much the problem, it's lack of bug fixes with existing features.

@ptomasroos, would definitely appreciate a report back that this has no effect on using it as a Pod.

ptomasroos:
Yes, will try it out during the day

ptomasroos:
Works as expected but I'm leaning to @npomfret conclusion here, allowing multiple old versions will be seriously hard to test and maintain.

Any opinions from you @rt2zz ?

sibelius:
I think we should only support the latest version of react native

sibelius:
0.40 is too old now

PR #693, opened 2017-05-05
remove peerDependencies in package.json

liuqiang1357:
In react-native 0.44, this will cause "UNMET PEER DEPENDENCY" warning.


npomfret:
Thanks for this.

PR #695, opened 2017-05-06
[ios + android] Added "width", "height" and "uri" to captured image when target is "temp"

kesha-antonov:


npomfret:
Thanks for this PR.

Why only 

> when target is "temp"

?

kesha-antonov:
Hey! @npomfret 

No special purpose. Just I'm using only "temp"

You want it for all?

npomfret:
Sure.  I don't think it make sense to change the return type for one and not the others.  Maybe images should have the same shape too?

kesha-antonov:
I think so.
I'll add this. I think it's not muck work

pewh:
Hello,

Width & height attribute is very important for me. I'm curious why these not merge to repo?

spikeisryu:
Yes, please merge it to repo ASAP.

victorbadila:
Why is this not merged yet? looks clean :/

PR #697, opened 2017-05-08
Update from View.propTypes to ViewPropTypes to match RN v0.44.0

TBouder:
Update the ViewPropTypes to match the last revision of React Native V0.44.0

> View.propTypes to ViewPropTypes (53905a5) - @bvaughn
https://github.com/facebook/react-native/releases

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,6 +7,7 @@ import {
   StyleSheet,
   requireNativeComponent,
   View,
+  ViewPropTypes
 } from 'react-native';
 
 const CameraManager = NativeModules.CameraManager || NativeModules.CameraModule;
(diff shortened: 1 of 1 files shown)


rt2zz:
As I recall, the peer dep was recently removed which makes me concerned to ship this change. Prematurely.

npomfret:
I'm getting a bunch of warnings from my RN0.44 app now - I think this change will help.  I'll test it from my fork.

npomfret:
Looks good to me. @rt2zz what are your concerns re merging this?

rt2zz:
nothing lets do it. ViewPropTypes will not exist in older versions of react-native, but we cannot wait on that forever.

PR #698, opened 2017-05-09
Add totalSeconds description in README.md

colus001:
- Add missing `totalSeconds` capture options to document


npomfret:
Thanks for this PR.

Is this option supported on both iOS and android?

colus001:
@npomfret Actually I haven't tested on android but I can see that code in android side code

https://github.com/lwansbrough/react-native-camera/blob/d54eae9ea5c41d5a1cf955cc5c7ec2f88dff81ba/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java#L315

npomfret:
thanks

PR #703, opened 2017-05-12
Add some new features on Android.

Symous:
Hi, awesome lib. :-)

I have added some new features on Android Platfrom.

1. add width/height/duration/size properties(which is ios supported only before) of the returned object when we capture a video.
2. make it possible for us to know when the camera is start to record a video actually by add 2 props onRecordVideoStarted and onRecordVideoStopped. Because when I use this lib in my app,when I invoke capture() and create a progress animation, there is always a 'delay' and not smoothly enough. I have solve after add these 2 prop. **But notice:Android only.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,14 +214,6 @@ Values: `true` (default) or `false`
 
 This property allows you to specify whether a shutter sound is played on capture. It is currently android only, pending [a reasonable mute implementation](http://stackoverflow.com/questions/4401232/avfoundation-how-to-turn-off-the-shutter-sound-when-capturestillimageasynchrono) in iOS.
 
-#### `onRecordVideoStarted` `(Android only now)`
-
-Will call the specified method when the camera start recording a video.
-
-#### `onRecordVideoStopped` `(Android only now)`
-
-Will call the specified method when the camera finish recording a video.
-
 #### `onBarCodeRead`
 
 Will call the specified method when a barcode is detected in the camera's view.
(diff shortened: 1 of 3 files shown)


rt2zz:
what features does `onRecordVideoStarted` and `onRecordVideoStopped` give you that cannot be accomplished by simply calling your handler before starting the recording and again after completion?

Symous:
@rt2zz ,Hi, the reason of why I add onRecordVideoStarted prop is that we need to know when the camera of our phone has switch to ‘Video Mode’ and ready for recording(In your android native code when **mMediaRecorder.start();** execute ).  

I write thie in my app like this before,
`
this.showProgressAnimation(); // I show a 10 seconds animation here.
this.camaera.capture({mode:  Camera.constants.CaptureMode.video,totalSeconds: 10}) // record 10 seconds.
`
but I found that the camera need time to ready and start recording.these 2 step will never stop and a same time.

Or can you please tell me is ther any other way to do this without using onRecordVideoStarted ?

rt2zz:
I see - yes this seems valid. Naturally this will be approached with caution as we do not want the complexity to grow here, and furthermore would need a matching iOS implementation.

It would be great if others can weigh in on this feature as well 👍

Symous:
@rt2zz , well , I guess that is necessart since I have met the problem. But there is another feature  in this RP that add width/height/size/duration of the recorded video which is only supported by iOS before. Will I need to create another PR ?

phuochau:
@Symous how is it going? I hope this PR will be merged soon.

Symous:
@phuochau Hi,the author haven't give any reply yet.

npomfret:
Can you try and get someone to test your PR?  Maybe offer to test someone else's PR in exchange for testing yours.  Make it clear what needs testing (what is the expected change in behaviour), and how to test it (if it's not obvious)/  The testing process can be quite lengthy and so the contributors don't often have time to test them all.

phuochau:
@Symous yay, thanks so much for adding these features. Maybe I will try to clone from your PR.

JLLLinn:
It's been quite long, is this PR still under consideration? I for one would be really interested in the duration feature in Android

Issue #328, opened 2016-06-22
Starting with TorchMode.On is broken


JAStanton:
### Steps to reproduce
1. `<Camera torchMode={Camera.Constants.TorchMode.On} />`
2. Observe the flash turn on, then immediately off.
### Expected behaviour

The light should stay on.
### Actual behaviour

The light flickers off.
### Environment
- **Node.js version**: v6.2.1
- **React Native version**: 27.1
- **React Native platform + platform version**: iOS 9.3.2
### react-native-camera

**Version**: npm version or "master": 0.3.8
### Related issue

similar: https://github.com/lwansbrough/react-native-camera/issues/123 and fixed here: https://github.com/lwansbrough/react-native-camera/pull/261 but it doesn't fix the light staying on.  mentioned here as well: https://github.com/lwansbrough/react-native-camera/issues/123#issuecomment-217403124

theopolisme:
+1 encountering as well

Shawn8585:
+1

andrewvmail:
+1

Guillane:
+1

mesbahsabur:
+1

mesbahsabur:
See my comment at #123

wouteralberts:
I fixed this by using state for torchMode (initially `off`) and setting it to `on` after showing my camera component:

```
this.setState({showCameraComponent: true}, () => {

    this.setState({torchMode: Camera.constants.TorchMode.on})

});
```
`<Camera ... torchMode={ this.state.torchMode }/>`

Symous:
@wouteralberts that's works, thank u!

arberK:
+1

imapolaris:
yes, just as @wouteralberts said. It works well for me!

ma-pe:
Thanks for the _workaround_ @wouteralberts. 

Nevertheless this needs to be fixed.

ospfranco:
Hey! this repository was recently transferred to the react-native-community, as part of our ongoing cleaning process we are closing old tickets (there is just too many of them!), please feel free to try the latest versions or the master branch to see if your problem has been solved.

If you are still experiencing this problem feel free to re-open this or fill a new ticket.

Issue #123, opened 2015-11-18
torchMode not turning on light


brancusi:
Hi,

I would like to enable torchMode during barcode scanning. I am setting the following:

``` javascript
<Camera
   onBarCodeRead={::this._onBarCodeRead}
   type={Camera.constants.Type.back}
   torchMode={Camera.constants.TorchMode.on}
/>
```

For some reason this is not starting the torch though. Everything else works just fine, camera view and scanning.

What's strange is that when I navigate away then back to the scene, I see a flicker of the torch but it shuts off.

Device info: iphone 6 plus
OS: ios 9.1

Any ideas?

lwansbrough:
Interesting, not sure about this. I'll look into this when I get a chance. If you're familiar with Obj-C I might recommend you dive into it a bit yourself if you're willing - might be faster than waiting for me. I'd definitely accept a PR if you discover the issue.

cava23:
So far I've traced this down to the setType method in RCTCamera.m. Torch mode is on until initializeCaptureSessionInput:AVMediaTypeVideo is called in setType. At this point, the currently configured device input is replaced, disabling torch mode. If I comment out the call to initializeCaptureSessionInput, torch mode stays on correctly. Commenting that out probably breaks something else, but that's just as far as I've gotten. @seanj

cava23:
Here ya go, wrap setType with:
```
if (self.manager.presetCamera != type) {
   // change the camera type
}
```
I'll make a PR and see what you think.

rajatmehndiratta:
Running into the same issue; was @cava23's PR made? I can't find it when I search + don't see the change in RCTCamera.m.

cava23:
I couldn't quite get it right, so I never made a PR. We ended up initializing the camera on componentDidMount, after the props were set. You can see our fork here: https://github.com/Tredsite/react-native-camera

rdhaliwal:
FWIW @brancusi I had the same issue where I tried to initialise the Camera with TorchMode.on, but when I changed it so it was set with a variable and it was initially set to off, then toggled to on, it started to work for me. Maybe it'll help? ¯\_(ツ)_/¯

```
<Camera torchMode={this.props.flash ? Camera.constants.TorchMode.on : Camera.constants.TorchMode.off}
                ref={(cam) => { this.camera = cam; }} />
```

Obooman:
[...]

PR #709, opened 2017-05-18
Fix torch init status

Symous:
Hi, I have found that if we set the torchMode as Camera.constants.TorchMode.on initial , the torch is still in off status, So I check the source code and slove this by fix the wrong method you invoke before. 
I also notice that many people have met this before , see issues :　#328 , #123 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,14 +214,6 @@ Values: `true` (default) or `false`
 
 This property allows you to specify whether a shutter sound is played on capture. It is currently android only, pending [a reasonable mute implementation](http://stackoverflow.com/questions/4401232/avfoundation-how-to-turn-off-the-shutter-sound-when-capturestillimageasynchrono) in iOS.
 
-#### `onRecordVideoStarted` `(Android only now)`
-
-Will call the specified method when the camera start recording a video.
-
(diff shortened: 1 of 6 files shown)


npomfret:
Many thanks for this PR!  Can you please help testing by describing:

 * what is the expected change in behaviour
 * what steps must someone who's testing your PR make in order to observe the change in behaviour
 * what operating systems are affected and need to be tested

HINT: a good way to get it tested and merged is to offer to test someone else's PR in exchange for testing yours.

Symous:
@npomfret Thanks for your feedback.
1. If we create a Camera component and set torchMode prop as true, the excepted behaviour of the camera is the flash light keeps lighting.(Like a torch).
2. In the old version it can not be lighting because of the fixed bug of the Android Native code(the reason is it invoke the wrong method which you can see it in the changed file.)
3. I use this with Android OS(there are other people also meet this issuse with Android, iOS is OK).

 this is a very obviously bug I just modify one line code whici will not take any other effect at this lib. I have tested it by myselft. :-)

rt2zz:
@npomfret lgtm, please merge if you agree

PR #713, opened 2017-05-22
Add option to save picture to app directory

logileifs:
This is a useful option for those who don't want to clutter the gallery with the images taken using their app


npomfret:
Thanks for the PR!

Can you describe for someone to test:

 * what is the expected change in behaviour?
 * what steps should be taken to test it (if that's not obvious)

HINT: testing someone else's pending PR in exchange for them testing yours is a good way to get it looked at.  The contributors don't alway have time to test.

logileifs:
expected change in behavior: on iOS when using disk as capture target images are saved in the application's directory and don't appear in the photo library. on Android all images taken with capture target disk appear in the gallery and are not saved in the application's directory. this new option provides the same functionality for Android for those who don't want their app's photos to appear in the gallery. 

to test: just take a few pictures and check the gallery and check the device's storage

npomfret:
Ah, this is partially solved by my recent PR which addresses a regression made a while ago. 

https://github.com/lwansbrough/react-native-camera/pull/716

If it works, perhaps this is good enough for you?

logileifs:
If I understand the other one correctly it only 'hides' the pictures from the gallery. However the images are still saved somewhere publicly on the device which is not my desired behavior. I think there should at least be an option to save the images in the application's directory, there are plenty of use cases where you don't want the images to appear publicly for the user to see.

npomfret:
>  there are plenty of use cases where you don't want the images to appear publicly for the user to see


That's a fair point.

I glanced at your changes (sorry, don't have time to test right now) and asked a couple of questions.  Also, would it be possible to do an iOS implementation?  And can you add the change to the documentation?

To get it tested you could offer to test someone else's PR in exchange for testing yours.

logileifs:
I'm sorry but my knowledge of objective C is far behind my Java skills and I don't think I have enough time to drive into such an assignment now. The iOS implementation is already there, it's just called disk, what needs to be done is just to offer the option of saving in some public directory on the iPhone but I'm not even sure if that's available on iOS. maybe someone with more iOS/objective C knowledge can pitch in here

npomfret:
thats ok, I can do it when I find time.  for now document it as an android only API and we'll fill the rest in another time.

Kernael:
Really useful feature, do you guys have any update on this ?

wassgha:
Bump!

PR #716, opened 2017-05-23
when saving images, only invoke the media scanner if the user is saving to the camera roll.

npomfret:
when saving images, only invoke the media scanner if the user is saving to the camera roll.


bheart:
I don't believe the `resolveImage` function is necessary to be quite honest, the call to `addToMediaStore` takes care of adding the image to the camera roll.

I personally solved the issue with this dad7fd8423e3c9f7645e9591271ed6185358d795 which is considerably simpler and doesn't invoke the MediaScanner twice.

DaBs:
Confirmed to be working, tested on the specific device mentioned in [the original issue](https://github.com/lwansbrough/react-native-camera/issues/708).

rt2zz:
thanks for confirmation 👍 

will cut a new release later this week

PR #728, opened 2017-05-29
Fix invalid reference when metadataOrientation is other than 3 or 6

aprimadi:
The variable 
rotatedCGImage is not initialized when metadataOrientation is other than 3 or 6. Image attached.

screen shot 2017-05-29 at 12 28 39 pm

### DIFF
--- a/ios/RCTCameraManager.m
+++ b/ios/RCTCameraManager.m
@@ -605,7 +605,7 @@ RCT_EXPORT_METHOD(hasFlash:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRej
           } else if ([options objectForKey:@"fixOrientation"]){
             // Get metadata orientation
             int metadataOrientation = [[imageMetadata objectForKey:(NSString *)kCGImagePropertyOrientation] intValue];
-            
+
             bool rotated = false;
             //see http://www.impulseadventure.com/photo/exif-orientation.html
             if (metadataOrientation == 6) {
@@ -614,8 +614,10 @@ RCT_EXPORT_METHOD(hasFlash:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRej
             } else if (metadataOrientation == 3) {
               rotatedCGImage = [self newCGImageRotatedByAngle:cgImage angle:180];
               rotated = true;
+            } else {
+              rotatedCGImage = cgImage;
             }
-            
+
             if(rotated) {
               [imageMetadata setObject:[NSNumber numberWithInteger:1] forKey:(NSString *)kCGImagePropertyOrientation];
               CGImageRelease(cgImage);


npomfret:
@aprimadi  Thanks for fixing this.

Are you actually using the `fixOrientation` option still?  If so, have you tried not using it?  We suspect it isn't required any more since the fix to the image metadata was made.

If you think it is required, can you let us know what device you are testing on?

aprimadi:
I'm using auto orientation. But the app detects orientation using react-native-orientation and re-renders the camera depending on the orientation. Here's how I call the Camera in the code:

```
          <Camera
            key={`cam-${orientation}`}
            ref={(cam) => { setCamera(cam) }}
            style={styles.preview}
            aspect={this.state.camera.aspect}
            captureTarget={this.state.camera.captureTarget}
            type={this.state.camera.type}
            flashMode={this.state.camera.flashMode}
            defaultTouchToFocus
            mirrorImage={false} />
```

Let me know if there's a better way to solve the orientation problem.

npomfret:
When you capture, do you pass fixOrientation as an option?

aprimadi:
No.

aprimadi:
If it helps, I was testing on an iPad Mini 1 device which runs iOS 9

npomfret:
I don't understand how your code ended up inside this if block

`} else if ([options objectForKey:@"fixOrientation"]){...`

aprimadi:
That's a good point, upon debugging, the line https://github.com/lwansbrough/react-native-camera/blob/master/ios/RCTCameraManager.m#L605 always returns true.

The statement
```
[options objectForKey:@"fixOrientation"]
```
returns an NSCFBoolean which is always evaluated to true since the propTypes fixOrientation is set to false in React.

The correct way is to do:
```
[[options objectForKey:@"fixOrientation"] boolValue] == YES
```

screen shot 2017-05-31 at 4 02 47 pm

I created another pull request for this.

npomfret:
Looks like that same bug is repeated just above in `if ([options objectForKey:@"rotation"]) {`

aprimadi:
It is kind of okay in the `if ([options objectForKey:@"rotation"]) {` because no default props is set in React, if no props is set, the statement `[options objectForKey:@"rotation"]` evaluates to nil which is then evaluated to false.

npomfret:
But if someone sets it to false in their code the behaviour will be wrong, no?

aprimadi:
Yes, but rotation is supposed to be a number. What I think we should do is just add prop types check

```
rotation: PropTypes.number
```

For checking propTypes passed in to the component.

If they add a rotation of 0, I think the statement inside the if should be able to handle that case as well. What do you think?

npomfret:
That's a very good point - ignore everything I just said!

PR #746, opened 2017-06-08
Add onBarCodeRead example

superyarik:
Many people looking for barcodescanner only, so there is direct example provided.


sibelius:
nice improvement

PR #759, opened 2017-06-20
Removed deprecated React CocoaPod from Podspec

BluDotz:
The CocoaPod dependency is deprecated since React is now installed during **react-native init**

### DIFF
--- a/react-native-camera.podspec
+++ b/react-native-camera.podspec
@@ -18,5 +18,4 @@ Pod::Spec.new do |s|
   s.preserve_paths = 'LICENSE', 'README.md', 'package.json', 'index.js'
   s.source_files   = 'ios/*.{h,m}'
 
-  s.dependency 'React'
 end


ptomasroos:
No its not. You can choose either to install and maintain your dependencies via linking and the react-native cli or you rely on using cocoapods. Every 3rd party component I know if uses a pointer to react as 3rd party

BluDotz:
https://cocoapods.org/?q=react <https://cocoapods.org/?q=react>

"This library is deprecated. <https://cocoapods.org/?q=react#>”

ptomasroos:
Yes it has been since 0.11.0 thats quite some time.
Dependency is being resolved locally from node_modules when running cocoapods and its being declared in the React.podspec

BluDotz:
I think you are missing the point. Because React appears in the react-native-camera Podspec, pod install installs it and gives the deprecated warning.

Pod doesn’t check if the dependency has been met by other installation means.

ptomasroos:
Well we've been using the Podfile for about 6 months and are still sucessfully doing it in a app with about 100K users.

In our own project's Podfile we declare
```
  pod 'React', subspecs: [
    'ART',
    'RCTAnimation',
    'RCTPushNotification',
    'RCTActionSheet',
    'RCTGeolocation',
    'RCTImage',
    'RCTLinkingIOS',
    'RCTNetwork',
    'RCTSettings',
    'RCTText',
    'RCTVibration',
    'RCTWebSocket',
    'DevSupport',
    'BatchedBridge',
  ], path: '../node_modules/react-native'
``` 

Which this podspec will resolve against.

BluDotz:
That’s a completely different deal entirely. Not really comparing apples with apples.

ptomasroos:
No its not, since React is a dependency of react-native-camera's podspec it can be resolved when someone declares pod 'React' in its own Podfile? How can that not make sense?

We have around 10 3rdparty podspecs being used and all declare dependency against React. I do not really see the use-case of having react-native-camera running without running the react-native app / project with cocoapods?

ptomasroos:
https://github.com/idehub/react-native-google-analytics-bridge/blob/master/react-native-google-analytics-bridge.podspec#L22
https://github.com/Microsoft/react-native-code-push/blob/master/CodePush.podspec#L18
https://github.com/bugsnag/bugsnag-react-native/blob/master/BugsnagReactNative.podspec#L19
https://github.com/rebeccahughes/react-native-device-info/blob/master/RNDeviceInfo.podspec#L16

BluDotz:
Well you seem to think you know what you’re talking about.

No doubt the majority of your "100,00 users” got the app because they are impressed by your knowledge of Cocoapod and npm dependencies.

ptomasroos:
Can you please explain to me then why it would make sense to remove the dependency in the podspec that's needed by us in react-native-camera? Since we rely on headers from the react package.

I'm definitely open to learn something new here if it makes sense! But I can't get my head around it at least.

BluDotz:
It seems to me that what you have done is create a new - additional - pod dependency, which then maps against the various entities in node-modules.

So, during “pod install”, your “local” dependency is accessed rather than the “global” Cocoapods one.

That is perfect reasonable, and I can absolutely see the purpose/ value in doing so.

But that is not the same as the default Podspec provided in the react-native-camera node_module, which ONLY has the Cocoapods React dependency; it doesn’t have your additional pod definition, and so unnecessarily loads the deprecated React Cocoapod.

ptomasroos:
That is correct if, but only if the person who uses react-native-camera does not use cocoapods for their app as well to link the React dependencies.

And I think its a fair assumption to say that the user of the podspec needs to use cocoapods as their primary dependency handler on a react-native ios project in the case one wants to use podspecs.

BluDotz:
As a first-time user of react-native-camera, what I did was follow the instructions on the GitHub README.

Of the various options offered to get the code running in Xcode, I chose to use the Cocoapods approach, because that’s what I use for my iOS development work.

However, when I ran “pod Install” the log advised that the Cocoapod was deprecated. So I thought it worthwhile suggesting a change to the Podspec to reflect that situation.

Then you decided I was wrong in suggesting this change - which is absolutely your administrative right - and closed it down.

But then it transpired that you had an additional “local” pod mapping - which was totally unknown to me or the rest of the world - against which your build resolved. So it was all good as far as you were concerned. Though actually that still left the original issue that for me pod unnecessarily downloaded an outdated, deprecated React CocoaPod.

Maybe what you should do is create a new React CocoaPod whose Podspec is merely the “local” dependency list you use. That way, we can keep the react-native-camera exactly as it is, but pod will then be able to check that the node_module dependencies are met.

Or we could just forget about the whole thing and I will get on with my explorations and you with yours.

Deal?

brunolemos:
Reopening to continue the discussion as I believe this deprecated pod should not be a dependency.

After adding the react-native-camera to an existing app with cocoapods, it installed the deprecated React pod and gave the warning. No other dependency required the React pod, only react-native-camera.

n1ru4l:
@brunolemos If your podfile is set up like in the official documentation (https://facebook.github.io/react-native/docs/integration-with-existing-apps.html) no deprecated react version will be installed.

brunolemos:
@n1ru4l interesting. That should definitely be in react-native-camera docs.

BTW @BluDotz thanks for your contribution! We will probably add this info in the docs instead of removing the dependency as it is required. If you want you can edit this PR too and I'll merge.

PR #761, opened 2017-06-20
Remove overriden createJSModules

ptomasroos:
Since of recently in master the unused createJSModules has been removed and ReactPackage's no longer contain the definition of createJSModules. There for it must be removed in order to not yield a compile error. This is breaking change.

https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8


ptomasroos:
Will close this since #821 was merged.

PR #771, opened 2017-06-28
Preview camera before trying to capture on Android

dave-irvine:
Fixes https://github.com/lwansbrough/react-native-camera/issues/770

Not sure if there is a better place to perform the preview, or why preview isn't being run by default, but in my case it isn't, and needs to be.


dave-irvine:
Ah, ok, so if your `<Camera>` element has no children, no preview is rendered, so bad things happen.

Alternatively if your `<Camera>` element does have children, but you style your element as `display:none;` worse things happen.

I don't think anything bad happens from calling `startPreview()` more than once, so this should be fairly safe to add.

PR #777, opened 2017-07-06
Add Face detection for iOS?

elliottkember:
Hi! Love this library and I've been using it for a thing. I noticed it doesn't have Face detection, although the barcode detection works great. On iOS this is part of the same barcode detection, so I thought it might be neat to add face detection as an iOS-only feature. Turned out to be much easier than I thought and it works great.


elliottkember:
Just noticed the other Face detection pull-request covers Android as well. Will close this one in favour.

Issue #689, opened 2017-05-04
Feature request: Real-time Preview Processing


kayazinc:
#### Use-case
While in camera preview mode, it is able to detect object such as "Rectangle" or "Facial" and draw a rectangle around the object.

Is it possible to achieve the use-case in javascript, such as having a new onPreivew() method that returns the image and then perform the detection algorithm on the image?

Thanks, any advice are welomed.

dotnetprofessional:
I'd like to add to this - I'm currently attempting to add an event that returns the preview image so that additional processing can be done. Taking captures is just too slow on Android (iOS is fairly quick). On Android this looks to be fairly straightforward as the byte[] is available, but for iOS its seems a bit more involved. Though any pointers are welcome, as Obj-C isn't a language I know :)

npomfret:
@dotnetprofessional  Have you tried master?  There are some changes which may improve capture performance on android that I don't think have been released yet.

The iOS version of the android byte[] is [here](https://github.com/lwansbrough/react-native-camera/blob/master/ios/RCTCameraManager.m#L590) i think.

dotnetprofessional:
@npomfret Thanks for the details - I've not tried master. I took a look at the iOS link you gave me, but that's if you do a capture. I'm looking at getting the byte[] from a preview so that its really fast. My scenario (which is odd I'll admit) is to capture the video preview and save the image as a JPG so that a webview can keep refreshing it. The idea is to provide a pysdo stream within a webview ie a web page. We can use WebRTC on Android but iOS doesn't support that.

barron9:
please take a look at anyline ocr react-native demo for this

dotnetprofessional:
@barron9 Thanks but the pricing model would bankrupt us given our product.

npomfret:
What are the rules regarding plugging commercial products on here?

barron9:
no just take the idea, their custom camera draws rectangles around objects words etc in every 10 or 20 ms. to recognize cam refreshes itself so processes while on preview

cesarve77:
+1

sibelius:
@dbishai is this https://github.com/dbishai/react-native-camera/commit/75553a6a1d09842558ad16fd6cb16abcd67cdcad performa performatic?

dbishai:
@sibelius, what does "performa performatic" mean?

ducpt2:
+ 1.

rafaelcavalcante:
This should be reopened

Related: #2984

PR #781, opened 2017-07-08
implement feature request in issue #689

dbishai:
This pull request provides the ability for camera preview frames to be accessed in real time for post-processing on Android. Admittedly, this was put together quickly for experimentation purposes and is not well tested but I believe a feature like this is very much needed.

The implementation borrows very heavily from `onBarCodeRead` and I was careful to make sure the code integrated as cleanly as possible. The major changes include refactoring the asynchronous task class used by  the barcode scanner in `OnPreviewFrame` to accept callbacks which I've implemented as anonymous functions. This allows for a "plugin" system where the asynchronous task class is essentially used as a task runner.

The event emitted contains the image encoded as a Base64 string which seems to be the recommended way of passing binary data from native to JS. A separate concurrency lock is used as well to avoid flooding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ The comprehensive camera module for React Native. Including photographs, videos,
 - Pull Requests are welcome, if you open a pull request we will do our best to get to it in a timely manner
 - Pull Request Reviews and even more welcome! we need help testing, reviewing, and updating open PRs
 - If you are interested in contributing more actively, please contact me (same username on Twitter, Facebook, etc.) Thanks!
-- We are now on [Open Collective](https://opencollective.com/react-native-camera#sponsor)! Contributions are appreciated and will be used to fund core contributors. [more details](#open-collective) 
+- We are now on [Open Collective](https://opencollective.com/react-native-camera#sponsor)! Contributions are appreciated and will be used to fund core contributors. [more details](#open-collective)
 
 #### Breaking Changes
 ##### android build tools has been bumped to 25.0.2, please update (can be done via android cli or AndroidStudio)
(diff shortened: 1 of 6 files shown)


sibelius:
it would be awesome to have a way to a bridge call another bridge

it will make possible to integrate face recognition and other video processing algorithms easily

sibelius:
anybody knows how to do the same on iOS part?

I'm trying something like this

```objective-c
AVCaptureVideoDataOutput *videoDataOutput = [[AVCaptureVideoDataOutput alloc] init];
      if ([self.session canAddOutput:videoDataOutput]) {
          // 2. Create a queue for capturing video frames
          //dispatch_queue_t captureQueue = dispatch_queue_create( "VIDEO_FACE", DISPATCH_QUEUE_SERIAL );
          
          [videoDataOutput setSampleBufferDelegate:self queue: self.sessionQueue];
          //[videoDataOutput setSampleBufferDelegate:self queue: self.sessionQueue];
          
          videoDataOutput.alwaysDiscardsLateVideoFrames = NO;
          
//          NSNumber * framePixelFormat = [ NSNumber numberWithInt: kCVPixelFormatType_32BGRA ];
//          videoOutput.videoSettings = [ NSDictionary dictionaryWithObject: framePixelFormat
//                                                                     forKey: ( id ) kCVPixelBufferPixelFormatTypeKey ];
//
          NSString *key = (NSString *)kCVPixelBufferPixelFormatTypeKey;
          NSNumber *value = [NSNumber numberWithUnsignedInt:kCVPixelFormatType_32BGRA];
          NSDictionary *videoSettings = [NSDictionary dictionaryWithObject:value forKey:key];
          videoDataOutput.videoSettings = videoSettings;
          
          [self.session addOutput:videoDataOutput];
          self.videoDataOutput = videoDataOutput;
      }
```

the delegate is not working well

sibelius:
I think this will be very slow, as we pass a lot of images through the bridge

Maybe this package could accept a class/object directly on Java to enable processing using only native code

PR #798, opened 2017-07-20
Update Android Build Tolls to 25.0.1 for @0.4

iegik:
RN 0.39.2

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -11,12 +11,12 @@ buildscript {
 apply plugin: 'com.android.library'
 
 android {
-  compileSdkVersion 23
-  buildToolsVersion "23.0.1"
+  compileSdkVersion 25
+  buildToolsVersion "25.0.1"
 
   defaultConfig {
     minSdkVersion 16
-    targetSdkVersion 22
+    targetSdkVersion 25
     versionCode 1
     versionName "1.0"
   }


rt2zz:
ah I did not realize this was against the 0.4 branch, is there a reason for this? I dont see us cutting a new 0.4 release

chirag04:
we should really be in sync with what react native is using other it will fail for most people.

Also refer: https://github.com/Microsoft/react-native-code-push/issues/919#issuecomment-315303332

rt2zz:
@chirag04 agreed that was hasty of me, I just rolled it back

@iegik if you can better explain why you desire this change it might help us understand, but for now 23 seems the better option.

iegik:
Sorry. In our company, we have an app that using RN 0.39.2 and react-native-camera@0.4
I have been tested new build on Android 25 (Build tools 25.0.1) and realized, that app do not want to build while dependency requires android-23.
So, I wrote hotfix:

```
sed -i -- "s/buildToolsVersion .*/buildToolsVersion \"25.0.1\"/" node_modules/**/build.gradle
sed -i -- "s/compileSdkVersion .*/compileSdkVersion 25/" node_modules/**/build.gradle
sed -i -- "s/targetSdkVersion .*/targetSdkVersion 25/" node_modules/**/build.gradle
```

rt2zz:
got it. I think for this case you are best off to keep the package pegged to your fork. 👍

PR #800, opened 2017-07-23
Update README.md

afaur:
Show how to make the `react-native` command available if not already installed globally.


ptomasroos:
Seems quite unlikely that someone who wants to install a react-native component hasn't walked through the guide and installed react-native

PR #809, opened 2017-07-27
Fixed README for the audio-Property for iOS

### LINKED ISSUE #805: [iOS] recorded Videos are without Audio-track (only on iOS)
We found out, that all videos, that are recorded from an iPhone (6 and 7, both @ iOS 10), has no audio. On Android-devices both video and audio are recorded well.

We use the property `captureAudio` =true as written in README.

```
      this.camera.capture({
          mode: Camera.constants.CaptureMode.video,
          captureAudio: true,
        })
```

We've also added required keys to Info.plist:

```
<key>NSCameraUsageDescription</key>
<string>This app requires access to the camera.</string>
 <key>NSMicrophoneUsageDescription</key>
 <string>This app requires access to the microphone.</string>
```

We've tested the recorded video-files on several devices (Android and iOS) and also in web browsers on desktop-PC and macbook. Those that was recorded in android contains the audio track and those on iOS not. 
Also nobody was asked if the audio-capture may be enabled.

Do we miss something else?

**Versions:**
react-native: 0.45.1
react-native-camera": master ( `"git+https://github.com/lwansbrough/react-native-camera.git"`)

itinance:
Fixes #805 

Readme states that "`captureAudio`" needs to be set to true, but in native code it is called "`audio`" instead of `captureAudio`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ Values: `Camera.constants.Aspect.fit` or `"fit"`, `Camera.constants.Aspect.fill`
 
 The `aspect` property allows you to define how your viewfinder renders the camera's view. For instance, if you have a square viewfinder and you want to fill the it entirely, you have two options: `"fill"`, where the aspect ratio of the camera's view is preserved by cropping the view or `"stretch"`, where the aspect ratio is skewed in order to fit the entire image inside the viewfinder. The other option is `"fit"`, which ensures the camera's entire view fits inside your viewfinder without altering the aspect ratio.
 
-#### `iOS` `captureAudio`
+#### `iOS` `audio`
 
 Values: `true` (Boolean), `false` (default)
 


rt2zz:
is this consistent for android and iOS?

itinance:
@rt2zz as it was already mentioned in the Readme, this setting is only available for iOS at all :)

rt2zz:
👍 thanks

manask88:
I may be missing something, but to me, captureAudio works just fine. Looking at the source code in this repo, I couldn't find the audio prop in ios. All I got was captureAudio.

itinance:
@manask88:

RCTCameraManager.m, Line 754
```
  if ([[options valueForKey:@"audio"] boolValue]) {
    [self initializeCaptureSessionInput:AVMediaTypeAudio];
  }

```

jtokoph:
Is this actually correct?

I couldn't get audio working because I was setting an `audio` prop instead of `captureAudio`. I think the component prop should be `captureAudio` and the `capture()` option should be `audio`.

jtokoph:
Evidence of this needing to be reverted:

View/component prop is called `captureAudio`: https://github.com/react-native-community/react-native-camera/blob/cebc2268a60e12175866042dae98ee05359103c0/ios/RCTCameraManager.m#L305

`capture` method option called `audio`: https://github.com/react-native-community/react-native-camera/blob/cebc2268a60e12175866042dae98ee05359103c0/ios/RCTCameraManager.m#L801

To keep the docs consistent, this should be reverted. The documentation for `capture` already mentions the `audio` option and references the `captureAudio` prop.

jgcmarins:
@jtokoph could you help us on the reverting process?

PR #811, opened 2017-07-28
Change to use external prop-types package as of React 15.5

sebandres:
From version 15.5 of react PropTypes [has moved](https://facebook.github.io/react/docs/typechecking-with-proptypes.html) into an external package 'prop-types'.

This change is to make the package compatible with React 15.5+ versions.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
-import React, { Component, PropTypes } from 'react';
+import { PropTypes } from 'prop-types';
+import React, { Component } from 'react';
 import {
   DeviceEventEmitter, // android
   NativeAppEventEmitter, // ios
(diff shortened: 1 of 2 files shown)


rt2zz:
prop-types should presumably either be a dep or a peer dep?

sebandres:
Yeah, thought about it as well but given the packages file barely states dependencies I thought the idea was to leave it to the consumer of the package to declare. 

I've added the explicit dependency.

wootwoot1234:
@lwansbrough can you accept this pull request or the https://github.com/lwansbrough/react-native-camera/pull/769 pull request.  I think they are the same.

Without this there is an error with the latest version of react and react-native.

Thanks!

rt2zz:
thanks all, I went with the other PR as I think deps makes more sense than devDeps 👍

wootwoot1234:
@rt2zz  Thanks!

PR #812, opened 2017-07-28
Fixes https://github.com/lwansbrough/react-native-camera/issues/743

iegik:


ptomasroos:
Thanks for this PR but how can the component be mounted ever in componentWillMount since the lifecycle explicitly in react is ctor -> cwm -> cdm ?

iegik:
@ptomasroos In react lifecycle will be like it was described only if all code will by synchronized and consistent. In our case we have `async componentWillMount`, it means that `componentWillMount` will continue running while next step starts (cdm).

index.js
```diff
@@ -160,10 +160,18 @@ export default class Camera extends Component {
 
     if (check) {
       const isAuthorized = await check();
-      this.setState({ isAuthorized });
+      if (this._isMounted) {
+        this.setState({ isAuthorized });
+      } else {
+        this.state.isAuthorized = isAuthorized;
```
chirag04: !

iegik: What do You propose? Delete this line? Call `setState` within `setTimeout`?
How to wait `render` is finished in RN, `requestAnimationFrame`?

PR #821, opened 2017-08-02
Remove createJSModules @override marker - RN 0.47 compatibility

radko93:
[`createJSModules` is now not required on Android](https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8) from RN 0.47. I actually can't build an app on RN 0.47 without removing this method/@ovveride marker. This is backwards compatible according to my tests in comparison to #761


rt2zz:
should not the whole method be removed rather than just the override decorator?

radko93:
This makes it compatible with old RN versions.

PR #842, opened 2017-08-17
include metadata into capture response

aksonov:
Sometimes it is needed to get some metadata, like orientation


sibelius:
is this going to work on android as well?

PR #857, opened 2017-08-25
[IOS]Fix no more audio recoded after second invoke of startCapture method

gyfelton:
## Steps to reproduce
1. After you invoke `startCapture` method for the <Camera> component, do `stopCapture` to stop video recording and get back the first video. Check that the first video has sound. (set audio=true) in the configuration of <Camera> component).
2. Now if you invoke `startCapture` again on the same component and `stopCapture` after it to get the new video.

## Expected behaviour

The second video should have audio if no setting is changed

## Actual behaviour

No sound is included in the second audio

## Environment
**Only on iOS**
react-native-camera: 10.0.0
iOS: 10
RN: 0.46.2

## Root Cause
It seems the audio input is invalidated after `stopRecording` is invoked on the `movieFileOutput` instance, also people mentioned similar issue here: https://stackoverflow.com/questions/26138207/avcapturemoviefileoutput-record-audio-only-for-the-first-video

## Fix
Re-add the audio input on every start of video recording. From the file changes you can see the author tries to *not* re-add audio input explicitly for a reason that I don't clearly see. The fix is to remove then re-add the audio input just like video input. So far in our testing no problem is raised due to this change.

### DIFF
--- a/ios/RCTCameraManager.m
+++ b/ios/RCTCameraManager.m
@@ -484,15 +484,6 @@ RCT_EXPORT_METHOD(hasFlash:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRej
 
 - (void)initializeCaptureSessionInput:(NSString *)type {
   dispatch_async(self.sessionQueue, ^{
-    if (type == AVMediaTypeAudio) {
-      for (AVCaptureDeviceInput* input in [self.session inputs]) {
-        if ([input.device hasMediaType:AVMediaTypeAudio]) {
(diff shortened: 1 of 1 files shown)


gyfelton:
This is reproducible on our production app: https://itunes.apple.com/app/id1251957849 version 1.5.0, will be fixed on v1.5.1

gyfelton:
Closing this PR as it actually introduce another bug: first frame of video can be dark due to re-init of the view.
The root cause of no audio is because of the camera still mounted when another view takes over the audio session

Issue #569, opened 2017-01-27
[Android] Battery drain issue 


kevinnguy:
Has anyone come across having a battery issue on their Android devices? When I'm using the camera for an extended period of time (10-30 minutes), my battery drains unexpectedly fast. I'm using the camera to scan barcodes and was wondering if there are any ways of improving camera/battery performance. Thanks!

### Environment
- **Node.js version**: 3.10.10
- **React Native version**: 0.40
- **React Native platform + platform version**: Android 5.0 >

### react-native-camera
**Version**: 0.5.1

cfournel:
As mentionned in https://github.com/lwansbrough/react-native-camera/issues/460 , the only way is to turn off camera by not displaying the view.

enieber:
@sibelius  this issue can it is being closed?

Issue #460, opened 2016-10-20
Camera is still active when user locks the device


tadas-subonis:
### Steps to reproduce
1. Go to a scene with CameraView so you can see the picture Preview
2. Press turn off button on the device
3. Check logs with adb logcat and you will see that camera is still active (at least on LG G3)
### Expected behaviour

Camera preview shouldn't be active
### Actual behaviour

Camera preview is active while phone is "turned off" and keeps the battery draining

Actually, when does the "preview mode" get turned off? When the component is unmounted?
### Environment
- **Node.js version**:
- **React Native version**:react-native 0.35
- **React Native platform + platform version**: react-native 0.35, Android 5.0
### react-native-camera

**Version**: master

proProbe:
We are currently killing the camera when we go to another scene. This is not optimal as the camera needs  to mount again when we enter the camera scene. This leads to serious speed drops on android (is kinda OK on ios though). Best would be if its somehow possible to paus or just turn off the camera without destroying the whole view.

baskan:
I am wondering the same question:  Actually, when does the "preview mode" get turned off? When the component is unmounted?

When our app shows the camera once we cannot make it stop capturing.

kesha-antonov:
@proProbe Hey man! How do you "kill" camera?

proProbe:
@kesha-antonov we just destroy the view. basically we just write
```
if (notCameraView) {
    return null;
} else {
    return this._renderCamera();
}
```

kesha-antonov:
Yeah, thanks! Tried that.

But I have subview:

```
render() {
  <View>
    {
      if shoulRenderCamera
        <Camera ... />
      else
        <Something else ... />
    }

  </View>
}
```

And this is not working.

carleryd:
This is still a big deal for Android devices because re-creating the Camera component whenever switching to and from the camera view in my app is very slow. This same logic run super-fast on an iPhone device however...
I am using react-native-router-flux and have a tab dedicated to the camera component and use this logic
```
      const shouldRenderCamera = this.props.redux_permission_status.camera === "authorized"
          && this.props.redux_router_scene.name === "createView"
          && (this.state.currentAppState !== "background");

      return shouldRenderCamera === true
          ? <Camera
              ref={this._setCamera}
              style={styles.preview}
              aspect={this.state.cameraConfig.aspect}
              captureTarget={Camera.constants.CaptureTarget.disk}
              captureAudio={true}
[...]

PR #859, opened 2017-08-27
Add start/stopPreview for Camera

cbrevik:
Attempting to resolve battery drain for iOS/Android by adding functionality to start/stop preview for Camera. Might also help against slow down when Camera is running in background view.

See issue #460, #569.

The recommended approach today is to not display the view when you want the camera to be inactive. This nukes the capture session, and you have to re-display the view when you want to use camera again. This can be slow for some devices.

The idea is that instead we allow the developer to choose when to pause the current capture session with `stopPreview` and `startPreview`. Can be used in conjunction with `AppState`, and other logic like navigation events, to decide when to have the camera running.

I've only tested this on my own devices, and not 100% sure yet how much impact this will have this will have on battery/perf. Opening up this PR for discussion and feedback/testing from others.

Test app here: https://github.com/cbrevik/cameraPreviewTest

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,6 +346,14 @@ Returns whether or not the camera has flash capabilities.
 
 Ends the current capture session for video captures. Only applies when the current `captureMode` is `video`.
 
+#### `stopPreview()`
+
+Stops the camera preview from running, and natively will make the current capture session pause.
+
+#### `startPreview()`
+
+Starts the camera preview again if previously stopped.
+
 ## Component static methods
 
 #### `iOS` `Camera.checkDeviceAuthorizationStatus(): Promise`
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
@@ -660,13 +660,23 @@ public class RCTCameraModule extends ReactContextBaseJavaModule
 
     @ReactMethod
     public void stopPreview(ReadableMap options) {
-        Camera camera = RCTCamera.getInstance().acquireCameraInstance(options.getInt("type"));
+        RCTCamera instance = RCTCamera.getInstance();
+        if (instance == null) return;
+
+        Camera camera = instance.acquireCameraInstance(options.getInt("type"));
+        if (camera == null) return;
+
         camera.stopPreview();
     }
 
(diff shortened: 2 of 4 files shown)


ptomasroos:
Thanks @cbrevik. This seems like valid addition!

cbrevik:
Thanks @ptomasroos! 

From my testing, the biggest impact on battery is for Android. Seems like iOS is pretty good at turning off the camera when not needed anyway.

I haven't seen that much of a difference on perf yet, but only been testing with a smaller app.

ptomasroos:
Thanks. I will review this next week! Sorry for being a bit fed up

joonhocho:
@cbrevik  @ptomasroos I think this is very important feature for production use.

tonyhb:
Seems solid. Going to fully test on our devices on Monday to ensure that the flow of `stopPreview` then `capture` works properly - this would make the picture taking "seamless".

duguying:
yes, I need this feature.

jodm:
When is this going to be merged? I need this feature too.

HarvWorks:
I would also like to know when this change will be merged,  I too need this feature for my app.

sibelius:
Great work @cbrevik 

the test app `https://github.com/cbrevik/cameraPreviewTest` is working fine!

and this will only affect users calling these 2 new functions

let me know if somebody find any regression on this one

cbrevik:
Thanks @sibelius! 🎉

sibelius:
this is available on 0.13.0 and on master as well

SirNeuman:
I see this is only available in 0.13.0 and was removed from 1.0.0

SirNeuman:
Oh i see. It's now under receiveCommand in https://github.com/react-native-community/react-native-camera/blob/5be1f0e727b292ba08ce36a87c3ff53ce2bcf880/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraViewManager.java

sibelius:
@SirNeuman feel free to send a pr for this on RNCamera

SirNeuman:
@sibelius I'd be happy to. Problem is I've only done Android development though so I don't think I'd be able to create an IOS implementation.

sibelius:
it won't be so hard to implement on iOS side either, just copy from the RCTCamera

PR #863, opened 2017-08-29
Remove nonexistant Android permission

mannol:
Android does not have RECORD_VIDEO permission. For a list of available permissions check [this link](https://developer.android.com/reference/android/Manifest.permission.html).


mikaello:
It is in the Example app as well: 
https://github.com/lwansbrough/react-native-camera/blob/ae9eab35333613c923c60cbf18482ec99e2b99b2/Example/android/app/src/main/AndroidManifest.xml#L8-L17

mannol:
@mikaello updated

PR #871, opened 2017-09-03
[captureWithOrientation()] Load SHUTTER_CLICK sound to reduce latency…

Progoogler:
… while playing during capture

Found this micro improvement in the docs while I was searching for a way to fix an issue with my on-capture sound response being halved by an unapparent latency issue. Preloading the media sound and only playing it during capture solved the issue for me. 

Perhaps this initial set and load regardless of the boolean value of `playSoundOnCapture` is worth the micro bits of improvement down the road for applications that do require sound.


ptomasroos:
Thanks!

PR #881, opened 2017-09-09
Do not render component when out of permission

iegik:
Relative to https://github.com/lwansbrough/react-native-camera/issues/743

### DIFF
--- a/index.js
+++ b/index.js
@@ -198,6 +198,7 @@ export default class Camera extends Component {
   }
 
   render() {
+    if (!this.state.isAuthorized) return this.props.children;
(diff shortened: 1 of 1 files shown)


sibelius:
should we render children or should we render null?

iegik:
@sibelius, as noted in documentation, You can pass your template trougth children. By default it equals null, isn`t it?

sibelius:
I think checking permission should be in user land instead

not in this package

ospfranco:
This should be addressed with #1076 I think we can close it

PR #910, opened 2017-09-27
Fix set after mount error

mark-friedman:
Under certain situations (which happened in our app) the camera component would attempt to set its state after the component was unmounted.  This was due to an asynchronous `check()` operation which which completed after the component was unmounted.  In most cases this is fairly benign, and not noticeable outside of dev mode, but in the case that a new instance of the camera component is instantiated immediately after the old one is unmounted, the error actually interferes with the operation of the new component.

This PR fixes the problem by making the asynchronous operation cancelable and by using the `isAuthorized` state that is set after the operation is complete to control the rendering of the component.

### DIFF
--- a/index.js
+++ b/index.js
@@ -58,6 +58,25 @@ function convertNativeProps(props) {
   return newProps;
 }
 
+// The makeCancelable function is from https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html
+const makeCancelable = (promise) => {
+  let hasCanceled_ = false;
+
+  const wrappedPromise = new Promise((resolve, reject) => {
+    promise.then(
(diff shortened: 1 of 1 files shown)


mark-friedman:
Just curious why this was closed with no comments or merge?

sibelius:
sorry about that, we fixed this on master already

we should probably move to componentDidMount as componentWillMount is deprecated already on React 16.3

jgfidelis:
@mark-friedman hello, in an effort to clean the repository that had more than fifity unmerged and old PRs, I decided to close all the older PRs that had conflict issues.

Also, we have recently deprecated RCTCamera and started to use RNCamera, so PRs for RNCamera are always welcome!

If you want to update your PR, feel free to send it again.

Thanks!

PR #914, opened 2017-09-28
Don't set background color. (Closes #913)

### LINKED ISSUE #913: Background color of camera
When `aspect` is set to `fit`, there would be some extra padding in black color for the longer dimension. I am interested in setting the color of such padding. Tried setting `backgroundColor` in camera's style but doesn't work. Any idea?

kevinresol:
Then it will respect backgroundColor in style automatically

### DIFF
--- a/ios/RCTCamera.m
+++ b/ios/RCTCamera.m
@@ -85,7 +85,6 @@
 {
   [super layoutSubviews];
   self.manager.previewLayer.frame = self.bounds;
-  [self setBackgroundColor:[UIColor blackColor]];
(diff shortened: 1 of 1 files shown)


kevinresol:
Reason of closing?

jgfidelis:
Because the pr was not merged in 6 months and I want to clean the PR log.

And I honestly do not think this would be a good solution. It solves your edge case problem but may cause problem to other people that are using rctcamera and want that black background problem.

RCTCamera is also deprecated.Ïf you send a pr with a solution that would not cause problem for others, we could merge it.

Issue #867, opened 2017-08-31
Mirror doesn't work for Videos


rad182:
Hi, It seems that `mirror` flag doesn't work on videos? the preview is mirror but the recorded video isn't. is there a workaround for this? TIA

PR #915, opened 2017-09-28
(iOS): Support for video mirror

rad182:
Added flag `mirrorVideo` for `iOS` only to output mirrored video if set to true best for front camera
issue: #867 But for iOS only


flybayer:
@rad182 why did you close this?

Issue #768, opened 2017-06-27
Android image save is very slow


wojcikiewiczm:
Hello, I want to ask about slow save performance on Android devices. For example on device Samsung Galaxy A5 (2016) with android 6 is saving taking something about 4 to 6 seconds. Is there any solution? It's really annoying because in our app We want to take several images continuosly. Thank you

RekkyRek:
I'm having the same issue.

When running in an emulator it's almost instant though. Only having this problem on hardware.

ronal2do:
+ 1

JCMais:
Same issue here.

PhillipWinder:
+1 

takes about 1-2 seconds until the photo `data`  s retrieved

```
  takePicture() {
    const options = {};
    this.camera.capture({metadata: options})
      .then((data) => console.log(data)) // <= this one here
      .catch(err => console.error(err));
  }
```

skovhus:
@wojcikiewiczm  @PhillipWinder Have you tried older versions of `react-native-camera`? And which camera options are you using?

PhillipWinder:
Camera.constants.Aspect.fill,
Camera.constants.CaptureTarget.temp,
Camera.constants.Type.back,
Camera.constants.Orientation.auto,
Camera.constants.FlashMode.off,

No - have not tested with older versions.

wojcikiewiczm:
@skovhus Yeah i did. It doesn't matter if I'm using Camera.constants.CaptureTarget.temp or Camera.constants.CaptureTarget.disk.

Borisboky:
+1

gpferreira:
+1

andrewgouin:
+1

loopingz:
From my test in the module, i think we should implement the shutterCallback:
09-30 00:53:24.297 E/RCTCameraModule: Capture takePicture.
09-30 00:53:24.536 E/RCTCameraModule: Shutter callback.
09-30 00:53:25.739 E/RCTCameraModule: Capture callback.
09-30 00:53:26.250 E/RCTCameraModule: Return switch.
So on the shutter callback we should freeze the image, with the current preview image, then probably display a small flash on the image to make the user wait while the capture callback is called

unnikrishnanc:
+1
I'm having the same issue after i takes a picture, it takes 2-3 secs to save the picture

ytallo:
+1

DhrubajitPC:
+1

[...]

PR #919, opened 2017-09-30
Freeze preview for issue #768

loopingz:
I just freeze the preview to make it more friendly user.

We could split into two promise so the capture return as soon as the shutter worked and then a getCaptureData() to retrieve the info. It would allow developers to change the screen.

### DIFF
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraModule.java
@@ -504,7 +504,7 @@ public class RCTCameraModule extends ReactContextBaseJavaModule
     }
 
     private void captureWithOrientation(final ReadableMap options, final Promise promise, int deviceOrientation) {
-        Camera camera = RCTCamera.getInstance().acquireCameraInstance(options.getInt("type"));
+        final Camera camera = RCTCamera.getInstance().acquireCameraInstance(options.getInt("type"));
         if (null == camera) {
             promise.reject("No camera found.");
             return;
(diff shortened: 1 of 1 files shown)


wassgha:
Much needed!

loopingz:
Don't hesitate to try from my repo and give me some feedback

wassgha:
Well you can't really do much with this before #859 is merged. Should also be implemented for iOS.

loopingz:
I just solved the Android issue because on IOS the speed was correct, i dont see the link with #859, i use my fork on my project and works fine

cloudsafe:
@loopingz 
hi,I have tried your codes.
Freeze the preview , this idea is very good,but it's looks like, no cancle freeze.
I think you can finished it.(I'm phper and uier,cannot do it).

loopingz:
Thanks, what do you mean by no cancelation? I m changing screen after the picture, you mean you would like the preview to restart after the promise is returned?

cloudsafe:
yeah,In my case,need take more photos or video.so I think your code can't flexible

tonyhb:
Mmm, I think if we combine this with #859 the change should just be to remove the `startPreview` call on line 535 (https://github.com/lwansbrough/react-native-camera/pull/919/files#diff-c0b5ae039b51e601ab6eba76124ab4c2R535).

[The documentation says](https://developer.android.com/reference/android/hardware/Camera.html#takePicture(android.hardware.Camera.ShutterCallback, android.hardware.Camera.PictureCallback, android.hardware.Camera.PictureCallback, android.hardware.Camera.PictureCallback)) that capturing `is only valid when preview is active` and that the preview will automatically be stopped; we can make it mandatory for users to call `startPreview` after the promise is handled with the photo to retake photos.

tonyhb:
In fact, I just removed `stopPreview` https://github.com/tonyhb/react-native-camera/commit/58740dc13ec9e829d4269897d5cbcbecd5979983 which fixes this issue.

Devs can then call `startPreview()` from #859 to retake photos.

sibelius:
check this https://github.com/lwansbrough/react-native-camera/pull/859

sibelius:
upgrade to RNCamera

PR #972, opened 2017-11-07
Fix for: Fast swiping and touching while loading crashes the app

LowieHuyghe:
Fast swiping and touching while component is being loaded can cause _camera to be null.


ospfranco:
seems like a good piece of defensive programming ;)

PR #973, opened 2017-11-08
Fixed hardcoded jpegQuality on android when saving to temp and disk

pettersolberg88:
For some reason the jpegQuality was hardcoded on Android when saving to temp and disk, but is configurable for camera roll and memory.  Fixed this by removing the hardcoded value with a available variable.


sibelius:
tks, this will improve android performance for some quality values

Issue #996, opened 2017-11-21
$500 bounty to help fix Android issues.


harrisrobin:
Hello!

I'm having a bit of an issue with this package on Android. on iOS it runs supper smoothly 👌

On Android, the performance is pretty bad and I keep getting the following 2 crashes:

## Crash 1
```
Attempt to invoke virtual method 'android.hardware.Camera$Parameters android.hardware.Camera.getParameters()' on a null object reference
```
(this now rarely happens as thanks to smarter rendering of the camera and destroying the view when exiting since I am using react-navigation, but i still see it once in a while).

## Crash 2
```
Unable to resume activity {com.heapr/com.heapr.MainActivity}: java.lang.ArrayIndexOutOfBoundsException: length=0; index=0
```

I will pay $500 to anyone who can either assist me in fixing these issues or who manage to land a PR that fixes them on Android. 
Bonus: if someone can get Android performance to improve as well, I will throw in a bonus.

Thanks,

React-Native: 0.49.3
React-Navigation: ^1.0.0-beta.7
react-native-camera: ^0.10.0

charpeni:
You can ping me @harrisrobin for the review & merge.

harrisrobin:
@grabbou Yes absolutely! I'll get it done today and get back to you here with confirmation.

I just have 2 questions:

1. When do you think would be reasonable for me to expect something I can test?
2. Is there anything I can do to help you and/or make your life easier with this?

Thanks!

grabbou:
@harrisrobin - no need to do it upfront. I think it's better to wait till I ship PR (which can happen quickly depending on the issue).

I'll dig into this today and get back to you re points 1 and 2.

RZulfikri:
Can you tell us more about your issue, like step to step when you got that issue and your android device.

harrisrobin:
@RZulfikri I will get back to you tomorrow with more details and documented steps.

grabbou:
Update from my side: I will be able to look into that on Monday. If anyone wants to start earlier, please do. Otherwise, I'll be looking into it in few days.

harrisrobin:
Update :
This is still on!
I have gotten a lot of feedback which has helped but I am still offering the $500 bounty to anyone that can land a PR that improves this library and it's stability, especially when it comes to crashes. I have not had enough time to show examples of the issues i am running into on Android but hoping to get that done by end of week :)

Thanks for following up everyone!

amscher:
[...]

Issue #960, opened 2017-10-26
app crashes while opening camera 


goel135:
i have tried by changing the version also but the problem is still persisting

jgcmarins:
@goel135 could you provide more details?

goel135:
@jgcmarins 
    "react-native-camera": "^0.10.0",
    "react-native": "0.46.4",
    "react-navigation": "^1.0.0-beta.11",

	i have already added permissions 
 <uses-permission android:name="android.permission.RECORD_AUDIO"/>
    <uses-permission android:name="android.permission.RECORD_VIDEO"/>
    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
    <uses-permission android:name="android.permission.CAMERA" />
    <uses-feature android:name="android.hardware.camera" />
    <uses-feature android:name="android.hardware.camera.autofocus" />
    <uses-permission android:name="android.permission.READ_PHONE_STATE" />

    compileSdkVersion 25
    buildToolsVersion "25.0.2"


Do tell me if anything else i should provide you

jgcmarins:
I guess it's ok.
Are you running it on simulator or device? Android? iOS?

goel135:
Android device
Android version 6.0.1

jgcmarins:
could you try to run `adb logcat` and paste the error message here?

LinFeiLong:
Hello, I'm actually having the same issue, but on iOS.

`<Camera
            captureQuality={Camera.constants.CaptureQuality.medium}
            captureTarget={Camera.constants.CaptureTarget.disk}
            type={Camera.constants.Type.front}
            ref={(cam) => {
              this.camera = cam
            }}
            style={[styles.pictureContainer, { height: Metrics.width, width: Metrics.width }]}
            aspect={Camera.constants.Aspect.fill} />
`

The app crashes if I'm using the component.

goel135:
I am getting null pointer exception and camera service failed error.

goel135:
java.lang.NullPointerException: Attempt to invoke virtual method 'android.hardware.Camera$Parameters android.hardware.Camera.getParameters()' on a null object reference
at com.lwansbrough.RCTCamera.RCTCameraViewFinder.onTouchEvent(RCTCameraViewFinder.java:349)
at android.view.View.dispatchTouchEvent(View.java:9306)
[...]

PR #1009, opened 2017-11-29
Added lock to onTouchEvent

nrakochy:
As referred to in [#996](https://github.com/lwansbrough/react-native-camera/issues/996) and  [Status React 1174](https://github.com/status-im/status-react/issues/1774), I am guessing that there are race conditions that occur when a user creates multiple touch events before `_camera` has completed instantiation. 

I added a lock which will cause the `TouchEvent` to wait if `_camera` is null. The other edits are `intellij` formatting.

Also, I had quite a difficult time getting the `Example` folder to bundle via `react-native` on Android, so I am just going to go ahead and get this PR out there to get the ball rolling. I am not sure how to go about testing it...

### DIFF
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraViewFinder.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCameraViewFinder.java
@@ -42,6 +42,7 @@ class RCTCameraViewFinder extends TextureView implements TextureView.SurfaceText
     private boolean _isStopping;
     private Camera _camera;
     private float mFingerSpacing;
+    private final Object _cameraInstanceLock;
 
     // concurrency lock for barcode scanner to avoid flooding the runtime
     public static volatile boolean barcodeScannerTaskLock = false;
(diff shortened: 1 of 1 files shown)


nrakochy:
Hi there, I wanted to follow-up on this and see if it might be reviewed? 

From what I can tell, bugs such as [#960](https://github.com/react-native-community/react-native-camera/issues/960#issuecomment-339673320) are the result of a user initiating `touchEvents` before the camera has completed initialization. 

As such, it throws a `NullPointer` exception. This PR would add a private lock object to sync the touch events with a camera that has been initialized. Thoughts?

ospfranco:
Should be fixed by correctly handling permissions on android 6.0+

nrakochy:
`!` Thanks for this!.

There is a note under `iOS` [in the README](https://github.com/react-native-community/react-native-camera#ios-onfocuschanged-event--nativeevent--touchpoint--x-y--) that mentions this, but I skipped over it, since I was zoned in on `Android`. Will drop a quick PR to update README...

PR #1020, opened 2017-12-03
Update Readme that CAMERA permission is requierd

JSandmark:
Developers are lazy, it would help to clarify to new people that normal camera permissions actually are required. Specially under the headline permissions.


sibelius:
it makes sense

tks

PR #1039, opened 2017-12-13
Remove overriden createJSModules

shiva6666:
Since of recently in master the unused createJSModules has been removed and ReactPackage's no longer contain the definition of createJSModules. There for it must be removed in order to not yield a compile error. This is breaking change.

https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8


ptomasroos:
This is handled in a backwards compatible way here 
https://github.com/react-native-community/react-native-camera/blob/160e1e5277d9adef724f0d9e7405a1a5fdcd5bdc/android/src/main/java/org/reactnative/camera/RNCameraPackage.java#L34

Issue #765, opened 2017-06-22
Cropping image taken to a certain size


AunMSheikh:
Hi guys, I am trying to get the an image with in a confined set of dimensions is that possible to do with react-native-camera. To elaborate a little more I build an overlay on top of the camera view

![screenshot_1498155755](https://user-images.githubusercontent.com/15763563/27449518-ad7ff15e-5756-11e7-80f5-2cf74644d340.png)

I am trying to get a cropped image where everything outside of the dimensions is cropped. Please let me know if there is any particular way of doing that.

RekkyRek:
I would recommend using the [imagemagick](https://www.npmjs.com/package/imagemagick) npm package as it is very simple to use.

What you are looking for would be something like this:
```javascript
im.resize({
  srcPath: <Path to image>,
  dstPath: <Output Path>,
  width:   <The new width of the image. Only using width and not height will retain aspect ratio>
}, function(err, stdout, stderr){
  if (err) throw err;
  console.log('resized image')
});
```

AunMSheikh:
I will give it a try, I had been messing with ImageEditor's crop method but the results varied from device to device which i suspect is because the dimensions of the overlay vary from device to device.

victorbadila:
Hi, I ma having this problem as well, did you manage to find a way? RN ImageEditor isn't exactly stable on Android devices, though works decent on iOS and I am unsure about using imagemagick, I need something happening on the mobile device, I used imagemagick on server side in the past, not sure it's compatible with RN.

Issue #648, opened 2017-03-25
How can I take a square photo?


c0z0:
Sorry for asking here, I know this isn't an issue per say, but I need to know how I can take square photos. I managed to get the preview to be squared but the photos taken where still rectangular.

SteveRad:
+1

xilonxiao:
+1
I don't if it applies here, but I remember once I wanted to accomplish this same effect when implementing an native android application. I had to crop the photo manually. Maybe it is also de case here.

RekkyRek:
You can use [imagemagick](https://www.npmjs.com/package/imagemagick) to crop it after capture.

bhrzk:
How about square video like instagram ? ( 720x720 for example )

sibelius:
I think this is better on stackoverflow

you can always take a photo and crop it later on

check this https://stackoverflow.com/questions/42331823/crop-image-with-react-native

ahmetilhann:
I find solution with `ImageEditor`
```
import React, { Component } from 'react';
import { Text, View, TouchableOpacity, Dimensions, StyleSheet, ImageEditor, ToastAndroid } from 'react-native';
import { RNCamera } from 'react-native-camera';

const { width } = Dimensions.get('window');

class CameraIndexScene extends Component {
    constructor(props) {
        super(props);

        this.state = {
            flashMode: RNCamera.Constants.FlashMode.off,
            type: RNCamera.Constants.Type.back,
            loading: false,
        };

        this.takePicture = this.takePicture.bind(this);
    }
    takePicture() {
        ToastAndroid.show('Please waiting...', ToastAndroid.LONG);
        if (this.camera) {
            this.setState({ loading: true });
            const options = { quality: 1, base64: false };
            this.camera.takePictureAsync(options)
                .then((data) => {
                    const cropData = {
                        offset: {
                            x: 0,
                            y: 0,
                        },
                        size: {
                            width: data.width, 
                            height: data.width,
                        },
                    };
                    ImageEditor.cropImage(
                        data.uri, 
                        cropData, 
                        (successURI) => { 
                            console.log(successURI);
                        }, 
                        (error) => {
                            console.log(error.message);
                        },
                    );
                });
        }
    }

    render() {
        return (
            <View style={styles.container}>
                <RNCamera
                    ref={(ref) => {
                        this.camera = ref;
                    }}
                    style={styles.preview}
                    type={this.state.type}
                    flashMode={this.state.flashMode}
[...]

Issue #37, opened 2015-05-23
Captured image does not crop image to  <Camera> viewfinder's ratio


ksokhan:
I expected the outputted image to have the same width/height ratio as the camera viewfinder. However, the returned image is always the full capture of the camera, irrespective of the camera viewfinder's dimensions.

lwansbrough:
This is intended - I could see this being a useful setting though.

ksokhan:
Ok thanks! I'll see if I can set up a PR for this.

nickgerig:
This would be a great feature. I think it's also about user expectations and even privacy. A user will frame the photo based on the preview so any subsequent action on the actual photo needs to be re-evaluated by the user if it's different to the preview. eg. if you are going to upload a photo

Are the cropping rules the same on iOS and Android? It could be that the cropping of the actual photo is handled outside of this library if the preview cropping rules were consistent cross platform and documented.

PR #1057, opened 2017-12-20
Crop to preview

ejkkan:
The cropping of passing the "preview" parameter to captureQuality did not work as expected. 
Since one should still have the choice of setting a certain image quality when wanting the image cropped aswell, a new property has been added so that one can distinguish between the two.

Property = cropToPreview={true} // default false

If true, the image will be cropped to the exact ratio of the camera preview window for both Android and IOS.

should be fixing some issues like: #765  #37  #648


mckn:
@lwansbrough is there anything more we can do to get this PR accepted? I will update the docs, but appart from that?

ejkkan:
@sibelius Can anything more be done to have this PR merged? We greatly need this functionality. 
Thank you

ios/RCTCameraManager.h
```diff
@@ -72,11 +71,11 @@ typedef NS_ENUM(NSInteger, RCTCameraTorchMode) {
 @property (nonatomic, assign) NSInteger videoTarget;
 @property (nonatomic, assign) NSInteger orientation;
 @property (nonatomic, assign) BOOL mirrorImage;
+@property (nonatomic, assign) BOOL cropToPreview;
 @property (nonatomic, strong) NSArray* barCodeTypes;
 @property (nonatomic, strong) RCTPromiseResolveBlock videoResolve;
 @property (nonatomic, strong) RCTPromiseRejectBlock videoReject;
 @property (nonatomic, strong) RCTCamera *camera;
-@property (nonatomic, assign) BOOL cropToViewport;
```
sibelius: is this a breaking change ?

mckn: Well, it is compared to the latest in master. But I don't think the version with that cropToViewport is released as a package?

We replaced the cropToViewport with cropToPreview and also implemented it so it works for all orientations for both android and iOS.

The old implementation was merged into master in this commit:
https://github.com/react-native-community/react-native-camera/commit/39f19cd7c7b39689f32af7ad08a26e248f0742f9

@sibelius

sibelius: well, it is time for a new release

@lwansbrough can we help us with that?

@mckn fix the conflicts and let's merge this, we can test more on master

sibelius:
@Ignigena what do you think of this?

Ignigena:
@sibelius the functionality looks to be the same judging by the code itself (don't have time to spin it up locally right now)

Definitely all for consistency between Android and iOS implementations if the change of naming gets us there. The previous was used since iOS doesn't have a "preview" capture quality mode and to try to replicate the existing Android preset name and function as much as possible.

mckn:
Well, that is one of the things that we did not really understand. Did we have a capture quality that was preview? If so how did that work? This feature crops the captured image to what you see inside the preview. 

The image still keeps the previous selected capture image quality.

Example: I can capture an image with high quality and still crop it to a square if we only display a square to our end user in the preview.

Ignigena:
@mckn As I understood it, previously the `preview` option on Android would automatically set the device capture quality to the closest match to the viewport. My changes on iOS only brought software cropping to match the viewport but left the device capture quality as high without attempting to match the resolution of the viewport on capture.

Seems with your changes in this PR things are a bit more consistent between platforms which is definitely 👍

sibelius:
great work guys

tks

sunweiyang:
Thanks for this! However I'm experiencing an issue on the latest master, which includes this commit. 

On iOS it works great, but on Android I'm seeing that it only crops the left/right sides off. In other words, I've defined my preview to be a square, but what I get instead is a tall skinny rectangular captured image.

Anyone else seeing this?

EskelCz:
Doesn't work for me, on Xiaomi Redmi 6A. The preview is not even respecting the width and height in style, on RNCamera component.

PR #1064, opened 2017-12-28
isNull check for Camera options on Android

timotew:
Somehow React allows null as a value for ReadableMap on Android.
This cause a lots of internal errors sometimes app crashing.


timotew:
@sibelius  Hi there! How do we proceed with this PR.

ospfranco:
@timotew there seems to be a typo here:

`_sensorOrientathionChecker.registerOrientationListener(new RCTSensorOrientationListener() {`

should be:

`_sensorOrientationChecker`

timotew:
@ospfranco  Thanks a lot!

sibelius:
@timotew isn't better to not send this parameters? instead of sending null values?

timotew:
@sibelius  Yes it is and that is ideal. But in most cases the parameters like location is null at initial state until the location is available.

Issue #246, opened 2016-04-04
1.0 Changes, TODO, and RFC


rt2zz:
**So far the following changes are slated for 1.0**
- Promise api instead of callbacks
- Android Support
- `mirrorImage` option
- `captureQuality` option
- `target: temp` option
- Capture to return object with metadata instead of a simple string value

**TODO: release blockers**
- [x] [Change RCT prefix](https://github.com/lwansbrough/react-native-camera/issues/283) (presumably to RN)
- [x] Android Video
- [x] Android support child views

**TODO**
- [x] Android Barcode
- [x] iOS - ImageStore support

Huge thanks to all of the contributors including @corbt @rpopovici @dhrrgn @niborb @cdimascio @sunnylqm ! I apologize if I left anyone's name off this list, github autocomplete is not being very helpful today.

Please weigh in with your thoughts. It feels like we have a solid community behind the project now, and if we coordinate efforts a bit we can get to 1.0 in the next month. :balloon: /cc @lwansbrough

lwansbrough:
I wanna send out a huge thanks to everyone who's made contributions to this repository as we move towards a v1. My work hasn't been too deeply involved in React Native lately so it's been difficult for me to find time.

Regarding the proposed changes: the current additions look good. I'm in favour of returning an object for the capture response.. Should have been doing that from the get go. Obviously sub views need to be prioritized for v1. I think barcodes on Android would be a nice to have but for me is the third most important thing there, followed by ImageStore support.

As we continue development I think it's important to highlight a few things:
- Be sure you'd use the code you're writing in production. We don't have tests so extra care should be taken care here and during merge requests. So far this has been pretty good but I've taken part in a couple discussions where a fix for one thing breaks another.
- Look at other modules when looking for naming patterns. Predictability is really important for a good library.
- Where possible, challenge yourself to write for both platforms, or at least be sure that your API would work for both platforms if someone wanted to extend your API.

Again, I'm super appreciative of those who are helping develop this module. Keep up the great work. Hopefully I can get some time at some point to review everything that's changed and get up to date and help out. Until then I'll try to be around more to at least answer questions. Cheers!

chirag04:
@rt2zz idk about the roadmap progress but can you publish an 1.0 alpha atleast on npm. Right now rn 0.26 is having a breaking change which is fixed on rn-camera master.

dhrrgn:
[...]

PR #1071, opened 2017-12-31
Add Camera implementation using CameraView on Android

sibelius:
This is most a copy of expo camera module

this still half baked, missing javascript part and package and modules
fixes

see #246

the idea is to call this as RNCamera, and keep the RCTCamera as the old one


android/src/main/java/org/reactnative/camera/CameraViewManager.java
```diff
@@ -41,10 +41,10 @@ public String toString() {
     }
   }
 
-  private static final String REACT_CLASS = "ExponentCamera";
+  private static final String REACT_CLASS = "RNnentCamera";
```
kraffslol: RNnentCamera? 😄

ptomasroos:
I'm on vacation will check on Monday!

sibelius:
we can go in 2 directions on this PR:

- [ ] use the interface closer to expo camera api, 
- [ ] or make this closer to current react-native-camera interface

going closer to expo camera api will make easier to keep expo camera modules changes in this package, and also easy to port some improvements of this package to expo itself. We could create a compatible interface from javascript to make easy to upgrade to the new implementation

ospfranco:
with the next release of rn (0.52) the createJSModules should no longer be supported, as we are going forward with this, we should also take care of it, #1039

wassgha:
@sibelius just to help with the decision, some of the features I've noticed are missing from react-native-camera but are present in the expo camera api are:
- Manual zoom
- Manual focus
- Face detection
- Separate still photo / video recording endpoints
We'd totally love to see these features in react-native-camera.

afilp:
@sibelius It is possible to first release a new npm version that handles the frequent crash problems on Android? I saw that you created a PR for this, can this be merged/released and then enhance the features? Thanks!

sibelius:
@afilp can you at least run both examples to make sure everything is working fine on android?

wassgha:
There's a `cameraview` dependency missing, I am getting
```
* What went wrong:
A problem occurred configuring project ':app'.
> Could not resolve all dependencies for configuration ':app:_debugApkCopy'.
   > Could not find com.github.react-native-community:cameraview:df60b07573.
     Required by:
         project :app > project :react-native-camera
```

sibelius:
we need to improve README

we need jitpack to run properly

https://github.com/react-native-community/react-native-camera/blob/master/RNCameraExample/android/build.gradle#L27

mckn:
What was the final decision regarding this PR?

sibelius:
@mckn I've decided to add this new implementation as RNCamera, so this won't break anything for old users, that will keep to use RCTCamera

We also need to bring the RNCamera implementation to ios part based on this https://github.com/react-native-community/react-native-camera/issues/1084

let me know if you wanna help us doing this

We can chat more about it on expo slack (https://slack.expo.io/)

DevarakondaV:
For some reason
`react-native run-android`

throws several errors of the same variation.

`C:\Users\devar\Documents\React\SnapMe\node_modules\react-native-camera\android\src\main\java\org\reactnative\facedetector\tasks\FileFaceDetectionAsyncTask.java:81: error: incompatible types: String cannot be converted to Throwable
      mPromise.reject(ERROR_TAG, "The file does not exist. Given path: `" + mPath + "`.");`

and 

`C:\Users\devar\Documents\React\SnapMe\node_modules\react-native-camera\android\src\main\java\org\reactnative\camera\events\BarCodeReadEvent.java:27: error: no suitable method found for init(int)
    event.init(viewTag);
         ^
    method Event.init(int,long) is not applicable
      (actual and formal argument lists differ in length)
    method BarCodeReadEvent.init(int,Result) is not applicable
      (actual and formal argument lists differ in length)`

I've looked at the implementation for the first error in Promise.java and Promise.reject is an overloaded function. For some reason it still seems to fail. I also tried passing the second argument as a Throwable. This leads to a different error. 

The second error also doesn't make much sense because the Even.init function is also implemented.

Any insights to what is causing this?

sibelius:
we decided to go in this way

- [x] use the interface closer to expo camera api,

so it will be easy to move from and to expo

wassgha:
Could you confirm if `recordAsync` works in this version? It did not work for me.

sibelius:
open an issue, and we can track this

PR #1076, opened 2018-01-06
Add permission handling for android 6+

ospfranco:


src/index.js
```diff
@@ -181,7 +201,11 @@ export default class Camera extends Component {
     const style = [styles.base, this.props.style];
     const nativeProps = convertNativeProps(this.props);
 
-    return <RCTCamera ref={CAMERA_REF} {...nativeProps} />;
+    if(this.state.isAuthorized) {
```
sibelius: can we have something like this one https://github.com/moaazsidat/react-native-qrcode-scanner/blob/master/index.js#L203

`pendingAuthorizationView` and
`notAuthorizedView`

Issue #1094, opened 2018-01-12
Does not work on android below 6.0


punksta:
It say camera is not authorized on android below 6.0 always.

Issue is here:
https://github.com/react-native-community/react-native-camera/blob/bc975b4bac49a31b8ff885efddd36c456da7c262/src/Camera.js#L188-L197

According to documentation:
>On devices before SDK version 23, the permissions are automatically granted if they appear in the manifest, so check and request should always be true.

so if sdk is less 23, PermissionsAndroid.request returns `Promise<boolean>`

We should check android 6.0 result separately.

PR #1095, opened 2018-01-12
fix camera permission handle in android below 6.0

punksta:
fix https://github.com/react-native-community/react-native-camera/issues/1094

### DIFF
--- a/src/Camera.js
+++ b/src/Camera.js
@@ -191,10 +191,18 @@ export default class Camera extends Component {
         message: this.props.permissionDialogMessage,
       });
 
+      // On devices before SDK version 23, the permissions are automatically granted if they appear in the manifest, 
+      // so check and request should always be true.
+      // https://github.com/facebook/react-native-website/blob/master/docs/permissionsandroid.md
+      const isAuthorized = Platform.Version >= 23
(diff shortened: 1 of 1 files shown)


src/Camera.js
```diff
@@ -191,10 +191,18 @@ export default class Camera extends Component {
         message: this.props.permissionDialogMessage,
       });
 
+      // On devices before SDK version 23, the permissions are automatically granted if they appear in the manifest, 
+      // so check and request should always be true.
+      // https://github.com/facebook/react-native-website/blob/master/docs/permissionsandroid.md
+      const isAuthorized = Platform.Version >= 23
+		      ? granted === PermissionsAndroid.RESULTS.GRANTED
+          : granted === true;
```
sibelius: can we do something simpler, like this:

`granted === PermissionsAndroid.RESULTS.GRANTED || granted === true` ?

punksta: with `Platform.Version >= 23` it looks more strict for me.

sibelius:
it makes sense, tks

punksta:
Thanks for fast reply!

HexiChen:
It helps me,ths a lot

Issue #1086, opened 2018-01-10
Could not find com.github.react-native-community:cameraview:df60b07573


rcorrie:
I am getting the following error running `react-native run-android` with the latest code on master.

```
* What went wrong:
A problem occurred configuring project ':app'.
> Could not resolve all dependencies for configuration ':app:_debugApkCopy'.
   > Could not find com.github.react-native-community:cameraview:df60b07573.
     Required by:
         project :app > project :react-native-camera

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
```

sibelius:
Check README, now you need jitpack

nicemanis:
I added `maven { url "https://jitpack.io" }` to the android buildscript repositories

sibelius:
what is the error with jitpack?

nicemanis:
there is no error with jitpack, it works with react-native v0.52.0, but not with v0.51.0

sibelius:
What is the erro with 51?

nicemanis:
The same which opened this issue:
`Could not find com.github.react-native-community:cameraview:df60b07573`

It seems that rn v0.51.0 ignores the jitpack

nicemanis:
Ahh, the react-native-camera v0.6.0 also doesn't work on rn v0.52.0, I have the PropTypes error after successful build. At first I couldn't see it because rn v0.52.0 also breaks react-native-vector-icons

nicemanis:
Note that I'm forcing all subprojects to use android build tools version 25.0.2:
```
subprojects {
    afterEvaluate {project ->
        if (project.hasProperty("android")) {
            android {
                compileSdkVersion 25
                buildToolsVersion '25.0.2'
            }
        }
    }
}
```

sibelius:
We use android 26 now

nicemanis:
I checked the ./node_modules/react-native-camera/android/build.gradle:
```
buildscript {
  repositories {
    jcenter()
  }

  dependencies {
    classpath 'com.android.tools.build:gradle:1.2.3'
  }
}

apply plugin: 'com.android.library'

android {
  compileSdkVersion 23
  buildToolsVersion "23.0.1"

  defaultConfig {
    minSdkVersion 16
    targetSdkVersion 22
    versionCode 1
    versionName "1.0"
  }
  lintOptions {
    abortOnError false
    warning 'InvalidPackage'
  }
}

repositories {
  mavenCentral()
}

[...]

PR #1106, opened 2018-01-16
README update for com.android.support build issue

nicemanis:
This update adds the solution to the README mentioned [here](https://github.com/react-native-community/react-native-camera/issues/1086).

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,7 +450,7 @@ To see more of the `react-native-camera` in action, you can check out the source
 
 ## Q & A
 
-#### meta-data android 26	
+#### Q: When I try to build my project, I get following error:
 ```
 AndroidManifest.xml:25:13-35 Error:
      Attribute meta-data#android.support.VERSION@value value=(26.0.2) from [com.android.support:exifinterface:26.0.2] Android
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -450,34 +450,27 @@ To see more of the `react-native-camera` in action, you can check out the source
...
 ```
+#### A: As the error message hints `com.android.support:exifinterface:26.0.2` is already found in `com.android.support:support-v4:26.0.1`
+To fix this issue, modify your project's `android/app/build.gradle` as follows:
+```Gradle
+dependencies {
+    compile (project(':react-native-camera')) {
+        exclude group: "com.android.support"
+    }
 
-Add this to your AndroidManifest.xml:
```
sibelius: can we keep this lines? as well

nicemanis: I'm not sure if they are necessary as they didn't really fix the issue, at least for me.
The issue here is a conflict between two versions of `com.android.support:exifinterface`
But if there is different case where the changes to the manifest would fix the issue, then I can leave in also the manifest lines

nicemanis: I'm not sure if they are necessary as they didn't really fix the issue, at least for me.
The issue here is a conflict between two versions of `com.android.support:exifinterface`
But if there is different case where the changes to the manifest would fix the issue, then I can leave in also the manifest lines

sibelius: I think they are 2 different kind of problems

and we should provide solutions for both

nicemanis: Ok, then I will leave also those lines, I will also add this `exclude group: "com.google.android.gms"` line which is necessary if you use react-native-firebase

sibelius: tks

README.md
```diff
@@ -450,12 +450,40 @@ To see more of the `react-native-camera` in action, you can check out the source
...
+      android:name=".MainApplication"
+      android:allowBackup="true"
+      android:label="@string/app_name"
+      android:icon="@mipmap/ic_launcher"
+      android:theme="@style/AppTheme"
+      tools:node="replace"
+    >
+```
+
 #### Q: When I try to build my project, I get following error:
```
sibelius: remove `Q:`

sibelius:
@nicemanis great work on this, tks

nicemanis:
@sibelius thanks

PR #1118, opened 2018-01-19
Qr code

martonczirjak:
Please add a prop to stop qr code reading, stopPreview doesnt stop reading the qr code and my camera still read it.


sibelius:
there are too many commits in this PR

can you create a new PR from master?

PR #1119, opened 2018-01-19
Return picture size in capture result

paiou:
This PR add width and height properties in JSON returned by capture method (iOS and Android).


paiou:
I have updated README accordingly.
This is not a breaking change, only additional properties in returned JSON.

sibelius:
tks

PR #1121, opened 2018-01-19
Add setZoom API method

bluedaniel:
This adds `setZoom` so you can additionally zoom the camera.

Example usage with a slider.

```js
<Camera ref={c => { this.camera = c; }} />

<Slider
  value={1}
  minimumValue={1}
  maximumValue={16}
  onValueChange={value => this.camera.setZoom(value)}
/>
```

### DIFF
--- a/android/src/main/java/com/lwansbrough/RCTCamera/RCTCamera.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/RCTCamera.java
@@ -362,6 +362,22 @@ public class RCTCamera {
         }
     }
 
+    public void setZoom(int cameraType, int zoom) {
+        Camera camera = this.acquireCameraInstance(cameraType);
+        if (null == camera) {
+            return;
+        }
+
+        Camera.Parameters parameters = camera.getParameters();
+        int maxZoom = parameters.getMaxZoom();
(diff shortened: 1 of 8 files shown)


sibelius:
@bluedaniel this looks good

can somebody test this on real devices on ios and on android, so we can merge?

bluedaniel:
@sibelius this is an iOS only PR.

Java isn't my strong suit but I can give it a go.

bluedaniel:
@sibelius I added some Android code for zooming but I don't have a device to test it on :man_shrugging:

Small gist to test:
https://gist.github.com/bluedaniel/436ad3130b07dd12db2d3f3ec7ded115

sibelius:
@bluedaniel I think it looks good

can u rebase?

tirrorex:
What would be great would be to wrap a PanResponder in react-native-camera.
Basically one could set a zoom to false or with a different value (to zoom faster for example) in the props.
This way we could have the same implementation on both ios and android (that is actually what we have in our apps, but with the native camera wrapped by ourselves).
Thoughts?

sibelius:
@tirrorex I think this can be in user land

or in a recipe in this repo

PR #1126, opened 2018-01-20
Fix performance issue on Android: add option to clear window background when camera starts

paiou:
I encountered a big performance issue on Android (frame rate dropping from 60 to 10 FPS) while using a window background drawable to display as splash screen until the app is loaded.
The problem doesn't occur on all devices, but it occurs on Samsung Galaxy S6 for example.

This PR adds a clearWindowBackground prop, if set to true window background drawable will be set to null once the camera is started, hence fixing the performance issue.
By default, the prop is set to false to preserve current behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,6 +347,14 @@ from javascript.
 
 If set to `true`, the device will not sleep while the camera preview is visible. This mimics the behavior of the default camera app, which keeps the device awake while open.
 
+#### `Android` `clearWindowBackground`
+
+Values:
+`true`
+`false` (default)
+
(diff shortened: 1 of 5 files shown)


sibelius:
tks

witalobenicio:
@paiou how can you drop the fps rate?

paiou:
I don't know what happened behind the scene, just noticed that having a window background made the camera very slow on some devices. Passing clearWindowBackground=true only remove the window background (which is hidden behind the camera anyway), and everything is smooth...

PR #1130, opened 2018-01-21
feature(rncamera): enable record on RNCamera on android, fix #1101

### LINKED ISSUE #1101: RNCamera.recordAsync not working
### Steps to reproduce
1. Implement `RNCamera.recordAsync` on the RNCameraExample
2. Run the example and try recording a video

### Expected behaviour
`RNCamera.recordAsync` should return a `Promise` that resolves as soon as `stopRecording` is called or the video duration is exceeded.

### Actual behaviour
`RNCamera.recordAsync` never resolves

### Environment
- **Node.js version**: 5.6.0
- **React Native version**: 0.52.0
- **React Native platform + platform version**: Android 6.0

### react-native-camera
**Version**: "master"

sibelius:


wassgha:
This is exactly the code I had in a PR I was working on, but I kept getting "Starting video recording failed. Another recording might be in progress" (which I also get with this PR). Not sure what the problem is but it seems like `super.record` is not returning true...

sibelius:
Could u open an issue so we can track this? The only different from expo is that we do t check for audio permissions

We can do that on js code

PR #1150, opened 2018-01-23
Chore/update version

Rawphs:


Rawphs:
oops, that was an accident haha

PR #1153, opened 2018-01-23
Get rid of cocoapods dependency

jgfidelis:
Now react-native-camera ios does not need for a pod install to work properly. Just follow steps described in the readme.md (Installing GMV in iOS).


tirrorex:
I can confirm the build fail.
What are your plans for the future?
Lazy me removed the RN and Detector files from build to not have to build unecessary dependency but that's not something i can push

PR #1154, opened 2018-01-23
Fix startPreview and stopPreview on Android

paiou:
Currently on Android, after calling `startPreview` the camera preview is not initialized correctly, leading to 2 bugs:

- After a capture, the preview is not displayed anymore and continue to show the last picture taken (the preview texture is set to null after capture, but not set back when calling `startPreview`)
- After calling `stopPreview` then `startPreview`, bar code detection does not work anymore (the preview callback is not set back by `startPreview`).

`stopPreview` and `startPreview` methods are called on the native module, and don't have access to camera view instance. Consequently, native camera methods `startPreview` and `stopPreview` are called, but important parts of code are missing, especially re-setting preview callback and preview texture to the current ViewFinder.
This PR use `UIManager.dispatchViewManagerCommand` to send `startPreview` and `stopPreview` commands to the current view instance, which delegates execution to the current ViewFinder instance methods that are called to start the camera initially and take care of the proper initialization.

### DIFF
--- a/src/Camera.js
+++ b/src/Camera.js
@@ -6,12 +6,14 @@ import {
   NativeModules,
   Platform,
   StyleSheet,
+  findNodeHandle,
   requireNativeComponent,
   ViewPropTypes,
   PermissionsAndroid,
   ActivityIndicator,
   View,
   Text,
+  UIManager,
 } from 'react-native';
 
 const CameraManager = NativeModules.CameraManager || NativeModules.CameraModule;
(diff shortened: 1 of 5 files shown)


src/Camera.js
```diff
@@ -300,21 +302,21 @@ export default class Camera extends Component {
...
   }
 
   stopPreview() {
     if (Platform.OS === 'android') {
-      const props = convertNativeProps(this.props);
-      CameraManager.stopPreview({
-        type: props.type,
-      });
+      UIManager.dispatchViewManagerCommand(
+        findNodeHandle(this),
```
sibelius: can we save ref like this commit https://github.com/expo/expo-sdk/commit/c8f80cc563a5415c84c7f8dd139a85fa5bdc6c97?

paiou: Thanks for the feedback, thanks for your reactivity, and your great contribution in general :)

PR #1163, opened 2018-01-26
[FIX] mBarCode undefined in BarCodeReadEvent

seblau:


android/src/main/java/org/reactnative/camera/events/BarCodeReadEvent.java
```diff
@@ -24,7 +24,7 @@ public static BarCodeReadEvent obtain(int viewTag, Result barCode) {
     if (event == null) {
       event = new BarCodeReadEvent();
     }
-    event.init(viewTag);
+    event.init(viewTag, barCode);
```
sibelius: mBarCode?

seblau: The parameter name in this method is `barCode`. `mBarCode` gonna bet set in the init method

seblau:
yes, I did. its working

sibelius:
Tks, it is correct

Issue #1161, opened 2018-01-25
`Camera` no propType for native prop `RNCamera.accessibilityLabel` of native type `String`


emmajam:
### Which implementation are you using

RNCamera

### Steps to reproduce
1.  "react-native-camera": "git+https://git@github.com/react-native-community/react-native-camera" 
`npm install` in package.json

2. react-native link react-native-camera
3. react-native run-android OR run in android studio (either or)
4. error pops up

### Expected behaviour
Load app successfully, no errors

### Actual behaviour
Error occurs:
![screenshot_20180125-164334](https://user-images.githubusercontent.com/21029500/35400547-fa703ae4-01ee-11e8-89e3-eb1530cfc9a5.png)

### Environment
- **Node.js version**: v9.2.1
- **React Native version**: 
react-native-cli: 2.0.1
react-native: 0.43.0
- **React Native platform + platform version**: Android 7.0

### react-native-camera
**Version**: "git+https://git@github.com/react-native-community/react-native-camera"

seblau:
Should be fixed by https://github.com/react-native-community/react-native-camera/pull/1165

emmajam:
thanks @seblau 🎉

PR #1165, opened 2018-01-27
Fix issue with running on android. Add native props

seblau:
Running on android 7.0 with react-native 0.43.4 leads to an error that various props are not defined in js, but are present in the native implementation. After I complete the native only props for RNCamera it worked.


seblau:
Also see https://github.com/react-native-community/react-native-camera/issues/1161

PR #1179, opened 2018-01-30
Improve android performance with Expo picture resolve implementation.

jgfidelis:
Get rid of MutableImage.
Return base64 only if asked via options map.


android/src/main/java/org/reactnative/MutableImage.java
```diff
@@ -1,282 +0,0 @@
-package org.reactnative;
```
sibelius: is this used on RCTCamera?

jgfidelis: No. RCTCamera has its own MutableImage file in its own package.

PR #1190, opened 2018-01-31
Add captureAudio flag for RNCamera as a prop

jgfidelis:


jgfidelis:
User can pass captureAudio = {true} if it wants to capture audio and then RNCamera can request for audio and video if necessary.

If captureAudio is false, it will only request for permission for camera (iOS)

newlifepop:
Hi, just a question, I've been working on it for quite a while now but couldn't get it solved. Also asked here and on SO. I have no problems when capturing a photo, but when trying to record videos, videos are not being recorded. Could you maybe share the part of your code where to take a video? Thanks!

PR #1235, opened 2018-02-13
Incorrect placement of plugin.

TitanKing:
I believe this is incorrectly used here. This will cause various issues, especially using it with Android Studio. This is also the reason you have to specify the QA compile error. The plugin `com.android.support:exifinterface:+` should at the very least be used but then not even, rather have it as part of the installation process to include the dependency in the users build.gradle.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -42,6 +42,7 @@ dependencies {
   compile "com.google.zxing:core:3.2.1"
   compile "com.drewnoakes:metadata-extractor:2.9.1"
   compile 'com.google.android.gms:play-services-vision:+'
+  compile "com.android.support:exifinterface:+"
 
   compile 'com.github.react-native-community:cameraview:d5d9b0d925494ef451ce3eef3fdf14cc874d9baa'
 }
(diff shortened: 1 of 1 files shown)


jgfidelis:
But if you take this off, wont it crash in ResolveTakenPictureAsync when it tries to reach for the exif interface class?

sibelius:
I think we should improve the docs make people exclude `com.android.support` from react-native-camera

sibelius:
let's keep exifinterface:+

and see if it works better

TitanKing:
Sounds good, so far with my test, the Q&A would not even be needed if we use `exifinterface:+`, I also noted that very popular libraries does the same. 

Should I create new pull request with the fix?

sibelius:
u can do in this pr

TitanKing:
Ok done.

jgfidelis:
I will see the impact of this in some of our apps that use rncamera. Thanks

jgfidelis:
BTW, we are very noobs in regarding to Android development so any help with it is always appreciated. The `+` tip looks very promising.

TitanKing:
Nice, yeah, I doubt anyone would have any more of the compile errors as addressed in Q&A, which does not work on API 27 anymore anyway, so I think this will all be good now 👍 

Thanks for your attention to details man, what a useful plugin.

Issue #1145, opened 2018-01-23
Could not find method api() for arguments


condorman:
1. Follow installation steps https://github.com/react-native-community/react-native-camera
2. Install package: "react-native-camera": "git+https://git@github.com/react-native-community/react-native-camera.git"
3. compile: react-native run-android

**Problem:**
A problem occurred evaluating project ':react-native-camera'.
> Could not find method api() for arguments [com.facebook.react:react-native:+] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.

**Fix:** 
 in: react-native-camera/android/build.gradle  I change this line:

dependencies {
  api 'com.facebook.react:react-native:+'
  api "com.google.zxing:core:3.2.1"
  api "com.drewnoakes:metadata-extractor:2.9.1"
  api 'com.google.android.gms:play-services-vision:+'
  api "com.android.support:exifinterface:26.0.2"

  api 'com.github.react-native-community:cameraview:df60b07573'
}

**to**

dependencies {
  compile 'com.facebook.react:react-native:+'
  compile "com.google.zxing:core:3.2.1"
  compile "com.drewnoakes:metadata-extractor:2.9.1"
  compile 'com.google.android.gms:play-services-vision:+'
  compile "com.android.support:exifinterface:26.0.2"

  compile 'com.github.react-native-community:cameraview:df60b07573'
}

I hope I have been helpful, thank you! great plugin!

PR #1269, opened 2018-02-20
Update build.gradle to support latest Gradle-plugin/wrapper

SimonErm:
Fixes the problem that occurs when an app with the latest gradle-wrapper and plugin tries to include the module.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -38,11 +38,11 @@ repositories {
 }
 
 dependencies {
-  compile 'com.facebook.react:react-native:+'
-  compile "com.google.zxing:core:3.2.1"
-  compile "com.drewnoakes:metadata-extractor:2.9.1"
-  compile 'com.google.android.gms:play-services-vision:+'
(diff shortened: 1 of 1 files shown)


sibelius:
did u test it on debug and release versions?

SimonErm:
I recognized that the occuring problem was caused by a gradle-plugin i used and not by this, but yes it works on debug and release

punksta:
`api` won't work on Android Plugin for Gradle 2.x.x . Most of apps are still use version 2. We will have issues like this: https://github.com/react-native-community/react-native-camera/issues/1145

facebook still use version [2.2.2](https://github.com/facebook/react-native/blob/master/build.gradle#L9) in react-native plugin

`compile` keyword is supported by new android plugin (3.x.x). So I don't see reasons to merge it right now.

punksta:
Please reopen pl when plugin version 3.x.x will be supported by most of apps.
thanks

PR #1278, opened 2018-02-22
Add ability to force portrait orientation on IOS

Michaelvilleneuve:
iOS uses exif based orientation, this causes orientation failure when displaying image on a web page.
Indeed web browsers don't read exif when displaying images and iOS images appear as landscape.

The forceUpOrientation boolean fixes this by rotating the actual image data and therefore does not rely on exif data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+#### 1.0.1 (2018-02-14)
+
+##### New Features
+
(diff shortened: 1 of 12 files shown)


sibelius:
android needs this as well?

RCTCamera had this issue as well?

Michaelvilleneuve:
Yes RCTCamera also had this issue. Android works well for us, at least for most devices.
To reproduce the issue : 

1. Take a picture as portrait with RNCamera
2. Upload it to a webserver (base64 or whatever)
3. Create a simple html page and put an `img` tag with the uploaded image
-> The image will appear as landscape even though the EXIF says otherwise

PR #1283, opened 2018-02-23
Width control and rotation fix

Michaelvilleneuve:
Following on this issue #768 I propose the following changes : 

- Controlling returned image size on IOS and Android by providing `width` option in `takePictureAsync`

I think that width option is a great addition to this lib, indeed, apart from reducing picture taking time, it removes the obligation to resize image after capture

- Ability to enable/disable the orientation fix (that takes up to 6 seconds on Android) by providing `fixOrientation` option (android only, not needed on iOS) 

Note that reducing the image size with the new `width` property allows to reduce orientation fixing time exponantially : 

For an image with a size of 4000x3000, providing `width: 1000` reduces picture taking time from 6 seconds to 1.7seconds on a Samsung Galaxy A3.

### DIFF
--- a/android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
+++ b/android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
@@ -62,7 +62,6 @@ public class ResolveTakenPictureAsyncTask extends AsyncTask<Void, Void, Writable
         try {
             if (inputStream != null) {
                 ExifInterface exifInterface = new ExifInterface(inputStream);
-
                 // Get orientation of the image from mImageData via inputStream
                 int orientation = exifInterface.getAttributeInt(ExifInterface.TAG_ORIENTATION,
                         ExifInterface.ORIENTATION_UNDEFINED);
@@ -71,9 +70,8 @@ public class ResolveTakenPictureAsyncTask extends AsyncTask<Void, Void, Writable
                     mBitmap = resizeBitmap(mBitmap, mOptions.getInt("width"));
                 }
 
-                // Rotate the bitmap to the proper orientation if needed and asked
-                if (mOptions.hasKey("fixOrientation") && mOptions.getBoolean("fixOrientation")
-                        && orientation != ExifInterface.ORIENTATION_UNDEFINED) {
+                // Rotate the bitmap to the proper orientation if needed
+                if (mOptions.hasKey("fixOrientation") && mOptions.getBoolean("fixOrientation") && orientation != ExifInterface.ORIENTATION_UNDEFINED) {
                     mBitmap = rotateBitmap(mBitmap, getImageRotation(orientation));
                 }
 
@@ -86,7 +84,6 @@ public class ResolveTakenPictureAsyncTask extends AsyncTask<Void, Void, Writable
                     WritableMap exifData = RNCameraViewHelper.getExifData(exifInterface);
                     response.putMap("exif", exifData);
                 }
-
             }
 
             // Upon rotating, write the image's dimensions to the response
@@ -146,7 +143,7 @@ public class ResolveTakenPictureAsyncTask extends AsyncTask<Void, Void, Writable
         int width = bm.getWidth();
         int height = bm.getHeight();
         float scaleRatio = (float) newWidth / (float) width;
-
+    
         return Bitmap.createScaledBitmap(bm, newWidth, (int) (height * scaleRatio), true);
     }
 
--- a/android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
+++ b/android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
@@ -61,6 +61,9 @@ public class ResolveTakenPictureAsyncTask extends AsyncTask<Void, Void, Writable
 
       try {
           if (inputStream != null) {
+              if (mOptions.hasKey("width") && mOptions.hasKey("height")) {
+                  mBitmap = resizeBitmap(mBitmap, mOptions.getInteger("width"), mOptions.getInteger("height"));
+              }
               ExifInterface exifInterface = new ExifInterface(inputStream);
               // Get orientation of the image from mImageData via inputStream
               int orientation = exifInterface.getAttributeInt(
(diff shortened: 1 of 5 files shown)


android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
```diff
@@ -24,18 +24,18 @@
...
-  private ReadableMap mOptions;
-  private File mCacheDirectory;
-  private Bitmap mBitmap;
-
-  public ResolveTakenPictureAsyncTask(byte[] imageData, Promise promise, ReadableMap options) {
-    mPromise = promise;
-    mOptions = options;
-    mImageData = imageData;
-  }
+    private static final String ERROR_TAG = "E_TAKING_PICTURE_FAILED";
```
Michaelvilleneuve: My editor messed up indentation on android files I guess, sorry about that

jgfidelis:
This looks promising but I have two comments:
1 - You know this by your own comment, but it is difficult for us to see easily what changed with so many indentation changes...

2- Why only `width`? I think it would be better to add a `resize` option, it would be a dictionary with `width` and `height` key (optional). Or event a `scale` key.

If `scale` = 0.5, the image will have `width` and `height` of half of its original size.

So an input could be: `picOptions = { resize: { width: 300, height: 400 } };` or `picOptions = { resize: { scale: 0.8 } }`.

It would be fairly simple additions.

jgcmarins:
Agreed with @jgfidelis

Michaelvilleneuve:
Hey,

I was thinking only `width` because the need was just about reducing image size, and we are often asked to provide images that are like XX pixels large. Providing only width also implied that the ratio would be identical to the original.

A full resize option would require more things such as where to start the resize (top left, center, right, etc).

The width option appeared to be the most obvious and simplest addition to the quality option.

I do not think that a scale option would be relevant, indeed there are thousands of devices available with many different Mpx quantities and, depending on the device, you would get a completely different image size and the result would be inconsistent and hard to work with once sent over to a server.

Regarding the indentation I am sorry about that I noticed it once I already sent you the PR. 
Depending on what you guys choose to implement I could send you a new one without the indent issue. 🙂

jgfidelis:
I suggested the `resize` option because for me, as a programmer that does not know a lot about cameras and pictures, the `width` option looks like something that would only change the width of the picture. You always have to think like the guy who just downloaded this lib and has no idea how to use it haha.

Given the scenario above I will leave it to you the decision: if you can come up with a better option name than `width`. Or implement `height` as well, if you think it is relevant.

The other option is: leave as it is and we merge it. Thanks!

jgfidelis:
Is this the same as #1278? Can we close the other?

sibelius:
I think `width` is good enough, we just need to make clear that this will keep the aspect ratio

the crop part should be in another package not this one, but the resize based on `width` should be here to make takePicture fast enough

Michaelvilleneuve:
@jgfidelis No it not the same problem, but because I submitted #1278 on our master branch, other commits added themselves to the stack when I merged for our needs... I'll close the other one and resubmit it on a clean dedicated branch for you to review it :). 

Regarding this PR, I'll try to fix the indent issue before merging. I will also clarify the fact that `width`keeps the same aspect ratio in the documentation then. 

Thank you !

android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
```diff
@@ -44,99 +44,112 @@ public ResolveTakenPictureAsyncTask(byte[] imageData, Promise promise, ReadableM
...
+        }
+
+        try {
+            if (inputStream != null) {
+                ExifInterface exifInterface = new ExifInterface(inputStream);
+
+                // Get orientation of the image from mImageData via inputStream
+                int orientation = exifInterface.getAttributeInt(ExifInterface.TAG_ORIENTATION,
+                        ExifInterface.ORIENTATION_UNDEFINED);
+
```
Michaelvilleneuve: First change below, adding width key

android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
```diff
@@ -44,99 +44,112 @@ public ResolveTakenPictureAsyncTask(byte[] imageData, Promise promise, ReadableM
...
+                if (mOptions.hasKey("width")) {
+                    mBitmap = resizeBitmap(mBitmap, mOptions.getInt("width"));
+                }
+
```
Michaelvilleneuve: Below, making the orientation fix optional on Android

android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
```diff
@@ -44,99 +44,112 @@ public ResolveTakenPictureAsyncTask(byte[] imageData, Promise promise, ReadableM
...
+        // An exception had to occur, promise has already been rejected. Do not try to resolve it again.
+        return null;
+    }
 
     private Bitmap rotateBitmap(Bitmap source, int angle) {
         Matrix matrix = new Matrix();
         matrix.postRotate(angle);
         return Bitmap.createBitmap(source, 0, 0, source.getWidth(), source.getHeight(), matrix, true);
     }
```
Michaelvilleneuve: Last change below, adding the resize method for when `width` property is given

android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
```diff
@@ -44,99 +44,109 @@ public ResolveTakenPictureAsyncTask(byte[] imageData, Promise promise, ReadableM
...
+
+        try {
+            if (inputStream != null) {
+                ExifInterface exifInterface = new ExifInterface(inputStream);
+                // Get orientation of the image from mImageData via inputStream
+                int orientation = exifInterface.getAttributeInt(ExifInterface.TAG_ORIENTATION,
+                        ExifInterface.ORIENTATION_UNDEFINED);
+
+                if (mOptions.hasKey("width")) {
+                    mBitmap = resizeBitmap(mBitmap, mOptions.getInt("width"));
```
Michaelvilleneuve: First change here, adding with option and resizing if provided

android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
```diff
@@ -44,99 +44,109 @@ public ResolveTakenPictureAsyncTask(byte[] imageData, Promise promise, ReadableM
...
+                }
+
+                // Rotate the bitmap to the proper orientation if needed
+                if (mOptions.hasKey("fixOrientation") && mOptions.getBoolean("fixOrientation") && orientation != ExifInterface.ORIENTATION_UNDEFINED) {
```
Michaelvilleneuve: Making rotation optionnal with the `fixOrientation` boolean

android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
```diff
@@ -44,99 +44,109 @@ public ResolveTakenPictureAsyncTask(byte[] imageData, Promise promise, ReadableM
...
+        return null;
+    }
 
     private Bitmap rotateBitmap(Bitmap source, int angle) {
         Matrix matrix = new Matrix();
         matrix.postRotate(angle);
         return Bitmap.createBitmap(source, 0, 0, source.getWidth(), source.getHeight(), matrix, true);
     }
 
+    private Bitmap resizeBitmap(Bitmap bm, int newWidth) {
```
Michaelvilleneuve: Last change here, adding the `resize`method. Any other change in the file is related to the indent failure

nkov:
FYI, Just tried this fix, and it crashes the app when passing a `float` to `width` instead of an `integer` (such as that which is returned from `Dimensions.get('window').width`). Might be nice to cast it in the module.

nkov:
FYI, Just tried this fix, and it crashes the app when passing a float to width instead of an integer (such as that which is returned from `Dimensions.get('window').width`). Might be nice to cast it in the module.

`

Issue #1263, opened 2018-02-19
Image has weird properties, causing flipped images


benjdlambert:
### Which implementation are you using
*RNCamera*

### Steps to reproduce
1. Take picture to base64

### Expected behaviour
Image is displayed properly.

### Actual behaviour
- So this is an interesting one that I can't really explain.
- The image seems to be rotated 90 degrees when it's used in places.
- I can provide an example here: (I just pasted this link in here: `https://s3-eu-west-1.amazonaws.com/picture.dev.justfootball.io/8b7c50a9-5eed-4874-a619-c96795ebebb2/8212e579-0c62-430b-86dc-42360be8928a`)

Notice the difference when it's loaded in Chrome, when something else goes to fetch it. By something else, I mean github's service that caches images in usercontent. This is not just a github thing, everthing fails to display it properly, even the `<img>` component of react native.

The image also loads horizontally, instead of vertically which is insane, i've never seen that before.

![image](https://user-images.githubusercontent.com/3645856/36379388-a6e27c04-157e-11e8-9f1f-3c039a719c10.png)

### Environment
- **Node.js version**: 8
- **React Native version**: 0.52.0
- **React Native platform + platform version**: iOS 11.2.1 not tested on android
### react-native-camera
**Version**: "master" but happens on ^1.0.1 also

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate

jgfidelis:
Can you please test on Android as well?

johnbodenhamer:
I am having this problem on a Pixel 2

nkov:
We are seeing this issue on Android and iOS, and we have checked the EXIF properties which show a field `Orientation: Rotate 90 CW`

Michaelvilleneuve:
can you try with forceUpOrientation using 
`"react-native-camera": "git+https://git@github.com/Snapp-FidMe/react-native-camera",`
see https://github.com/react-native-community/react-native-camera/pull/1278 for more info

Fconstant:
Fixed by #1290

arapocket:
Is anyone still getting landscape photos even when using forceUpOrientation?

Testing on a simulator it worked properly. Testing on an Iphone X it did not. Has anyone had trouble with any other devices?

aditya-rumbleon:
I still see the issue with landscape photo after using forceUpOrientation. Testing on iPhone 6S

n1mda:
[...]

PR #1290, opened 2018-02-26
 Add ability to force portrait orientation on IOS

Michaelvilleneuve:
iOS uses exif based orientation, this causes orientation failure when displaying image on a web page.
Indeed web browsers don't read exif when displaying images and iOS images appear as landscape.

The forceUpOrientation boolean fixes this by rotating the actual image data and therefore does not rely on exif data.

This PR is a fix for https://github.com/react-native-community/react-native-camera/issues/1263


johnbodenhamer:
What would be a good fix for android devices?

PR #1322, opened 2018-03-06
Fine-grained podspec (iOS)

ognen:
Due to the rather involved installation setup for the use of the face detector on iOS with cocoapods, it is perhaps better that it is turned off by default when using the library via cocoapods.

Users wanting to use the face detector, can opt-in to do so.

To achieve this, this PR introduces subspecs for the 3 major components within the library:
- RN
- RCT
- FaceDetector

When using just 

```
pod 'react-native-camera' path: ...
```

Only RN and RCT are included by default. 

To also include the FaceDetector code, one should use

```
pod 'react-native-camera', subspecs: ['RCT', 'RN', 'FaceDetector'], path: '../node_modules/react-native-camera'
```

Ideally, one should also use the [pod for google vision](https://cocoapods.org/?q=google%20vision) when using cocoapods, but I leave that to someone that needs the feature to improve it ;).

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ cp node_modules/react-native-camera/postinstall_project/projectWithoutFaceDetect
 
 And add something like this to the `scripts` section in your `package.json`:
 
+*Note:* The face detection code is excluded by default for the **CocoaPods** installation.
 ```
 "postinstall": "./scripts/post.sh",
 ```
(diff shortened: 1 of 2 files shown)


sibelius:
`patch-package` could solve for non cocoapods users?

n1ru4l:
@sibelius I mean that I am using patch-package to fix the current podspec file and would love to have this change in the next release, so I do not need to patch it.

sibelius:
@ognen tks for this

how can we improve the workflow for non cocoapods users?

n1ru4l:
I think not using cocoapod is a mess and I am so happy that I ported our project to cocoapods for managing everything.

But maybe we could add a script that can be added as `prepare` which downloads the libraries?

 E.g. if there is a script that downloads boost https://github.com/react-native-community/boost-for-react-native you install react via cocoapods

ognen:
@sibelius You would have to make the Google vision sources available and ready to be compiled within the Xcode project (so that `react-native link` just works). This could either be done via:
- an npm `prepare` script that does the downloading, unpacking and patching for you (as @n1ru4l suggest).
- git submodules

PR #1341, opened 2018-03-12
[iOS] Make GoogleMobileVision framework optional

brunolemos:
Make it optional for real. 
Having to add a postinstall script was not the ideal DX.

_PS1: It would be nice to make it really optional on Android too, removing the need of installing `play-services-vision`._

_PS2: Please test it on a project with face detection before deploying. I only tested on a project without face detection._


jgfidelis:
I will take a look into this during the week, thanks, this could be really useful

jgfidelis:
@brunolemos can you resolve the conflicts please?

brunolemos:
@jgfidelis resolved

jgfidelis:
@brunolemos , can you test if it works on https://github.com/react-native-community/rncamera-example? That is the rncamera demo app and it uses face detection. If it works there, then we can safely merge it :)

PR #1345, opened 2018-03-13
Types: update types for video recording codec

Fconstant:
Update types according to #1339 

Just one thing:
if you have a `const camera: RNCamera;`...
```typescript
camera.recordAsync().then((response) => {
    if (response.codec === 'H264') { // This will result in an error, You can't compare 'symbol' to 'string'
        // CODE
    }
    if (response.codec === RNCamera.Constants.VideoCodec['H264']) { // This won't
        // CODE
    }
})
```
I don't know if it should behave like that. Because I don't know if `response.codec` really comes as a `keyof VideoCodec` (`string`). Because of that, I made this safe comparison which shouldn't give any errors in runtime.


n1ru4l:
@Fconstant `H264` is the key of the value. Therefore you will have to do `response.codec === RNCamera.Constants.VideoCodec['H264']`

Fconstant:
@n1ru4l Oh ok. So I did the right thing. Thanks! :+1:

PR #1355, opened 2018-03-15
Fix crash if convert exposure time to double format is null

6pm:


android/src/main/java/com/lwansbrough/RCTCamera/MutableImage.java
```diff
@@ -218,8 +219,9 @@ public void writeDataToFile(File file, ReadableMap options, int jpegQualityPerce
     // Reformats exposure time value to match ExifInterface format. Example 1/11 -> 0.0909
     // Even the value is formatted as double it is returned as a String because exif.setAttribute requires it.
     private String convertExposureTimeToDoubleFormat(String exposureTime) {
+
         if(!exposureTime.contains("/"))
-          return "";
+            return null;
```
gla5001: 👍 i suggest this but wasnt sure if it was the correct route

6pm: Please check usages of this methods:
`String convertedTime = convertExposureTimeToDoubleFormat(object.toString());
if (tagName.equals(ExifInterface.TAG_EXPOSURE_TIME) && convertedTime != null)`

android/src/main/java/com/lwansbrough/RCTCamera/MutableImage.java
```diff
@@ -46,7 +46,7 @@ public int getHeight() {
         return this.currentRepresentation.getHeight();
     }
 
-    public void mirrorImage() throws ImageMutationFailedException {
+    public void mirrorImage() throws com.lwansbrough.RCTCamera.MutableImage.ImageMutationFailedException {
```
jgfidelis: com.lwansbrough.RCTCamera.MutableImage.?

Issue #1262, opened 2018-02-19
Pluggable Extension System


ma-pe:
Hi all!

@sibelius and I were thinking about a system of extensions for react-native-camera.

It would be nice if users could choose extensions as post-processing (`Google Vision`, `OpenCV`, ...) or any other extension (ideally community written). These extensions would be optional resulting in a minimal setup with minimal footprint for users that just need the camera without extensions.

Is there anyone that could give some insights about how to structure a system like that for iOS/ObjectiveC and Android/Java? Or even better: anyone with experience who'd like to help?

jgfidelis:
I think we could leave the basic `react-native-camera` without any extension as a package. Using it would be as easy as adding the package and linking it.

Then, create another repo for camera with exentions, like Face Detection, that already adds some bit difficult steps to the installation process.

For iOS/ObjC we could make many targets and select the files to compile to each target. The user would then add the correct targets to its app's xcode project, but I am still a junior developer so anyone with more experience may have a better idea.

n1ru4l:
## iOS

On Objective-C we can do something like:

```objc
#if __has_include(<react-native-camera/ExtensionXY.h>)
```

**Default Setup:**
With https://github.com/react-native-community/react-native-camera/issues/1980 we could make enabling the extensions easier.

**CocoaPods Setup:**

Here we can use SubPods to enable specific features (I already did something similar for react-native video https://github.com/react-native-community/react-native-video/blob/b2aa8c4e383047475fe6ad794fdf40f058bdec9b/react-native-video.podspec#L23)

## Android

I currently have no Idea how to conditionally add extensions **Help Wanted**


____

Personally I would also prefer if all features that are not basic should not be enabled/implemented by default

PR #1388, opened 2018-03-22
Integrating Google Vision's text recognition in Android

alculquicondor:
I need OCR for a project I'm currently working on. I decided to add the changes to the library itself, since I didn't find a way to add custom detection mechanisms and I also would like to contribute.

It is already working, but I would like some feedback, would that be code structure or other missing things (a js type for TextBlock maybe?). Once it is settled, I can document too.

PS: I saw the reverted merge regarding OpenCV, which was also implementing text detection. Is this the direction the project is going? (CC @ma-pe).

### DIFF
--- a/src/RNCamera.js
+++ b/src/RNCamera.js
@@ -31,6 +31,22 @@ type TrackedFaceFeature = FaceFeature & {
   faceID?: number,
 };
 
+type TrackedTextFeature = {
+  type: string,
+  bounds: {
+    size: {
+      width: number,
+      height: number,
+    },
+    origin: {
+      x: number,
+      y: number,
+    },
+  },
+  value: string,
+  components: Array<TrackedTextFeature>,
+};
+
 type RecordingOptions = {
   maxDuration?: number,
   maxFileSize?: number,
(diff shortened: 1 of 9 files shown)


ma-pe:
No, OpenCV is not the direction the project is going.

I accidentally merged my fork, that implements text block detection, but not text recognition.

sibelius:
we would like to support both, but in an optional way

we need to investigate this more https://github.com/react-native-community/react-native-camera/issues/1262

sibelius:
@alculquicondor could u take a stab on iOS code?

alculquicondor:
Sadly no. I don't program for iOS and Google Vision's text recognition is Android only.

sibelius:
could u add this to the docs as well?

alculquicondor:
Done. Wasn't there a RNCameraExample project?

src/RNCamera.js
```diff
@@ -58,6 +58,7 @@ type PropsType = ViewPropTypes & {
   autoFocus?: string | boolean | number,
   faceDetectionClassifications?: number,
   onFacesDetected?: ({ faces: Array<TrackedFaceFeature> }) => void,
+  onTextRecognized?: Function,
```
sibelius: can u improve this flowtype?

what is the shape of this function?

alculquicondor: Done

jgfidelis:
Looks nice. Could you take a screenshot of it working?

For iOS there is a built-in solution, looks like... https://www.appcoda.com/vision-framework-introduction/

Issue #1246, opened 2018-02-14
Torch is deactivated on record start


ma-pe:
### Which implementation are you using

*RNCamera*

### Steps to reproduce
1. Use camera with torch activated
2. Start video recording

### Expected behaviour
The torch stays activated. 

### Actual behaviour
The torch is switched off, when recording is started.

### Environment
- **React Native version**: 0.53.0
- **React Native platform + platform version**: iOS 11.2.5 (Android works as expected)

### react-native-camera
**Version**: "master"

jgfidelis:
this is the second bug related to ios torch (also #1189).

Can you test if it happens with the expo camera component as well?

use this app: https://expo.io/@community/camerja

ma-pe:
I sure would, but you can't record videos (only photos) with the expo-test-app. :(

jgfidelis:
Sorry haha. I will take a look into it eventually.

imspikey:
I have the same problem.
also I have used  a react-native-flash for turning flashlight manually and whenever camera starts recording flash turns of instantly ???!!
Please can you fix this issue 

Environment
React Native version: 0.53.0
React Native platform + platform version: iOS 11.2.5 (Android works as expected)
"react": "16.2.0",
 "react-native": "^0.53.3",
"react-native-camera": "^1.0.1",

sibelius:
does it make sense to incorporate this https://github.com/KrowdBeat/react-native-flash

in this repo?

edi:
I'm posting about this everywhere ... but it would be great if `react-native-flash` gets added to Expo ... 

There are so many use cases for it. One of them is that we need a toggle for the camera's flash when you want to scan a barcode during the night.

Another one is morse code communication between two devices at night, when there's no connection between a couple of households.

Please advise.

Issue #1249, opened 2018-02-15
Black flashing screen when recording video (also saved in the recording)


n1ru4l:
### Which implementation are you using

*RNCamera*

### Steps to reproduce

1. Mount a video component with autoFocus on
2. Use the reference prop to start recording when pressing the screen

### Expected behaviour
The video start recording without the black going screen for a few milliseconds. Also the video should be saved without black frames at the beginning

### Actual behaviour
The screen turns black for a few milliseconds. The few milliseconds of dark frames are also present in the recording. I think this is related to autoFocus.

### Environment
- **Node.js version**: v8.9.4
- **React Native version**: v0.52.2
- **React Native platform + platform version**: iOS 10

### react-native-camera
**Version**: v1.0.1

n1ru4l:
Duplicate of https://github.com/expo/expo/issues/1057

n1ru4l:
@jgfidelis @sibelius Would you accept a PR on `RNCamera` that implements a property for the current mode (`recording` and `picture`), in order to prevent the flashing when recording?

Also there must be a solution to this problem, bot snapchat and instagram can handle this, so it should not be impossible. We might get some inspiration (or even reuse?) some open source solution which is written in objective-c?

sibelius:
We will accept a PR to fix this

n1ru4l:
@delta98 Any ideas what snapchat/instagram is doing differently? Because from what I can see they are also switching between the modes without flashings

delta98:
@n1ru4l it looks like it happens after calling `[self setupMovieFileCapture];`

delta98:
Actually nailed this down even further, when `record` is called, the `session` configuration is changed, this causes the current session to be unlocked, which causes the flicker. 

`updateSessionPreset` and `updateSessionAudioIsMuted` use this within their function calls. 

Removing these calls from record and moving `setupMovieFileCapture` to `initializeCaptureSessionInput` or `initWithBridge` allows capture to active more _seamless_. 

I'll put a PR together which takes into account the above and provides the same functionality, but likely will need to limit the camera mode to an instance of camera.

delta98:
@n1ru4l could you checkout this branch and confirm whether this stops the flicker you see https://github.com/react-native-community/react-native-camera/pull/1542

Will also need to soften some of the changes and re-introduce functionality for face detection.

delta98:
@sibelius ^^

n1ru4l:
[...]

PR #1395, opened 2018-03-23
fix: issue #1246 - torch will be disabled when starting the record

ma-pe:
As the preview (https://github.com/react-native-community/react-native-camera/issues/1249), also the torch will shortly be turned off, when starting the record.

Though with this fix/workaround the torch is reactivated and is present while recording.

Fixes https://github.com/react-native-community/react-native-camera/issues/1246


jgfidelis:
nice

PR #1397, opened 2018-03-25
Fix underexposed beginning of video recording when AVCaptureMovieFileOutput is added

Ignigena:
In our application we need to support both Still capture and Movie capture modes. While still mode performs as expected, whenever the user is capturing a Movie, the first few frames of the output are badly underexposed before returning to the correct exposure as seen on the preview.

In fact, you can see the effect on-screen: immediately after calling `this.camera.recordAsync()` the preview layer dips down to almost black and then back to normal. Setting a breakpoint, I was able to confirm that the camera hardware itself is performing an exposure change as `self.videoCaptureDeviceInput.device.adjustingExposure` is `TRUE`.

This appears to happen whenever a new `AVCaptureOutput` is added to the currently running `AVSession`. In this case, because the `AVCaptureMovieFileOutput` is added immediately before `startRecordingToOutputFile` the recording starts before the exposure has been fully adjusted.

In addition, because the output is removed after every capture, even subsequent video recordings without switching to Still mode all exhibit the same issue where the beginning is badly underexposed.

From what I can tell of the code, this is being done to support the Google Vision facial detection APIs since `AVCaptureMovieFileOutput` and `AVCaptureVideoDataOutput` cannot run off the same `AVSession`. In our case, we are not using these APIs and it's more important to have a consistent exposure across the entire video recording.

This fix only changes behavior for installations without the Google Vision API. Under these installations, the `AVCaptureMovieFileOutput` will be instantiated alongside the `AVCaptureStillImageOutput`. This allows for both Still and Movie capture without the camera re-exposing in between. For installations that include the Google Vision API the previous behavior is left in place where the Movie output is added or removed as needed for the recording. I don't have the ability to properly test a solution with a full Google Vision installation.

I've attached a screenshot of a video recorded from an iPad Pro 10.5" tablet which consistently produces the underexposed beginning. With this fix in place, both Still and Video capture work without causing the exposure to change dramatically, even when switching between the two in the same user session.

screen shot 2018-03-24 at 6 46 57 pm

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -59,7 +59,7 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
         //                                                 selector:@selector(bridgeDidBackground:)
         //                                                     name:EX_UNVERSIONED(@"EXKernelBridgeDidBackgroundNotification")
         //                                                   object:self.bridge];
-        
+
     }
     return self;
 }
@@ -128,7 +128,7 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
 {
     AVCaptureDevice *device = [self.videoCaptureDeviceInput device];
     NSError *error = nil;
-    
+
     if (self.flashMode == RNCameraFlashModeTorch) {
         if (![device hasTorch])
             return;
@@ -176,7 +176,7 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
             }
         }
     }
-    
+
     [device unlockForConfiguration];
 }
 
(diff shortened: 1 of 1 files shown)


Ignigena:
I don’t have an app in production that uses the Google Vision API which is why I left that part untouched. Didn’t want to introduce any new side effects I wasn’t aware of for those that are using it.

delta98:
Sorry @Ignigena, I don't believe this resolves the issue, still observing this behaviour.

Ignigena:
@delta98 are you using Google Vision in your app? This PR won’t fix the issue on projects with Vision running unfortunately.

delta98:
@Ignigena I'm not using Google vision, here is my project https://github.com/delta98/laff-camera if you are able to help understand what's happening it would be greatly appreciated.

thomasobrien99:
I would second that this issue does not appear to be fixed.

PR #1410, opened 2018-03-28
fix(types): fix types for #1402

Fconstant:
Fix length problem pointed at #1402 comment. Also leaves a cleaner type reading


Fconstant:
Also, there are missing docs for this feature...

Issue #1404, opened 2018-03-27
Android bar code read not workig and camera preview streched


ShahoG:
### Which implementation are you using
*RNCamera*

### Steps to reproduce
1. Open Camera
2. Scan bar code

### Expected behaviour
Bar code should be read and data sent to function and camera should have a correct preview image

### Actual behaviour
Nothing happens on bar code scan and camera preview image is streched

### Environment
- **Node.js version**: 8.9.3
- **React Native version**: 0.52.0
- **React Native platform + platform version**: iOS 10.3.1, Android 7.0

### react-native-camera
**Version**: master

I'm testing the same code on iOS and it's working fine, but not on Android (Samsung galaxy edge S7 or any other device). I have it set to type EAN-39 code in the example, but EAN-13 isn't working either. Although QR is working

My code in the render:
```
<View style={{ flex: 1 }}>
   <RNCamera
      ref={ref => { this.camera = ref }}
      onBarCodeRead={data => this._scanOrder(data)}
      style={styles.cameraView}
      barCodeTypes={[RNCamera.Constants.BarCodeType.code39]}
     />
</View>

ShahoG:
Also the QR read is extremely slow on android as in iOS it is extremely fast

Swappdevelopment:
Having the same issue here. Working perfectly on iOS but on Android it's not. 

React Native version: 0.54.2

ShahoG:
Already tried that, getting the same result

ShahoG:
Already tried that, getting the same result

brunoksato:
I have the same problem. I'm starting debugging android now if I find the problem I send a PR.

Swappdevelopment:
I've been able to solve my issue. The problem I had was that my <RNCamera> component was in an absolute view having a width and height of 100%. I've changed the 100% width and 100% height to the device width and height! Try this by just adding a fix width and height!

brunolemos:
Stretch issue is duplicated of this: #1243
Please focus on the other issue here

brunolemos:
Stretch issue is duplicated of this: #1343
Please focus on the other issue here

jgfidelis:
Please try on master branch

ShahoG:
Yes can confirm that this is now working, thanks a lot! However the barcode scanning is a lot faster and more accurate on iOS.

ShahoG:
I can also add that QR code scanning is only slow on larger QR codes. Ones containing a small amount of info are relatively fast on android too.

Issue #1272, opened 2018-02-21
EAN-13 barcode doesn't scan vertically


karpie28:
### Steps to reproduce
1. Try to scan barcode EAN13 format
2. Present barcode vertically (does not scan)
3. Present barcode horizontally (Scans quickly)

### Expected behaviour
Should scan both horizontally and vertically

### Actual behaviour
Present barcode horizontally (does not scan)
Present barcode vertically (Scans quickly)

When book is vertically:

![untitled](https://user-images.githubusercontent.com/16509861/36494250-d36f87da-1731-11e8-9c8c-9fe3132f7612.png)

When book is horizontally:

![untitled2](https://user-images.githubusercontent.com/16509861/36494252-d4f5662e-1731-11e8-8961-8807058ee6d7.jpg)

### Environment
- **Node.js version**: 8.4.0
- **React Native version**: 0.50.4
- **React Native platform + platform version**:  Android 6.0.1

### react-native-camera
**Version**: master

CodeAddictSebastianGolinski:
Same problem here 😕

OzoTek:
Same, since I upgraded from RCTCamera to RNCamera last version, was working fine before

K-Leon:
Probably related to aspect ratio like discussed in #1258

JorgeCeja:
What aspect ratio are you using? @karpie28

csumrell:
I am getting this same error on Android only, and only when my phone is positioned in Vertical/Portrait mode. It just stares at the barcode without doing anything, however if i stop the camera, turn the phone sideways (Landscape) and scan it, it processes it very quickly.

movie4:
Get seems issues on RNCamera 1.0.1 - EAN13 — barcode

SimonAmpleman:
Also have the same problem. Application is in portrait mode. I've tried every aspect ratio possible. If I turn the phone in landscape, it works fine.

Also, when I save a picture, it appears in landscape mode (it's rotated 90 degrees).

Thank you for your help!

SimonAmpleman:
I'll contribute a 50$ USD if this can be fixed : Camera in portrait mode. Taking picture in portrait mode. And detecting UPC in both portrait or orientation at the same time.

Thanks !

ChangeSomeCode:
same problem

joselazarosiqueira:
Same problem!

jslok:
+1

uikolas:
Same on android

[...]

PR #1422, opened 2018-03-30
Fix for missing barcodescan on Android

SimonErm:
Fixes scan problem of #1404 and  #1272 by rotating the received data and switching height and width of onFramePreview because the given data is always landscape. Tested with useCamera2Api= true and false.

### DIFF
--- a/android/src/main/java/org/reactnative/camera/RNCameraView.java
+++ b/android/src/main/java/org/reactnative/camera/RNCameraView.java
@@ -116,26 +116,43 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
         }
       }
 
+      private byte[] rotateImage(byte[]imageData,int height,int width){
+        byte[] rotated = new byte[imageData.length];
+        for (int y = 0; y < width; y++) {
+          for (int x = 0; x < height; x++) {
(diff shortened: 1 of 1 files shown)


jgfidelis:
Does this affect performance?

SimonErm:
I can't determine any performance problems on OnePlus3t and it's the same rotation function used in RTCCamera, so i think it's ok, but we could use a rotation matrix with the bitmap instead.

jgfidelis:
I asked because a lot of people already complain about performance on Android, so I was a bit of afraid of a rotation after every frame... If you have an optimization, it would be great :D. But great work, regardless

marszall87:
Actually it does impact performance significantly. The UI threads drops to 20-25fps on my Nexus 5X with this rotation is enabled, after removing this function it's smooth 60fps.

Is it possible to perform the rotation only when looking for specific barcodes? Eg. it's totally unnecessary for QR codes, not sure about text and face recognition, although those functions get the rotation parameter anyway.

ducpt2:
it not working in my case. Can not using barcode scan. Even i use https://github.com/react-native-community/rncamera-example to test.
Any help pls?

PR #1428, opened 2018-04-03
Add skip processing on android for fast picture saving

ospfranco:
I'm trying to solve the slow picture taking problem on android, the problem is the decoding/enconding everytime a image is taken, I solved it by skipping processing all together and directly saving the pic to the file system, this makes it really fast.

However one problem remains, orientation, I've been trying to get around it in several ways, but if you want to keep it fast, the only way seems to be by setting the view orientation before taking the photo, this causes two consequences:

1) we need to detect orientation before the photo is taken (I need help here, not an android developer)*
2) if you rotate the preview View, then x and y axis are inverted and it is really weird to use

* one other problem I had was that our app is locked to portrait mode, so I read somewhere we need to use accelerator information and not the usual config change workflow.

I need some help getting the last stretch done. @sibelius @jgfidelis

### DIFF
--- a/android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
+++ b/android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
@@ -53,6 +53,31 @@ public class ResolveTakenPictureAsyncTask extends AsyncTask<Void, Void, Writable
         WritableMap response = Arguments.createMap();
         ByteArrayInputStream inputStream = null;
 
+        if (mOptions.hasKey("skipProcessing")) {
+            try {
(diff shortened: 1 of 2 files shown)


sibelius:
How fast is this?

ospfranco:
Decoding, processing and re-coding takes ~3 seconds on a moto g5 running android 7, just saving the image goes down to ~20-60ms in my tests

ospfranco:
rotating the image to correct device orientation, since I'm saving the byte array straight to memory, decoding/encoding is not possible, so this needs to be done on the preview (before the takePicture call)

ospfranco:
merged into master, the rotation problem can be fixed later.

PR #1441, opened 2018-04-06
feat(play-sound): play sound on capture (android)

jgfidelis:
simple thing. iOS has this by default (but cannot be turned off).

This is android only.

If true, will play shutter sound when taking picture.


wilav-dev (commented):
Ready for merge

PR #1442, opened 2018-04-06
Add iOS support for start/stop preview in RNCamera

### LINKED ISSUE #1438: RNCamera request: still/freeze frame during takePhotoAsync
I'd like to be able to capture a photo and have the experience be similar to the built in camera. Currently, the video preview continues beyond pressing the shutter & then a preview will show a prior frame from the video. This feels very odd when using the camera vs. the native experience.

### Steps to reproduce

1. With a mounted `RNCamera` component, call `takePhotoAsync`
2. Replace component w/ `<Image source={ ... } />` with the `uri` from the capture

### Does it work with Expo Camera?

It probably works the same way. I've opened a duplicate issue there, https://github.com/expo/expo/issues/1613.

### Expected behaviour

The (video) preview should be able to be frozen during the capture of the still image to mimic what happens when using the built in camera. (At least on iOS, that's how it works. I presume Android is the same, but don't have a device.)

### Actual behaviour

The component continues to show the video preview beyond what will be the actual captured image leading to an inconsistent user experience vs. the built in camera.

### Environment
- **Node.js version**: 9.5.0
- **React Native version**: 0.54.2
- **React Native platform + platform version**: iOS 11.2.5

### react-native-camera
**Version**: 1.0.2

wbyoung:
This is based on the work in #859. It adds only the iOS version of these because the Android API is now based on the preview of `android.cameraview.CameraView` ([fork](https://github.com/react-native-community/cameraview)) rather than the prior custom implementation. Hopefully someone else can work through adding an Android version in the future.

It also adds a flag to the `takePictureAsync` to stop the preview immediately once the still capture completes (and before any additional async activity occurs).

Test environment:

iOS 11.2.5
React Native 0.54.2
Node 9.5.0

Resolves #1438.

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -310,8 +310,13 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
 {
     AVCaptureConnection *connection = [self.stillImageOutput connectionWithMediaType:AVMediaTypeVideo];
     [connection setVideoOrientation:[RNCameraUtils videoOrientationForDeviceOrientation:[[UIDevice currentDevice] orientation]]];
+    
     [self.stillImageOutput captureStillImageAsynchronouslyFromConnection:connection completionHandler: ^(CMSampleBufferRef imageSampleBuffer, NSError *error) {
         if (imageSampleBuffer && !error) {
+            if ([options[@"autoStopPreview"] boolValue]) {
+                [self stopPreview];
+            }
+
             NSData *imageData = [AVCaptureStillImageOutput jpegStillImageNSDataRepresentation:imageSampleBuffer];
 
             UIImage *takenImage = [UIImage imageWithData:imageData];
@@ -433,6 +438,30 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
         [self.movieFileOutput startRecordingToOutputFileURL:outputURL recordingDelegate:self];
         self.videoRecordedResolve = resolve;
         self.videoRecordedReject = reject;
+    });
+}
+
+- (void)stopPreview
+{
+#if TARGET_IPHONE_SIMULATOR
+    return;
+#endif
+    dispatch_async(self.sessionQueue, ^{
+        if ([self.session isRunning]) {
+            [self.session stopRunning];
+        }
+    });
+}
+
+- (void)startPreview
+{
+#if TARGET_IPHONE_SIMULATOR
+    return;
+#endif
+    dispatch_async(self.sessionQueue, ^{
+        if (![self.session isRunning]) {
+            [self.session startRunning];
+        }
     });
 }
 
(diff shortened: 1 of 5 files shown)


ios/RN/RNCamera.m
```diff
@@ -310,8 +310,13 @@ - (void)takePicture:(NSDictionary *)options resolve:(RCTPromiseResolveBlock)reso
 {
     AVCaptureConnection *connection = [self.stillImageOutput connectionWithMediaType:AVMediaTypeVideo];
     [connection setVideoOrientation:[RNCameraUtils videoOrientationForDeviceOrientation:[[UIDevice currentDevice] orientation]]];
+    
     [self.stillImageOutput captureStillImageAsynchronouslyFromConnection:connection completionHandler: ^(CMSampleBufferRef imageSampleBuffer, NSError *error) {
         if (imageSampleBuffer && !error) {
+            if ([options[@"autoStopPreview"] boolValue]) {
+                [self stopPreview];
```
jgfidelis: If you stop preview here, when will you call startPreview again? Shouldn't you call just before resolving the promise?

wbyoung: @jgfidelis I thought that could lead to flickering as the resolution of the promise is likely not synchronous with the next JS code to be executed. So it didn’t really solve my use case. Honestly, @aailces has some better ideas in https://github.com/expo/expo/issues/1613#comment-379173558. I think it’d be better to see what she comes up with there and then merge it in here when it lands in `expo`.

wbyoung:
As for a prop and a default — a prop makes sense, but as the option only applies to the taking of a picture, this does better associate it with where it’s used. Honestly, it probably doesn’t matter. This component can’t be purely declarative by nature, so some things are bound to be different from “the React way”.

And I agree on the default, but would consider that a breaking change & didn’t want to introduce that and wait on a v2.0.

jgfidelis:
If you want to merge it, feel free, or wait for @aalices work before

jtweaver:
Any update on this? Would be nice to have.

JanErikFoss:
I'd love this functionality

jgcmarins:
we should resolve conflicts
@wbyoung could you help us on this?

sibelius:
We have another implementation of this on master, check if works fine

sibelius:
@wbyoung we used @aalices implementation of start/stop preview, can you check if https://github.com/react-native-community/react-native-camera/pull/1600 solves your use case?

jtweaver:
@sibelius Is there any documentation on the new implementation? I'm not finding out how to use it. this.camera.pausePreview() is throwing an error (not a function).

sibelius:
you need to use master, check code for now

we gonna accept a pr to improve docs

jtweaver:
I am using master. I don' see any changes to RNCamera.js that would enable this feature. Has anyone actually made stop preview or pause preview work?

jtweaver:
I added the functions in RNCamera.js and created PR #1622

MrPluto:
Synchronize doc and npm module version? seems some features are not be like what the doc say

jtweaver:
How does that typically go? Should the docs reflect what is in the branch (master), or should they reflect the current npm module?

PR #1444, opened 2018-04-07
Add state flow type for RNCamera component

almost:
Makes it work with Flow 0.63.0


jgfidelis:
@sibelius you are the flow guy here

sibelius:
tks

I prefer 

type Props
type State 

instead

PR #1457, opened 2018-04-10
Add GoogleVision barcode scanner

avshenuk:
This adds an implementation of the barcode scanner for Android based on GoogleVision API.
The one actually used by the famous library https://github.com/ekreutz/react-native-barcode-scanner-google

The implementation uses separate properties `googleVisionBarcodeType` and `onGoogleVisionBarcodesDetected` but if anybody could introduce refactoring changes (having some sort of `barcodeDetectionType`) it would be greatly appreciated. Currently, 2 barcode scanning mechanisms use different property types and that's why the decision was made to leave it like this for now.

Tested only on few devices and barcode detection worked almost immediately (as expected) but additional testing will be performed later. If anybody could test this out on their devices - it's much appreciated as well!

Additionally, fixes some small bugs in FaceDetector.

### DIFF
--- a/android/src/main/java/org/reactnative/barcodedetector/BarcodeFormatUtils.java
+++ b/android/src/main/java/org/reactnative/barcodedetector/BarcodeFormatUtils.java
@@ -1,0 +1,58 @@
+package org.reactnative.barcodedetector;
+
+import android.util.SparseArray;
+import com.google.android.gms.vision.barcode.Barcode;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class BarcodeFormatUtils {
+
+  public static final SparseArray<String> FORMATS;
+  public static final Map<String, Integer> REVERSE_FORMATS;
+
+  private static final String UNKNOWN_FORMAT_STRING = "UNKNOWN_FORMAT";
+  private static final int UNKNOWN_FORMAT_INT = -1;
+
+  static {
+    // Initialize integer to string map
+    SparseArray<String> map = new SparseArray<>();
+    map.put(Barcode.CODE_128, "CODE_128");
+    map.put(Barcode.CODE_39, "CODE_39");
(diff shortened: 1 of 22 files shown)


maxmijn:
How would I be able to use the vision api instead of the one used now? Can I set the properties in some way?

ducpt2:
Can you make it work on ios too? Thanks.

ShahoG:
@maxmijn You need to set these two properites:
```
onGoogleVisionBarcodesDetected={()=> SCAN_FUNCTION()}
googleVisionBarcodeType={RNCamera.Constants.GoogleVisionBarcodeDetection.BarcodeType.QR_CODE}
```

To have additional barcode types just add them to googleVisionBarcodeType with a pipe character ( | )

This should be in the documentation btw

sibelius:
@ShahoG feel free to improve docs

avshenuk:
Hm. But will this work? I mean RNCamera.Constants.GoogleVisionBarcodeDetection.BarcodeType.QR_CODE is a string constant that will be converted to an int. How could the “|” be applied to strings?
I’d say this should be implemented inside the Java code. It should get an array of QR code types from JS and “bitwise-or” them when they’re integers.

Correct me if I am wrong...

hardcodet:
@avshenuk @ShahoG 
Piping the constants indeed doesn't work. So for my `code 128` barcode, the commented out setup breaks the scanner:

        const barcodes = RNCamera.Constants.GoogleVisionBarcodeDetection.BarcodeType;

        //doesn't work
        //const appliedBc= barcodes.QR_CODE | barcodes.CODE_128 | barcodes.EAN_13 | barcodes.EAN_8;

        //works, but only for code 128
        const appliedBc= barcodes.CODE_128;

ekreutz:
Glad to see my work being implemented in this community effort! Hopefully we can get rid of ZXing fully one day... 😉
Keep it up!

PR #1462, opened 2018-04-13
Fix camera issues.

avshenuk:
Fixes a critical bug when the app is paused and resumed.

The camera view is removed from the view hierarchy but still keeps receiving lifecycle events like onResume etc. On the other hand, the view manager creates a new camera view when the app is resumed (Not sure why this works like that because the app is only paused and not closed completely). 
This way multiple views are trying to acquire the camera which results in a mount error + new camera view has no access to camera + views number start to grow.

Not sure if removing the old view as a lifecycle event listener solves the problem of propagating the views. But I suppose since we remove the last reference to it - it must be garbage collected. 

I am new in Android development so what do you think, guys?


jgfidelis:
Removing the lifecycle will solve a memory leak the camera was probably having as well

PR #1464, opened 2018-04-13
Only initialize barcode, face detector and text detectors when needed

almost:
Currently everything is initialized even when not needed. That crashes for me, I think maybe the text detector needs a later min sdk version. Either way it seems silly to have to load all this stuff when all you want is a camera. So I've made it all initialize on demand.

### DIFF
--- a/android/src/main/java/org/reactnative/camera/RNCameraView.java
+++ b/android/src/main/java/org/reactnative/camera/RNCameraView.java
@@ -53,9 +53,9 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
 
   // Scanning-related properties
   private final MultiFormatReader mMultiFormatReader = new MultiFormatReader();
-  private final RNFaceDetector mFaceDetector;
-  private final RNBarcodeDetector mBarcodeDetector;
-  private final TextRecognizer mTextRecognizer;
+  private RNFaceDetector mFaceDetector;
+  private RNBarcodeDetector mBarcodeDetector;
+  private TextRecognizer mTextRecognizer;
   private boolean mShouldDetectFaces = false;
   private boolean mShouldDetectBarcodes = false;
   private boolean mShouldScanBarCodes = false;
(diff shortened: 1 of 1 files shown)


android/src/main/java/org/reactnative/camera/RNCameraView.java
```diff
@@ -261,6 +256,9 @@ private void initBarcodeReader() {
   }
 
   public void setShouldScanBarCodes(boolean shouldScanBarCodes) {
+    if (shouldScanBarCodes && !mBarcodeDetector) {
```
avshenuk: I think this is Java, not JavaScript. How does this compile? 
I mean “!mBarcodeDetector” part. It can’t be converted to boolean automatically, can it?

jgfidelis: Yup, should be:
`if (shouldScanBarCodes && mBarcodeDetector == null) {`

jgfidelis:
Other than that, looks good and should be really helpful

android/src/main/java/org/reactnative/camera/RNCameraView.java
```diff
@@ -311,6 +310,9 @@ public void setFaceDetectionMode(int mode) {
   }
 
   public void setShouldDetectFaces(boolean shouldDetectFaces) {
+    if (shouldDetectFaces && !mFaceDetector) {
```
jgfidelis: Here too, `mFaceDetector == null`

android/src/main/java/org/reactnative/camera/RNCameraView.java
```diff
@@ -382,6 +392,9 @@ public void onBarcodeDetectingTaskCompleted() {
   }
 
   public void setShouldRecognizeText(boolean shouldRecognizeText) {
+      if (shouldRecognizeText && !mTextRecognizer) {
```
jgfidelis: and here `Here too, mTextRecognizer == null`

almost:
Sorry @jgfidelis, I pushed the wrong thing. The one I have on my machine does actually compile :) Will make a new PR now...

almost:
Ok, should be fixed now!

PR #1468, opened 2018-04-14
fix(Android): image stretched instead of cropped

### LINKED ISSUE #1343: [RNCamera] Image is stretched instead of cropped on Android

RNCamera

### Steps to reproduce
1. Create a **square** RNCamera component

### Expected behaviour
The square should be filled by cropping the camera (it does this on iOS), not by distorting/stretching the image.

Maybe it should also accept an `aspect` prop like the old camera implementation does (with `Camera.constants.Aspect.fill` as default).

### Actual behaviour
Image is distorted, the same way it would be if we used `Camera.constants.Aspect.stretch` on RCTCamera.

### Environment
- **Node.js version**: 8.9.0
- **React Native version**: 0.54.0
- **React Native platform + platform version**: Android 7

### react-native-camera
**Version**: 1.0.2

waitopiggu:
To resolve #1343
- Used the logic from RCTCameraView, but that only worked correctly in Landscape mode.
- Added an orientation check and more logic for Portrait mode.

### DIFF
--- a/android/src/main/java/org/reactnative/camera/RNCameraView.java
+++ b/android/src/main/java/org/reactnative/camera/RNCameraView.java
@@ -161,15 +161,26 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
     float width = right - left;
     float height = bottom - top;
     float ratio = getAspectRatio().toFloat();
+    int orientation = getResources().getConfiguration().orientation;
     int correctHeight;
     int correctWidth;
     this.setBackgroundColor(Color.BLACK);
-    if (height / width > ratio) {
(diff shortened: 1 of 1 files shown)


android/src/main/java/org/reactnative/camera/RNCameraView.java
```diff
@@ -161,15 +161,26 @@ protected void onLayout(boolean changed, int left, int top, int right, int botto
     float width = right - left;
     float height = bottom - top;
     float ratio = getAspectRatio().toFloat();
+    int orientation = getResources().getConfiguration().orientation;
```
sibelius: nice

PR #1476, opened 2018-04-15
chore(package): enforce no errors on lint and update packages

jgfidelis:


src/Camera.js
```diff
@@ -135,31 +135,12 @@ export default class Camera extends Component {
...
-      >
-        <Text
-          style={{
-            textAlign: 'center',
-            fontSize: 16,
-          }}
-        >
-          Camera not authorized
-        </Text>
+      <View style={styles.authorizationContainer}>
```
sibelius: can we share these permissions view components?

PR #1519, opened 2018-04-26
fix(cache): store video recordings in same directory as photos

Ignigena:
Currently, videos and photos are stored in two slightly different directories on the iOS file system.

It doesn't seem like this is intentional as the same `RNFileSystem generatePathInDirectory` function is used in both circumstances but with `stringByAppendingPathComponent` in one place and `stringByAppendingString` in another. This results to files being stored in two locations inside of the device app container:

Photos: `~/AppData/Library/Caches/Camera`
Videos: `~/AppData/Library/CachesCamera`

Super minor tweak to help my sanity when dumping the app container to debug issues :)


jgfidelis:
Nice catch, sorry about that :P

PR #1538, opened 2018-05-03
The correct capture preset for photos is AVCaptureSessionPresetPhoto

almost:
AVCaptureSessionPresetHigh sometimes means the camera is too zoomed in


almost:
I don't know other's use cases but I can't think why you'd want a different preset when capturing a still. Unless you're also recording video at the same time and want it to match maybe.

PR #1547, opened 2018-05-07
fix status for non authorized/pending camera

n1ru4l:


n1ru4l:
@sibelius Done. Please feel free to merge after CI passes 😊

src/RNCamera.js
```diff
@@ -318,18 +325,21 @@ export default class Camera extends React.Component<PropsType, StateType> {
 
   getStatus = (): Status => {
     const { isAuthorized, isAuthorizationChecked } = this.state;
-    return isAuthorized ? 'READY' : isAuthorizationChecked ? 'PENDING_AUTHORIZATION' : 'NOT_AUTHORIZED';
-  }
+    if (isAuthorizationChecked) {
+      return CameraStatus.PENDING_AUTHORIZATION;
+    }
+    return isAuthorized ? CameraStatus.READY : CameraStatus.NOT_AUTHORIZED;
```
jgfidelis: I am not familiar with this logic but this is not equivalent to what what before.

Before, if `isAuthorized === true` then it would return READY regardless of the `isAuthorizationChecked`.

Is this modification a correction or a "simplification" (I also dislike oneliners with multiple checks)?

src/RNCamera.js
```diff
@@ -102,6 +102,12 @@ type StateType = {
 
 type Status = 'READY' | 'PENDING_AUTHORIZATION' | 'NOT_AUTHORIZED';
 
+const CameraStatus = {
```
sibelius: export this so end user can use it as well

n1ru4l: Is is exported on `RNCamera.Constants`

n1ru4l:
Please not merge yet. I think I made a mistake

jgfidelis:
I was going to comment that naming was not making much sense hehe. Can we merge it?

n1ru4l:
@jgfidelis Feel free to merge 😉

PR #1559, opened 2018-05-15
Fix runtime error missing facebook infer annotation

karenpeng:
facebook infer annotation is still missing in runtime.


sibelius:
@karenpeng this is causing trouble for people

https://github.com/react-native-community/react-native-camera/issues/1565

can u comment on this?

Issue #1486, opened 2018-04-16
java.lang.ArrayIndexOutOfBoundsException when loading camera view


nvme978:
Using RNCamera 
react-native-camera version 1.1.1
"react": "16.0.0",
"react-native": "0.51.0",

Testing on Genymotion and actual device

### Actual behaviour
App crashes with the following logcat on Android
`04-16 06:51:36.372   376  2146 V EmulatedCamera_CallbackNotifier: **** Currently enabled messages:
04-16 06:51:36.372   376  2146 V EmulatedCamera_CallbackNotifier:     CAMERA_MSG_ERROR
04-16 06:51:36.372   376  2146 V EmulatedCamera_CallbackNotifier:     CAMERA_MSG_FOCUS
04-16 06:51:36.372   376  2146 V EmulatedCamera_CallbackNotifier:     CAMERA_MSG_ZOOM
04-16 06:51:36.372   376  2146 V EmulatedCamera_CallbackNotifier:     CAMERA_MSG_PREVIEW_METADATA
04-16 06:51:36.372  2007  2007 E AndroidRuntime: FATAL EXCEPTION: main
04-16 06:51:36.372  2007  2007 E AndroidRuntime: Process: com.aerobittech.okfit, PID: 2007
04-16 06:51:36.372  2007  2007 E AndroidRuntime: java.lang.ArrayIndexOutOfBoundsException: length=115200; index=115679
04-16 06:51:36.372  2007  2007 E AndroidRuntime: 	at org.reactnative.camera.RNCameraView$1.rotateImage(RNCameraView.java:111)
04-16 06:51:36.372  2007  2007 E AndroidRuntime: 	at org.reactnative.camera.RNCameraView$1.onFramePreview(RNCameraView.java:126)
04-16 06:51:36.372  2007  2007 E AndroidRuntime: 	at com.google.android.cameraview.CameraView$CallbackBridge.onFramePreview(CameraView.java:558)
04-16 06:51:36.372  2007  2007 E AndroidRuntime: 	at com.google.android.cameraview.Camera1.onPreviewFrame(Camera1.java:713)
04-16 06:51:36.372  2007  2007 E AndroidRuntime: 	at android.hardware.Camera$EventHandler.handleMessage(Camera.java:1110)
04-16 06:51:36.372  2007  2007 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:102)
[...]

PR #1563, opened 2018-05-17
Fix java.lang.ArrayIndexOutOfBoundsException with image rotation on Android

Wiltzu:
This Fixes the issue https://github.com/react-native-community/react-native-camera/issues/1486. It happens to me every time I open  a camera view for a second time.

Solution was taken from here https://github.com/react-native-community/react-native-camera/issues/1486 and it was suggested by @nijolas.


jgfidelis:
nice

jpudysz:
Thanks! I will test it today

PR #1578, opened 2018-05-22
Bug/save to capture when app goes on pause

6pm:


6pm:
Please rerun tests on circleci. Something wrong with internet connection.

sibelius:
can you rebase and explain why this is needed?

6pm:
@sibelius how to reproduce on android device: take picture and immediately press 'home' button. After that go to app and try to take photo again - camera will be not works. 
It is edge case but I think that must be fixed.

sibelius:
tks, it makes sense

PR #1600, opened 2018-05-31
Feat/preview

jgfidelis:
Implementation based on Expo Camera. Thanks @aalices for her awesome work there.

Also add `fastMode` picture option so you can get the picture just after saving it on disk.
Add `pictureSize` prop

Needs testing.


piercefreeman:
This seems to make it mandatory to add the `pictureSize` prop to RNCamera.  Otherwise it yields a `Error while updating property 'pictureSize' of a view managed by RNCamera` error.  Not sure if this is desired behavior.

jgfidelis:
Should be fixed on #1608

Issue #1613, opened 2018-06-06
RNCamera: Front Camera shows black screen in example app on iOS


n1ru4l:
While ttrying testing (and fixing the feature #1606) with the rn-camera-example application ( https://github.com/react-native-community/rncamera-example). I discovered the following problems:

**Steps to reproduce:**

1. Clone `rn-camera-example` at the current master (https://github.com/react-native-community/rncamera-example/commit/4f87d52fe41be0bb85eb57b7e7f1456e2f7cdc5f)

```bash
git clone https://github.com/react-native-community/rncamera-example.git
cd rncamera-example
# checkout commit in which this occures (master at time of the posting)
git checkout 4f87d52fe41be0bb85eb57b7e7f1456e2f7cdc5f
```

2. Install `react-native-camera` at the version before #1606 got merged

```bash
yarn add react-native-camera@https://github.com/react-native-community/react-native-camera.git#9cddb2a7213697663f6e53ad8c9603cd11c16754
```

3. Open xcode project and run App on device.

4. Use the toggle button to switch between cameras.

**Observation:**

- **The Front Camera only shows a black screen (sometimes the app crashes while switching)**
- **After the RNCamera Component shows a black screen pressing the SNAP button results in  ~~an application crash~~ a thrown exception**

```error
ExceptionsManager.js:73 Exception thrown while executing UI block: *** -[AVCaptureStillImageOutput captureStillImageAsynchronouslyFromConnection:completionHandler:] Inactive/invalid connection passed
```

**Investigation:**

When I install the following version of react-native-camera (4a9322cb8b7d455fc28f7e67a15bff2fd9d7ea3e) aka the commit before the preview feature was implemented:

```bash
yarn add react-native-camera@https://github.com/react-native-community/react-native-camera.git#4a9322cb8b7d455fc28f7e67a15bff2fd9d7ea3e
```

The app crashes upon start (with `EXC_BAD_ACCESS `)

Poor mans console output 😢:
```
2018-06-06 23:40:02.768087+0200 RNCameraExample[1240:505681] Instantiated <GMVFaceDetector: 0x1740917b0>
2018-06-06 23:40:02.798836+0200 RNCameraExample[1240:505681] Instantiated <GMVFaceDetector: 0x1702817c0>
(lldb)
```

After applying the following patch everything seems to be fine:

```diff
diff --git a/App.js b/App.js
index 47b560d..8440591 100644
--- a/App.js
+++ b/App.js
[...]

Issue #1612, opened 2018-06-06
BUILD FAILED - error: collection element of type 'NSInteger'


Grandmac:
### Which implementation are you using

react-native-camera@1.1.4

### Environment
- **React Native version**: 0.55
- **React Native platform + platform version**: iOS 11.3

### Bug under iOS

When running `react-native run-ios`

```
/.../node_modules/react-native-camera/ios/RN/RNCamera.m:327:23: 
warning: implicit conversion loses integer precision: 'long _Nullable' to 'int' [-Wshorten-64-to-32]
        orientation = [options[@"orientation"] integerValue];
                    ~ ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/.../node_modules/react-native-camera/ios/RN/RNCamera.m:444:37: 
warning: comparison between pointer and integer ('NSInteger' (aka 'long') and 'void *')
    if (self.videoStabilizationMode != nil) {
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~
/.../node_modules/react-native-camera/ios/RN/RNCamera.m:502:5: 
warning: code will never be executed [-Wunreachable-code]
    dispatch_async(self.sessionQueue, ^{
    ^~~~~~~~~~~~~~
/.../node_modules/react-native-camera/ios/RN/RNCamera.m:550:5:
warning: code will never be executed [-Wunreachable-code]
    dispatch_async(self.sessionQueue, ^{
    ^~~~~~~~~~~~~~
/.../node_modules/react-native-camera/ios/RN/RNCamera.m:842:11: 
warning: unused variable 'faceDetectorManagerClass' [-Wunused-variable]
    Class faceDetectorManagerClass = NSClassFromString(@"RNFaceDetectorManager");
          ^
/.../node_modules/react-native-camera/ios/RN/RNCamera.m:843:11: 
warning: unused variable 'faceDetectorManagerStubClass' [-Wunused-variable]
    Class faceDetectorManagerStubClass = NSClassFromString(@"RNFaceDetectorManagerStub");
          ^
/.../node_modules/react-native-camera/ios/RN/RNCamera.m:28:17: 
warning: method definition for 'updateFaceDetecting:' not found [-Wincomplete-implementation]
@implementation RNCamera
                ^
```

And finaly

```
/.../node_modules/react-native-camera/ios/RN/RNCameraManager.m:100:22: 
[...]

PR #1606, opened 2018-06-04
WIP: Implement video stabilization mode property for ios

n1ru4l:
This pull request allows setting the property `videoStabilization`for `iOS` devices.

This is required for smoother video recordings.

PR #1614, opened 2018-06-06
Fix failing iOS build #1612

n1ru4l:
This PR fixes the issues introduced in #1606 (and makes the application compilable again).

However `react-native-camera` seems still to be in a broken state after #1600 (Details in #1613).


jgfidelis:
(I told you that you would need a default prop :P)

n1ru4l:
@jgfidelis 👀

PR #1629, opened 2018-06-11
[BUG, ANDROID] width prop not works correctly when fixOrientation is enabled

6pm:
How to reproduce:
1. Add `width={720}` and `fixOrientation={true}`. Aspect ratio '16:9' in my case.
2. Take photo
Result: image width will be 490.
Expected result: image width 720.

React-native version: 0.55.4


bcalik:
@6pm This makes the rotate process slower on Android. Width should be set first, so it will reduce the size of image. So the following operations would work faster on smaller image.

Also we really need to do the jpeg compression operation at first instead doing at the end, so it will make image smaller for further processing.

More info here:
https://stackoverflow.com/questions/10664006/fastest-way-to-rotate-big-size-bitmap

sibelius:
@bcalik can you open an issue for this, and also send a PR?

PR #1641, opened 2018-06-18
Add pauseAfterCapture capture prop

eqyiel:
Adds a property to `takePictureAsync` that pauses the preview as soon as the image is captured.  I realise that `cameraRef.pausePreview()` exists, but I wanted to be able to pause the preview immediately after capture and found it insufficient for that.  You will need to call `cameraRef.resumePreview()` afterwards.

Note that I have no idea if the Camera2 code works.  I couldn't get the Camera2 API to work even before making these changes by changing true to false here: https://github.com/react-native-community/react-native-camera/blob/77e158dc2dedcd1303b2f077bee63ff3be2739f7/android/src/main/java/org/reactnative/camera/RNCameraView.java#L70

This is probably for the same reasons that it was disabled in this commit: https://github.com/react-native-community/react-native-camera/commit/8d9c06ad903b40abc8bef67927d4621c494aeb3b

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -25,6 +25,8 @@ import android.os.Build;
 import android.support.v4.util.SparseArrayCompat;
 import android.view.SurfaceHolder;
 
+import com.facebook.react.bridge.ReadableMap;
+
 import java.io.File;
 import java.io.IOException;
 import java.util.List;
@@ -377,7 +379,7 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
     }
 
     @Override
-    void takePicture() {
+    void takePicture(final ReadableMap options) {
         if (!isCameraOpened()) {
             throw new IllegalStateException(
                     "Camera is not ready. Call start() before takePicture().");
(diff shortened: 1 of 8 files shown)


eqyiel:
Rebased onto master

Fsarmento:
I've been looking for a solution like this since cameraRef.pausePreview() generates an error when takePictureAsync is called. +1 to merge this PR.

n1ru4l:
@sibelius What exactly was the reason for disabling the camera2 API by default and why it is currently not working?

sibelius:
RNCamera is based on expo camera module https://github.com/expo/expo/tree/master/packages/expo-camera

they disabled because it had some bugs

we can enable camera2 if somebody can fix some issues with it

equesteo:
This features is not working for me.  On React Native 0.55.4, react-native-camera 1.4.1

My camera looks like this: 
```
        <View style={{width, height: width}}>
          { this.props.showCam ? <RNCamera
            ref={ref => { this.camera = ref }}
            style = {styles.preview}
            type={this.state.backCamera ? RNCamera.Constants.Type.back : RNCamera.Constants.Type.front}
            flashMode={RNCamera.Constants.FlashMode.off}
          /> : null}
        </View>
```

And taking a picture looks like this:
```
takePicture () {
    if (this.camera) {
      this.camera.takePictureAsync({pauseAfterCapture: true}).then((data) => {
        const cropData = {
          offset: {
            x: 0,
            y: 0,
          },
          size: {
            width: data.width,
            height: data.width,
          },
        };
        ImageEditor.cropImage(
          data.uri,
          cropData,
          (successURI) => {
            ImagePicker.cleanSingle(data.uri).catch(e => logError(e))
            this.camera.resumePreview() // not working?
            this.props.stashNewRidePhoto(successURI)
          },
          (error) => {
            console.log(error.message);
          },
        );

      }).catch((e) => { logError(e) })
    }
  }
```

The viewfinder does not pause, it takes about 4 seconds to cut and crop, and the picture is changing the whole time. Any idea what I might have done wrong?

sibelius:
@equesteo could you open an issue for this case?

equesteo:
Yep. [Here](https://github.com/react-native-community/react-native-camera/issues/1930) for future searchers.

PR #1645, opened 2018-06-18
[WIP] Use AVCaptureVideoDataOutput when recording video

Ignigena:
Currently AVCaptureFileOutput and GMVDataOutput (if you are using Google Vision) are incompatible with each other. In our case, we aren't using Google Vision, but do need to use Apple's Vision framework which has a similar restriction.

By switching to AVCaptureVideoDataOutput both Google and Apple Vision frameworks can run on the same session even during video capture. There's a bit more involved since we're dealing with the actual frame buffer, but there's a lot more flexibility as a tradeoff.

While we are using the contents of this PR in a production app, there are a few things left to do in order to bring feature parity to the current react-native-camera implementation of AVCaptureFileOutput. However, one feature this PR does address that didn't exist previously is the `mirrorImage` option carries over on video not just picture capture.

Before we wrap this up, just wanted to get a temperature on whether this is something the community and this library would benefit from. If so, we'll finish up work on this PR with the goal of having everything backwards compatible with current available options/behavior.

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -12,6 +12,9 @@
 @property (nonatomic, weak) RCTBridge *bridge;
 
 @property (nonatomic, assign, getter=isSessionPaused) BOOL paused;
+
+@property (nonatomic, strong) RCTPromiseResolveBlock videoRecordedResolve;
+@property (nonatomic, strong) RCTPromiseRejectBlock videoRecordedReject;
 @property (nonatomic, strong) id faceDetectorManager;
 
 @property (nonatomic, copy) RCTDirectEventBlock onCameraReady;
@@ -396,9 +399,9 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
 
 - (void)record:(NSDictionary *)options resolve:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRejectBlock)reject
 {
-#if __has_include(<GoogleMobileVision/GoogleMobileVision.h>)
-    [_faceDetectorManager stopFaceDetection];
-#endif
+    if (_isRecording || _videoRecordedResolve != nil || _videoRecordedReject != nil) {
+        return;
+    }
 
     NSError *error = nil;
     NSString *path = [RNFileSystem generatePathInDirectory:[[RNFileSystem cacheDirectoryPath] stringByAppendingPathComponent:@"Camera"] withExtension:@".mov"];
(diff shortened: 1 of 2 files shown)


ios/RN/RNCamera.m
```diff
@@ -464,19 +452,45 @@ - (void)record:(NSDictionary *)options resolve:(RCTPromiseResolveBlock)resolve r
...
+
+            resolve(@{ @"uri": self.videoWriter.outputURL.absoluteString });
+        }];
     });
 }
 
 - (void)stopRecording
 {
-    [self.movieFileOutput stopRecording];
+    self.canAppendBuffer = NO;
```
jgfidelis: As of today, when stopRecording is called, it resolves the `recordAsync` method promise. Right now, it looks like this does not happen here, so people with `const data = await this.camera.recordAsync()` would make it hang forever.

Ignigena: @jgfidelis Correct, this is one of the areas that we need to bring back to parity with the existing behavior. Just didn't want to spend the time getting this perfect if the overall move to VideoDataOutput is not wanted :)

jgfidelis:
This looks good. We don't work with video recording on our apps so I am not the best person to say if this is needed, etc.

What are the pros and cons exactly?

Ignigena:
Pros:
- No need to turn on and off Google Vision tracking on the session between recordings. No switching between different modes that are incompatible with each other.
- Ability to introduce Apple Vision tracking on supported devices (this was the main reason we made the switch for our apps)
- Much more extensible in the future since we have access to the raw frame buffers during both preview and capture. I'm not exactly sure how big of a pro given the whole nature of React Native and cross-platform, but certainly the flexibility is there.

Cons:
- Because the output mode is not specifically tailored to file output, there's a bit more code involved dealing with the raw frames, writing them to the file out and timing on the session.
- Some options, like `maxFileSize` may require a bit more code to keep track of the incoming frame buffer sizes as they are being written to disk since the final output file handle isn't closed until recording is finished.

There may be some others I'm not thinking of right now, but overall once finished the behavior shouldn't change all that much from either the RN bridge/options or the final output file once recording has finished.

sibelius:
`AVCaptureFileOutput` produces better images than `AVCaptureVideoDataOutput`?

Ignigena:
@sibelius To my understanding, image quality is determined by the capture settings (`activeFormat`, `sessionPreset`, etc.) when setting up the recording session, regardless of whether you choose `FileOutput` or `DataOutput`.

So neither should necessarily produce "better" images if compared side by side with the same session configuration. Certainly in our case we haven't noticed any discernible change in quality when using the same output settings, but I'd be interested to know if that's not been your experience?

sibelius:
If the performance is the same, we can follow this approach for sure

Ignigena:
Sounds good. I'll try to get this wrapped up for testing!

reime005:
@Ignigena I'd be interested to know how far you came with that work? I also worked on something that allows me to use face recognition and recording at the same time, similar to your solution.

However, I had some issues regarding audio recording which I hope to have resolved. 

I'm also not using the google face recognition, and in the end had performance issues when doing FR and recording at the same time (couldn't process all the frames in time).

sibelius:
I think this would make things easier on iOS side

sibelius:
@n1ru4l as you are building a video app, this approach could be good for you too

PR #1651, opened 2018-06-19
Continued Recording

flippinjoe:
The purpose of this PR is to allow for continued recording across multiple start/stop by allowing the user to define their output location when capturing


jgfidelis:
nice!

halilb:
Hi @flippinjoe,

Thanks a lot for the PR. This is something we need in our app.

We couldn't get this working on iOS though. Could you check the issue I created and let me know if you can see anything I'm doing wrong?

Thanks again.

PR #1657, opened 2018-06-21
Update Camera1.java

htala:
Check setPictureSize for null to prevent an exception

Fixed a bug which android throw a null reference exception when there is no 'pictureSize' attribute on RNCamera component


jgfidelis:
Isn't is `isEmpty()`?

PR #1661, opened 2018-06-24
Refactor build.gradle

rghorbani:


punksta:
I think it's not common way of changing dependency versions in sub-projects. There are a lot of react-native projects with fixed versions in `build.gradle` files. All of them should use this way.  Also: you can break compiling by changing compile-sdk and target-sdk version automatically. 

if you really needs to change version in sub-projects, better to do it manually. Like this: 
https://docs.gradle.org/current/userguide/customizing_dependency_resolution_behavior.html#sec:dependency_resolve_rules

you can find working example [here(https://github.com/punksta/thecatnative)](https://github.com/punksta/thecatnative/blob/master/android/build.gradle)

rghorbani:
@punksta Actually I just refactored the already existing code. I didn't want to change the way.

punksta:
@rghorbani you are right, sorry. Well, when it were my thoughts about existing gradle file.

jgfidelis:
Is this a 1:1 change, what does this accomplish?

sibelius:
it is

just an improvement on .gradle

PR #1664, opened 2018-06-27
Update Camera.js

huynhtansi:
Checking CameraManager is valid before export


jgfidelis:
rctcamera is deprecated

huynhtansi:
I update for both modules, not only RCTCamera

asuri0n:
nice work ;)

n1ru4l:
This change only affects RCTCamera which is deprecated. Also there is no description why this change should be merged, what issues could be solved?

I am going to close this pull request until a valid explanation is given.

PR #1667, opened 2018-06-28
Moves frame rotation to AsyncTask

SimonErm:
Moves frame rotation to BarCodeScannerAsyncTask like it was in RCTCamera to increase Performance.
#1422

### DIFF
--- a/android/src/main/java/org/reactnative/camera/tasks/BarCodeScannerAsyncTask.java
+++ b/android/src/main/java/org/reactnative/camera/tasks/BarCodeScannerAsyncTask.java
@@ -41,14 +41,27 @@ public class BarCodeScannerAsyncTask extends android.os.AsyncTask<Void, Void, Re
       BinaryBitmap bitmap = generateBitmapFromImageData(mImageData, mWidth, mHeight);
       result = mMultiFormatReader.decodeWithState(bitmap);
     } catch (NotFoundException e) {
-      // No barcode found, result is already null.
+      BinaryBitmap bitmap = generateBitmapFromImageData(rotateImage(mImageData,mWidth, mHeight),mHeight,mWidth);
+      try {
+        result = mMultiFormatReader.decodeWithState(bitmap);
+      } catch (NotFoundException e1) {
+        //no barcode Found
+      }
     } catch (Throwable t) {
       t.printStackTrace();
     }
 
     return result;
   }
-
+  private byte[] rotateImage(byte[]imageData,int width, int height) {
+    byte[] rotated = new byte[imageData.length];
+    for (int y = 0; y < height; y++) {
+      for (int x = 0; x < width; x++) {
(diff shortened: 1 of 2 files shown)


android/src/main/java/org/reactnative/camera/tasks/BarCodeScannerAsyncTask.java
```diff
@@ -41,14 +41,27 @@ protected Result doInBackground(Void... ignored) {
...
+        //no barcode Found
+      }
     } catch (Throwable t) {
       t.printStackTrace();
     }
 
     return result;
   }
-
+  private byte[] rotateImage(byte[]imageData,int width, int height) {
```
jgfidelis: this is not a 1:1 copy from the old method.

Why the changes?

android/src/main/java/org/reactnative/camera/tasks/BarCodeScannerAsyncTask.java
```diff
@@ -41,14 +41,27 @@ protected Result doInBackground(Void... ignored) {
...
+    byte[] rotated = new byte[imageData.length];
+    for (int y = 0; y < height; y++) {
+      for (int x = 0; x < width; x++) {
+        rotated[x * height + height - y - 1] = imageData[x + y * width];
```
jgfidelis: this may lead to crashes, that is why there was an `if` inside the `for` on the old method.

android/src/main/java/org/reactnative/camera/tasks/BarCodeScannerAsyncTask.java
jgfidelis: Before it was:
`rotated[x * width + width - y - 1] = imageData[x + y * height];`

android/src/main/java/org/reactnative/camera/tasks/BarCodeScannerAsyncTask.java
jgfidelis: Please separate into `sourceIx` and `destIx` and check if those are valid, like before

android/src/main/java/org/reactnative/camera/tasks/BarCodeScannerAsyncTask.java
jgfidelis: Ok, looking again, the difference is because before y and x were `height` and `width` and now they are `width` and `height`. If it produces the same result, ok hehe

SimonErm: Yes on last PR i swaped the naming, but now i just copied from RCTCamera and yes it produce the same :-D

android/src/main/java/org/reactnative/camera/tasks/BarCodeScannerAsyncTask.java
SimonErm: Ok

PR #1684, opened 2018-07-06
 Handle return value from permissionsAndroid.request() in react-native[0.56]

s-o-r-o-u-s-h:
According to documentation:
`PermissionsAndroid.request()` returns a promise resolving to a string value('granted', 'denied' or 'never_ask_again' ), not a boolean value.
So there is no need to check device's API level because if it's below 23, It's always `PermissionsAndroid.RESULTS.GRANTED`.

### DIFF
--- a/src/handlePermissions.js
+++ b/src/handlePermissions.js
@@ -19,7 +19,7 @@ export const requestPermissions = async (hasVideoAndAudio, CameraManager, permis
           // On devices before SDK version 23, the permissions are automatically granted if they appear in the manifest,
           // so check and request should always be true.
           // https://github.com/facebook/react-native-website/blob/master/docs/permissionsandroid.md
-          const isAuthorized = granted === PermissionsAndroid.RESULTS.GRANTED;
+          const isAuthorized = granted === PermissionsAndroid.RESULTS.GRANTED || granted === true;
     
           return isAuthorized;
     }


jgcmarins:
hi @s-o-r-o-u-s-h where did you get this information?

s-o-r-o-u-s-h:
@jgcmarins from [here](https://facebook.github.io/react-native/docs/permissionsandroid.html) - third paragraph
I tested both on emulator and real device with API 19. Before my change, it's always say camera is not authorized.

s-o-r-o-u-s-h:
@sibelius If you think it's wrong you can test current code on a device with lower api than 23 and log the `granted` value.If camera permission exists in the manifest, its value is 'granted' not `true`.

s-o-r-o-u-s-h:
to @jgcmarins & @sibelius :
I found out what is the problem. It seems **PermissionsAndroid** component has been changed in the 
 react-native **v0.56**. `PermissionsAndroid.request()` in versions < 0.56 returns a promise resolved to a Boolean value(which is weird because its documentations says its a string). in 0.56 it returns a promise resolved to a string value(granted, denied, never_asked_again). My latest commit handles both of them and fix the permissions problem in 0.56.

jgcmarins:
Great!

PR #1688, opened 2018-07-09
Camera not authorized is always shown for Android Version Below Marshmallow for RNCamera

maulikdhameliya:
[Issue] **Camera not authorized**  is always shown for Android Version Below Marshmallow for RNCamera

[Cause] requestPermissions is always returning false for android version < 23. 

[Fix] Corrected the check condition to return correct value.


maulikdhameliya:
@jgfidelis : Hi Can you check this?

sibelius:
fixed on master

please test

Issue #1255, opened 2018-02-16
java.lang.IllegalArgumentException: Invalid image data size.


sibelius:
```
Caused by java.lang.IllegalArgumentException: Invalid image data size.
       at com.google.android.gms.vision.Frame$Builder.setImageData(Unknown Source)
       at org.reactnative.facedetector.RNFrameFactory.buildFrame(RNFrameFactory.java:16)
       at org.reactnative.camera.tasks.FaceDetectorAsyncTask.doInBackground(FaceDetectorAsyncTask.java:40)
       at org.reactnative.camera.tasks.FaceDetectorAsyncTask.doInBackground(FaceDetectorAsyncTask.java:10)
       at android.os.AsyncTask$2.call(AsyncTask.java:305)
       at java.util.concurrent.FutureTask.run(FutureTask.java:237)
       at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:243)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
       at java.lang.Thread.run(Thread.java:761)
```

wcandillon:
@sibelius We might be experiencing the same issue, did you find a fix/workaround for it?

andreseiji:
```
Caused by: java.lang.IllegalArgumentException: Invalid image data size.
        at com.google.android.gms.vision.Frame$Builder.setImageData(Unkno
        at com.google.firebase.ml.vision.common.FirebaseVisionImage.zza(U
        at com.google.android.gms.internal.firebase_ml.zzpw.zza(Unknown S
        at com.google.firebase.ml.vision.face.FirebaseVisionFaceDetector.Unknown Source)
        at org.reactnative.camera.tasks.ResolveTakenPictureAsyncTask.doInolveTakenPictureAsyncTask.java:109)
        at org.reactnative.camera.tasks.ResolveTakenPictureAsyncTask.doInolveTakenPictureAsyncTask.java:37)
        at android.os.AsyncTask$2.call(AsyncTask.java:295)
        at java.util.concurrent.FutureTask.run(FutureTask.java:237)
```

sibelius:
for the fix follow this https://github.com/expo/expo/pull/2772

MateusAndrade:
[...]

PR #1691, opened 2018-07-11
Fixes at least one cause of #1255

davidpricedev:
This fixes the [1255 issue](https://github.com/react-native-community/react-native-camera/issues/1255) - at least for the instance I was able to reproduce. 

It is an issue that only occurs on some android devices (for us it was a pixel). Occasionally, the data length comes through with significantly less data than it usually has which causes the error in that issue later on. 

On the phones I was able to test with the data.length is always 1.5 * width * height - except when the error occurs (when it was technically 1.5 * width * height / 100).

### DIFF
--- a/android/src/main/java/org/reactnative/camera/RNCameraView.java
+++ b/android/src/main/java/org/reactnative/camera/RNCameraView.java
@@ -118,6 +118,10 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
           return;
         }
 
+        if (data.length < (1.5 * width * height)) {
+            return;
+        }
+
         if (willCallBarCodeTask) {
           barCodeScannerTaskLock = true;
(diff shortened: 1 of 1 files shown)


jgfidelis:
@davidpricedev do you know why the data.length is wrong at that point? This looks more like a workaround. Did you test on iOS?

davidpricedev:
I'm not sure. It is a bit of a workaround. This was reported on a pixel2 - when you toggle in and out of the background a few times at some point the size of the data will be significantly less than it usually is while the width and height remain the same. This fix/workaround ensures that that bad frame gets ignored.

Presumably, there is an underlying bug with the particular device that occasionally leads to this problem while backgrounding or foregrounding. I was unable to reproduce it on other devices with a similar android os version.

PR #1726, opened 2018-08-02
Optional Android permissions and microphone permissions

benjaminreid:
This fixes the issue I mentioned along side this commit that @sibelius and I discussed https://github.com/react-native-community/react-native-camera/commit/ae35e4f2169a01f481ada98a00ceeccc94aa6e56#commitcomment-29939308.

Now permission is being required for the microphone, the same dialog containing the permission title and message for the camera was appearing for the microphone too, creating two duplicate dialogs.

This PR adds new props specifically for the microphone so unique titles and messages an can be added for the camera and microphone separately.

As well as making these dialogs optional. If you were to not specify the `permissionDialogTitle` and `permissionDialogMessage` prop, a blank dialog would appear. This stops the dialog from appearing all together.

A few notes implementation notes,
- prevents the dialog containing `permissionDialogTitle` and `permissionDialogMessage` if _both_ these props are empty
- adds two new props `microphonePermissionDialogTitle` and `microphonePermissionDialogMessage`, these function the same as `permissionDialogTitle` and `permissionDialogMessage` but apply specifically to microphone

### DIFF
--- a/src/handlePermissions.js
+++ b/src/handlePermissions.js
@@ -11,10 +11,12 @@ export const requestPermissions = async (hasVideoAndAudio, CameraManager, permis
             return isAuthorized;
         }
     } else if (Platform.OS === 'android') {
-        const grantedCamera = await PermissionsAndroid.request(PermissionsAndroid.PERMISSIONS.CAMERA, {
-            title: permissionDialogTitle,
-            message: permissionDialogMessage,
-          });
+        const cameraPermissionDialog = permissionDialogTitle || permissionDialogMessage ? {
+          title: permissionDialogTitle,
+          message: permissionDialogMessage,
+        } : undefined;
+
+        const grantedCamera = await PermissionsAndroid.request(PermissionsAndroid.PERMISSIONS.CAMERA, cameraPermissionDialog);
         if (!hasVideoAndAudio) {
             return grantedCamera === PermissionsAndroid.RESULTS.GRANTED || grantedCamera === true;
         }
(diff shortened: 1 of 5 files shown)


src/handlePermissions.js
```diff
@@ -11,17 +11,22 @@ export const requestPermissions = async (hasVideoAndAudio, CameraManager, permis
             return isAuthorized;
         }
     } else if (Platform.OS === 'android') {
-        const grantedCamera = await PermissionsAndroid.request(PermissionsAndroid.PERMISSIONS.CAMERA, {
-            title: permissionDialogTitle,
-            message: permissionDialogMessage,
-          });
+        const cameraPermissionDialog = permissionDialogTitle || permissionDialogMessage ? {
```
sibelius: no need for this, as we have defaultProps

this will always exist

benjaminreid: Great, so if change the default props to...

```js
permissionDialogTitle: null,
permissionDialogMessage: null,
microphonePermissionDialogTitle: null,
microphonePermissionDialogMessage: null,
```

What information do you want expanding in the docs? The docs don’t currently mention the default prop types so I just want to clarify what you think would be good to expand/edit on?

sibelius: if you don't provide a permissionDialogTitle it won't show any message, just the default OS message

benjaminreid: Oh strange... I kept getting blank dialog boxes. I will re-test it and get back to you.

sibelius:
fixed here https://github.com/react-native-community/react-native-camera/pull/1730

tks

PR #1739, opened 2018-08-09
chore(update): update package.json packages

sibelius:
fix a possible vulnerability with lodash


sibelius:
better use node 8 LTS on CI

PR #1757, opened 2018-08-23
Added width to styles.preview

seftrick:
Camera live preview doesn't appears without it


n1ru4l:
@seftrick Could you explain why you would want to set the default value to the window width? The `flex: 1` should be enough to fill a container that fills the whole screen.

I will close this pull request for now.

PR #1776, opened 2018-09-03
fix: android base64 string format for consistency with ios

### LINKED ISSUE #1764: base64 image gets returned with newline characters

RNCamera

### Steps to reproduce
1. Use [the RNCamera example](https://github.com/react-native-community/react-native-camera/blob/master/docs/RNCamera.md) as an example
2. Change the console.log for the photoData returned to:
`console.log(data.base64);`
3. In the debugger, the base64 string that is logged has newline characters in it. This should be one continuous string.

### Expected behaviour
The base64 string should not have new-line characters in it and should be one continuous string.

### Actual behaviour
The base64 string that comes back gets formatted to 76 characters per line and newline characters are inserted. This forces the consumer to find and replace whitespace characters.

### Environment
`"react-native": "0.56.0"`
Android and iOS

### react-native-camera
Master 
`"react-native-camera": "https://github.com/react-native-community/react-native-camera#3c1b47ca9ba98f2481f1a920c4f20e9431a57fa5"`

mvayngrib:
fixes #1764

### DIFF
--- a/android/src/main/java/com/lwansbrough/RCTCamera/MutableImage.java
+++ b/android/src/main/java/com/lwansbrough/RCTCamera/MutableImage.java
@@ -167,7 +167,7 @@ public class MutableImage {
     }
 
     public String toBase64(int jpegQualityPercent) {
-        return Base64.encodeToString(toJpeg(currentRepresentation, jpegQualityPercent), Base64.DEFAULT);
+        return Base64.encodeToString(toJpeg(currentRepresentation, jpegQualityPercent), Base64.NO_WRAP);
     }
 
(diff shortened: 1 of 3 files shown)


mvayngrib:
@sibelius as far as i understand, RNCamera user ResolveTakenPictureAsyncTask.java, which is covered here

sibelius:
is this a breaking change?

mvayngrib:
not unless someone depends on the base64 having line breaks for android but not ios. However, full disclosure: I've only tested taking photos with RNCamera (not videos, or RCTCamera).

sibelius:
it makes sense this behaviour to be consistency

PR #1802, opened 2018-09-19
Update build.gradle

tonmanayo:
googlePlayServicesVersion update - this is making it hard to include other libraries that need an updated google-services version


sibelius:
this can be handled in userland

try change `googlePlayServicesVersion` on `ext`

PR #1808, opened 2018-09-21
Add `onGoogleVisionBarcodesDetected` explanation

kigawas:


kigawas:
@akshay2604 
No. iOS's built in scanner is fast enough. On iOS it will remove this attribute: https://github.com/react-native-community/react-native-camera/blob/master/src/RNCamera.js#L440

akshay2604:
This will use google vision api's internally right? Is that paid?

kigawas:
@akshay2604 You don't have to pay. As long as there is google play service on the device, you can use it.

kigawas:
@akshay2604 

It should work with the default gradle config. If not, use android studio to update gradle, android sdk and play service to newer version (latest is better if possible). FYI:
```
// Top-level build file where you can add configuration options common to all sub-projects/modules.

buildscript {
    dependencies {
        classpath 'com.android.tools.build:gradle:3.2.0'
        classpath 'com.google.gms:google-services:3.2.1'

        // NOTE: Do not place your application dependencies here; they belong
        // in the individual module build.gradle files
    }
    repositories {
        jcenter()
        google()
    }
}

ext {
    buildToolsVersion = "28.0.2"
    minSdkVersion = 16
    compileSdkVersion = 28
    targetSdkVersion = 28
    supportLibVersion = "27.1.0"
    googlePlayServicesVersion   = "15.0.1"
}

allprojects {
    repositories {
        mavenLocal()
        google()
        jcenter()
        maven {
            // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
            url "$rootDir/../node_modules/react-native/android"
        }
    }
}

```

Issue #1708, opened 2018-07-25
takePictureAsync leads to a crash on android 21/22


TBouder:
### Which implementation are you using
RNCamera

### Steps to reproduce
1. Open a simulator with Lollipop (android 5.1.1 for instance)
2. Open the Camera
3. Take the picture

### Expected behaviour
Should not crash

### Actual behaviour
Crash

### Environment
- **React version**:  16.4.1
- **React Native version**: 0.56.0
- **React Native platform + platform version**: Android 5.1.1

### react-native-camera
**Version**: ^1.1.4 || master

Here is the code we are using. Basicaly it's working perfectly on the newest version of Android/IOS, but on android Lollipop (v21/22) everytime `takePictureAsync` is called, we have a crash (both on simulator and on physical device)
```
takePicture = async () =>
{
	if (!this.camera)
		return;

	const	options = {width: 1080, base64: false, quality: 1};
	const	data = await this.camera.takePictureAsync(options); //CRASH HERE
	//DO STUFF
}

render = () =>
{
	return (
		<RNCamera
			key={'CAMERA'}
			ref={ref => this.camera = ref}
			style={cameraStyle}
			mirrorImage={this.state.cameraType === 'front'} />
	);
}
```

![screenshot_1532515081](https://user-images.githubusercontent.com/9974362/43196239-9c7e421c-9007-11e8-824d-a49ff1e20205.png)
```
07-25 10:37:58.801 6746-11210/com.shokimap E/unknown:ReactNative: Exception in native call
    java.lang.NullPointerException: Attempt to invoke virtual method 'double java.lang.Double.doubleValue()' on a null object reference
        at com.facebook.react.bridge.ReadableNativeArray.getDouble(ReadableNativeArray.java:114)
        at com.facebook.react.bridge.JavaMethodWrapper$4.extractArgument(JavaMethodWrapper.java:64)
        at com.facebook.react.bridge.JavaMethodWrapper$4.extractArgument(JavaMethodWrapper.java:60)
        at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:359)
        at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:160)
        at com.facebook.react.bridge.queue.NativeRunnable.run(Native Method)
        at android.os.Handler.handleCallback(Handler.java:739)
        at android.os.Handler.dispatchMessage(Handler.java:95)
[...]

PR #1853, opened 2018-10-13
Fixed first example

xenu256:
Solves many repeated Android permissions and taking pictures error reports, ex: #1708


sibelius:
RCTCamera is deprecated

Issue #1112, opened 2018-01-18
ios crash when scan barcode many times


picksomething:
### Steps to reproduce
1.open camera to scan barcode
2.close camera after finish scanning
3.reopen camera to scan barcode

### Expected behaviour
Work as well

### Actual behaviour
Crash

### Environment
- **Node.js version**:v6.11.3
- **React Native version**:0.48.4
- **React Native platform + platform version**: iOS 11.1.2, iPhone x, iPhone 6s

### react-native-camera
**Version**:  "master"

### crash log
```
Incident Identifier: 77357851-2BE1-4DF9-BF62-E88444270145
Beta Identifier:     5AA92AC4-B9CB-4283-A406-77661F1240F5
Hardware Model:      iPhone10,3
Process:             express_cross_platform [4289]
Path:                /private/var/containers/Bundle/Application/E35E6BF3-CB95-4942-992B-1362E55949F6/express_cross_platform.app/express_cross_platform
Identifier:          com.eacent.kuaidihui
Version:             42 (2.2.1)
Beta:                YES
Code Type:           ARM-64 (Native)
Role:                Foreground
Parent Process:      launchd [1]
Coalition:           com.eacent.kuaidihui [1323]


Date/Time:           2018-01-17 10:55:19.3798 +0800
Launch Time:         2018-01-17 10:39:33.0876 +0800
OS Version:          iPhone OS 11.1.2 (15B202)
Baseband Version:    1.41.00
Report Version:      104

Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Triggered by Thread:  0

Last Exception Backtrace:
0   CoreFoundation                	0x181aa9d04 __exceptionPreprocess + 124 (NSException.m:166)
1   libobjc.A.dylib               	0x180cf8528 objc_exception_throw + 56 (objc-exception.mm:521)
2   CoreFoundation                	0x181aa9390 __NSFastEnumerationMutationHandler + 128 (NSEnumerator.m:133)
3   UIKit                         	0x18aedc2a4 -[UIView(Hierarchy) _makeSubtreePerformSelector:withObject:withObject:copySublayers:] + 372 (UIView.m:9727)
4   UIKit                         	0x18b1bff88 __UIViewWasRemovedFromSuperview + 136 (UIView.m:9232)
[...]

PR #1868, opened 2018-10-22
fix for "ios crash when scan barcode many times #1112"

alimek:


alimek:
@sibelius cannot make app crashing using RNCamera, so should be ok.

Issue #1883, opened 2018-10-24
Setup automated release


sibelius:
based on this https://github.com/callstack/react-native-testing-library/issues/23

using https://github.com/webpro/release-it

sibelius:
don by @n1ru4l 

great job

PR #1891, opened 2018-10-26
chore: add release-it for easier package publishing

n1ru4l:
implements #1883

If we would use angular like conventions (https://github.com/webpro/release-it#custom-or-conventional-changelog), we could automate the release process further by letting circle-ci do the releases (similar to https://github.com/semantic-release/semantic-release which is used by packages like https://github.com/paypal/downshift.


sibelius:
we can use release-it, it will release only when we create a new git tag?

what's your npm package user? I'm gonna add your there, so you can test it

n1ru4l:
@sibelius release-it will create the git tags. The idea was that if something is merged into master, release-it will check whether a patch,minor or major release should be published and if so will do all the stuff like bumping the package.json version, generating the changelog, creatign the tag and the github release and publishing to npm

My npm use is the same as on github (n1ru4l) https://www.npmjs.com/~n1ru4l

n1ru4l:
@sibelius I updated the setup

Conventional commit messages can now be done interactively by using `yarn commit`. We could go one step further and enforce the correct commit format for pull requests by using https://www.npmjs.com/package/commitlint-circle . However we could also squash a pull requests and add a conventional commit message ourselves.

What do you think?

I will also update the readme or add a contribution guide

PR #1893, opened 2018-10-29
Fix:oom caused by decodeByteArray

zhaomengfan:
Solve the oom problem on some Android device caused by decodeByteArray

### DIFF
--- a/android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
+++ b/android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
@@ -76,7 +76,7 @@ public class ResolveTakenPictureAsyncTask extends AsyncTask<Void, Void, Writable
 
         // we need the stream only for photos from a device
         if (mBitmap == null) {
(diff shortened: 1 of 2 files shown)


android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
```diff
@@ -76,7 +76,7 @@ protected WritableMap doInBackground(Void... voids) {
 
         // we need the stream only for photos from a device
         if (mBitmap == null) {
-            mBitmap = BitmapFactory.decodeByteArray(mImageData, 0, mImageData.length);
+            mBitmap = ImageCacheUtils.getResizedBitmap(mImageData, 400);
```
listicos: Is this a height limit ? So we can't have pictures bigger than 400 ?

sibelius: should this be a prop?

listicos: I think mImageData.length was correct and in that case it doesn't make sense this PR then.

PR #1896, opened 2018-11-01
feat: added inverted gv barcode scanning

Sjiep:
This PR adds the ability to scan inverted barcodes when using the Google Vision barcode detector through two additional "modes", `inverted`, `alternate` and `normal` (default). 

- `inverted`: each image will be inverted before being passed to the barcode detector. In this case only "inverted" barcodes can be detected.
- `alternate`: each second frame will be inverted before being passed to the barcode detector. This mode will detect both "normal" and "inverted" barcodes.

When no mode is provided the barcode scanner behaves like normal and only "normal" barcodes can be detected.

### DIFF
--- a/android/src/main/java/org/reactnative/barcodedetector/RNBarcodeDetector.java
+++ b/android/src/main/java/org/reactnative/barcodedetector/RNBarcodeDetector.java
@@ -9,6 +9,10 @@ import org.reactnative.frame.RNFrame;
 
 public class RNBarcodeDetector {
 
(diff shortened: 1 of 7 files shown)


n1ru4l:
@Sjiep Thank you for your efforts!

manishkumar03:
Now that Google Vision barcode detector is on working on iOS, can this functionality be added to iOS please?

sibelius:
sure, can you send a PR?

Issue #1292, opened 2018-02-26
RNCameraManager crash issues after QRCODE read


itrcz:
### Steps to reproduce
1. scanning QRCode with weird encoding on something like that
2. - (void)captureOutput:(AVCaptureOutput *)captureOutput didOutputMetadataObjects:(NSArray *)metadataObjects fromConnection:(AVCaptureConnection *)connection  calls
3. metadata.stringValue is NULL that causes application crash.

### Expected behaviour
metadata.stringValue should be not be NULL

### Actual behaviour
metadata.stringValue is NULL

### Environment
- **Node.js version**: 8.9.1
- **React Native version**: 0.51.0
- **React Native platform + platform version**: iOS 11.2

### react-native-camera
**Version**: npm version 1.0.1

### QRCODE that causes issue
![image 2018-02-26 19 23 25](https://user-images.githubusercontent.com/2022624/36681781-9fad29d2-1b2a-11e8-83cb-31d1be71fbba.jpg)

gre:
can reproduce with this QRCode:

![qrcode](https://user-images.githubusercontent.com/211411/37468218-a06c6604-2862-11e8-88b3-a269867b79b3.png)


(generated on https://smealum.github.io/ninjhax2/ )

gre:
@itrcz were you able to still reproduce?

Issue #926, opened 2017-10-03
Scanning QR code crashes app iOS only


justynspooner:
This QR code causes a crash on iOS only.

![screen shot 2017-10-03 at 11 57 44](https://user-images.githubusercontent.com/1853660/31121993-2a062dfc-a832-11e7-8ad9-36f8f3403c26.png)

in `RCTCameraManager.m`

in the method `- (void)captureOutput:(AVCaptureOutput *)captureOutput didOutputMetadataObjects:(NSArray *)metadataObjects fromConnection:(AVCaptureConnection *)connection`

```
@"data": metadata.stringValue,
```

calling stringValue causes the crash.

fokoz:
Mine crashes only when specify `barCodeTypes` to qr but if not specify, it is fine.

prajnarobo:
Any solution for the above issue? App is crashing even after removing barCodeTypes

hariDasu:
this works with the above merged PR, this issue can be closed now @prajnarobo @justynspooner

sibelius:
use RNCamera instead of RCTCamera

Issue #691, opened 2017-05-04
NSInvalidArgumentException is thrown by some QR Codes


januswel:
### Steps to reproduce
1. Write components using `onBarCodeRead`
2. Read below QR code

https://img.altema.jp/pokemon-sunmoon/qrcode/2.png

### Expected behaviour

Returns an event from Objective-C tier

### Actual behaviour

```
2017-05-04 22:42:26.168842+0900 QRCodeReader[19699:5828107] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[1]'
```

### Environment
- **Node.js version**: 4.7.1
- **React Native version**: 0.43.4
- **React Native platform + platform version**: iOS

### react-native-camera
**Version**: 0.7.0

PR #1904, opened 2018-11-05
Return raw barcode data if available

cozmo:
Currently on iOS if you scan a barcode that doesn't have a textual representation such as 

Bad QR code

The app will crash. This is obviously not the desired behavior, there should never be a situation where a user can scan something that causes the app to crash, especially when it's a _valid_ barcode. 

This issue has come up a few times, such as #1292, #926 and #691.

At the bare minimum we should check for `stringValue` to exist so we don't _crash_. That said, this isn't super useful for those of us who actually want to scan these (valid) barcodes. Better would be to return the raw barcode data if it exists, so even if it doesn't have a valid string representation the consumer can still read handle it as they'd like. 

This PR does just that. In android it's pretty straightforward since ZXing exposes `getRawBytes` to get well, the raw bytes. In iOS it's a fair bit more rigamarole. In iOS11+ we can use [CIBarcodeDescriptor](https://developer.apple.com/documentation/coreimage/cibarcodedescriptor?language=objc)'s subclasses to access `errorCorrectedPayload`, which is the data we want. On earlier versions we have to peak into the `AVMetadataMachineReadableCodeObject` internals which is a little annoying, but not [without](https://stackoverflow.com/a/36957666) [precedent](https://github.com/behestee/cordova-plugin-multipartqrcodescanner/blob/master/src/ios/CDVBarcodeScanner.mm#L604). 

Tested that the barcodes that previously failed to scan now scan properly. Additionally tested barcodes in formats that wouldn't have rawData (such as `code39`) to ensure we're backwards compatible.

I'm not an expert in Objective C/Java best practices to happy to make any changes if there are better ways to write anything.

### DIFF
--- a/android/src/main/java/org/reactnative/camera/events/BarCodeReadEvent.java
+++ b/android/src/main/java/org/reactnative/camera/events/BarCodeReadEvent.java
@@ -12,6 +12,7 @@ import com.google.zxing.Result;
 import com.google.zxing.ResultPoint;
 
 import java.util.Date;
+import java.util.Formatter;
 
 public class BarCodeReadEvent extends Event<BarCodeReadEvent> {
   private static final Pools.SynchronizedPool<BarCodeReadEvent> EVENTS_POOL =
@@ -68,6 +69,14 @@ public class BarCodeReadEvent extends Event<BarCodeReadEvent> {
 
     event.putInt("target", getViewTag());
     event.putString("data", mBarCode.getText());
+
+    Formatter formatter = new Formatter();
+    for (byte b : mBarCode.getRawBytes()) {
+      formatter.format("%02x", b);
+    }
+    event.putString("rawData", formatter.toString());
+    formatter.close();
+
     event.putString("type", mBarCode.getBarcodeFormat().toString());
     WritableArray resultPoints = Arguments.createArray();
     ResultPoint[] points = mBarCode.getResultPoints();
--- a/docs/RNCamera.md
+++ b/docs/RNCamera.md
@@ -319,7 +319,10 @@ Function to be called when native code emit onPictureTaken event, when camera ha
 
 Will call the specified method when a barcode is detected in the camera's view.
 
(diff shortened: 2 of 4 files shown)


cozmo:
Hey @n1ru4l 

Great questions. Your understanding of how I tested these changes is correct. I generally wrote them to be backwards compatible (aka we shouldn't ever fail to parse if we can't extract rawData), and then I wrote a very simple app (provided below) with each version of the SDK and ensured it could scan all the failing QR codes, as well as any normal codes that were already scannable. 

The app I used is provided below in it's entirety:
```js
class App extends Component {
  render() {
    return (
      <View style={{flex: 1}}>
        <RNCamera
          permissionDialogTitle="Permission to use camera"
          permissionDialogMessage="We need your permission to use your camera phone"
          style={{width: "100%", height: "100%"}}
          onBarCodeRead={(code) => {
            Alert.alert("Found Barcode", `${code.data}:${code.rawData}`);
          }}
        />
      </View>
    );
  }
}

```

With my branch I've confirmed this can scan all the barcodes listed in the above issue reports, as well as still being able to scan barcodes that don't have any `rawData`, such as a `code128` barcode. 

Please let me know if there's any additional testing you think would make sense. 

As for performance - For a QR code, adding the raw data is ~ the same size as the string version of the payload, maybe a bit more because it has control characters/formatting bytes etc, and the hex encoding of the raw bytes is less efficient than the underlying string encoding in some cases. 

That said, if you're scanning large binary payloads, today they will crash, which is certainly worse. Also I think it's in general reasonable to expect that the larger the barcode you scan the worse the performance, since at the end of the day we're always going to need to pass that data over the bridge. I'm not exactly sure on bridge performance, but since this change isn't changing the size by an order of magnitude, and since at the end of the data we need to pass that data over the bridge in some form or another, I'm not sure what our options are here?

sibelius:
One option is the have a prop or parameter to enable returning rawData, but if the size is not that bad, I’m cool with this

PR #1886, opened 2018-10-24
[iOS] Simple way to prevent camera flickering when starting a recording adding a new property

listicos:
Hi,

I added a new property to fix issue [#1249](https://github.com/react-native-community/react-native-camera/issues/1249)

This property fixes the black flicker that many people are experiencing at the start of a recording. By setting this property on the component itself, the camera no longer records a black flicker at the start of a recording and it no longer flickers!

Related [#1542](https://github.com/react-native-community/react-native-camera/pull/1542), [#1857](https://github.com/react-native-community/react-native-camera/pull/1857),

PR #1908, opened 2018-11-07
Fix default resolution on iOS

### LINKED ISSUE #1900: Incompatible AVCAptureSessionPreset causes crash on iPad Mini 4
**Versions**
react-native-qrcode-scanner: 1.1.0
react-native-camera: 1.3.1
react-native: 0.44
react: 16.0.0-alpha.6
Platform: iOS 11/iOS 12

In my project I am implementing a camera view for scanning QR codes, using the react-native-qrcode-scanner library. This lib is dependent on react-native-camera.
Using the latest release of react-native-camera on master (1.3.1), opening the camera causes the app to crash when using an iPad Mini 4. I am not specifying any camera resolution or format, so the default value is used. 
This error gets printed in Xcode:

*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[AVCaptureSession setSessionPreset:] AVCaptureSessionPreset3840x2160 is not a supported preset'

Seems like this happens because RNCamera is trying to use the highest resolution video preset (4K video), but the camera in the iPad Mini does not support 4K video. 

**Expected Behaviour**
The camera should start without crashing.

**Additionals**
In my package.json: "react-native-camera": "git+https://git@github.com/react-native-community/react-native-camera",
I am using RNCamera without the FaceDetector folder.

cozmo:
I believe this is the fix to #1900, fixing a bug that was introduced in #1886. 

The root of the issue is that [this enum](https://github.com/react-native-community/react-native-camera/blob/master/ios/RN/RNCameraManager.h#L50-L56) for video resolution defines `0` as `RNCameraVideo2160p`, and `defaultVideoQuality` was an Integer, so if it was not set it was mapping to `0` and thus to `RNCameraVideo2160p`, causing any user of this library who didn't pass in an explicit `defaultVideoQuality` to request `AVCaptureSessionPreset3840x2160` which is far higher than many iOS cameras support, leading to the `AVCaptureSessionPreset3840x2160 is not a supported preset` error. 

This is starting to stretch both my knowledge of Objective C best practices and `react-native-camera` design decisions, but I attempted to fix this issue by switching to use `NSNumber` and then explicitly checking for it being `NULL` when when we set the sessionPreset, thereby only setting it if it's been explicitly set by the consumer. 

Not sure if there's a better way to do this? Have tested it and confirmed (with the debugger) that
- If you do not provide a `defaultVideoQuality` prop that we fall back to `AVCaptureSessionPresetHigh` (which we then convert to `AVCaptureSessionPresetPhoto` for reasons beyond me)
- If you provide a valid `defaultVideoQuality` prop, we use that instead. 

Fixes #1900

### DIFF
--- a/ios/RN/RNCamera.h
+++ b/ios/RN/RNCamera.h
@@ -43,7 +43,7 @@
 @property (nonatomic, assign) BOOL canReadText;
 @property(assign, nonatomic) AVVideoCodecType videoCodecType;
 @property (assign, nonatomic) AVCaptureVideoStabilizationMode videoStabilizationMode;
-@property(assign, nonatomic) NSInteger defaultVideoQuality;
+@property(assign, nonatomic, nullable) NSNumber *defaultVideoQuality;
 
 - (id)initWithBridge:(RCTBridge *)bridge;
 - (void)updateType;
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -605,7 +605,9 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
             return;
         }
 
-        AVCaptureSessionPreset preset = [RNCameraUtils captureSessionPresetForVideoResolution:[self defaultVideoQuality]];
+        // Default video quality AVCaptureSessionPresetHigh if non is provided
+        AVCaptureSessionPreset preset = ([self defaultVideoQuality]) ? [RNCameraUtils captureSessionPresetForVideoResolution:[[self defaultVideoQuality] integerValue]] : AVCaptureSessionPresetHigh;
+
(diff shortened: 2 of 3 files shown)


cozmo:
@n1ru4l Thanks for merging - What is the process to get this change (and similarly #1904 and #1907) published on NPM? Should I make a PR with a version bump to package.json or is that something you all (maintainers) handle? 

Would love to go back to using npm+semvar to install this dep vs a git url+hash in my `package.json`.

n1ru4l:
@cozmo wenn are currently in the process of automating releases which I will hopefully complete on the weekend. You will have to use themaster branch for now, sorry!

PR #1911, opened 2018-11-08
Update Camera1.java

marekchen:
fix java.lang.RuntimeException: Camera is being used after Camera.release()

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -143,22 +143,25 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
 
     @Override
     void stop() {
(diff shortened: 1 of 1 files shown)


n1ru4l:
@marekchen First of all thank you for your contribution!

Before merging could you please explain how to reproduce the `fix java.lang.RuntimeException: Camera is being used after Camera.release()` crash and provide a test plan for this fix?

n1ru4l:
@SimonErm What is your opinion on these changes?

SimonErm:
@n1ru4l  I am not sure which of these functions throws the exception and your right, just catching without handle it is not the way we should do. I think if we know how to reproduces this, we can fix the root problem.

n1ru4l:
Closing this since it is not the correct solution. Please feel free to open a new pull request that addresses this issue correctly.

PR #1916, opened 2018-11-12
fix(barcode): resolve crash when barcode does not include raw bytes

n1ru4l:


VinceBT:
I would greatly like this to be merged and involve a new version 👍

n1ru4l:
@VinceBT Have you tested the branch?

Issue #1910, opened 2018-11-07
setPreviewTexture Failed


lothindil:
# Bug Report
**To Do First**
- "react-native-camera": "^1.2.0"
- "react-native-qrcode-scanner": "^1.0.1",

**Related Modules**
RNCamera / qrcode scanner

**Platforms**
Android 7

**Versions**
- Android: 7.12
- react-native-camera: 1.2.0
- react-native: 0.55.4
- react: 16.3.1

**Description/Current Behaviour**
Error "SetPreviewTexture" after read of qr code
https://drive.google.com/file/d/1BjTwd_zfrmTCwBHQaCxPyf8YME1cAHgU/view?usp=sharing

**Expected Behaviour**
Nothing, just go to next part of scene

**Steps to Reproduce**
My app' is conceived to use several times three times camera : 1 scan, 1 photo, 1 scan.

We must be able to do this several times.

We navigates screen to screen with modify state on a reducer. Basically our system is : 

(0) state : "Start" > welcome page (indexScene.js) > button "scan" > actionScene > reducer Scene
(1) state : "ItemScan" > scan qr code (scanScene.js) > actionScene > reducerScene
(2) state : "ItemForm" > form (scanScene.js) > actionScene > reducerScene
(3) state : "ItemPict" > photo  (pictureScene.js) > actionScene > reducerScene
(4) state : "WHScan" > scan qr code (scanScene.js) > actionScene > reducerScene

"continue" or "leave" button (scanScene.js) > actionScene > reducerScene 

If "continue" return to 1.
if "leave" return to 0.

leave + button scan = continue

The issue is in the beggining of step 2.

And don't depend about the choice.  I don't have any way to follow...

SimonErm:
Can you provide the code of your camera and scanner component?

lothindil:
For sure.

photoPage.js

    render()
    {
        return (
        
        <View style={styles.container}>
            <Text>Prendre une photo</Text>
          <RNCamera
              ref={ref => {
                this.camera = ref;
              }}
              style = {styles.preview}
              type={RNCamera.Constants.Type.back}
              flashMode={RNCamera.Constants.FlashMode.on}
              permissionDialogTitle={'Permission to use camera'}
              permissionDialogMessage={'We need your permission to use your camera phone'}
          />
          <View style={{flex: 0, flexDirection: 'row', justifyContent: 'center',}}>
          <TouchableOpacity
              onPress={this.takePicture.bind(this)}
[...]

Issue #1837, opened 2018-10-06
Incompatibilities with Camera2Api. New smartphones uses old Camera API


fxcosta:
# Bug Report
**To Do First**
- [x ] Did you try latest release?
- [x ] Did you try master?
- [x ] Did you look for existing matching issues?

**Related Modules**

**Platforms**

**Versions**
- Android: 8.1
- react-native-camera: ^1.2.0
- react-native: 0.57.0
- react: 16.5.0

**Description/Current Behaviour**
I'm implementing an app that uses a lot of camera features like exposure, manual focus, white balance, etc.

I saw that there is an API for setting depth of focus (this would be equivalent to manual focus?) And I tried to use it. Then I saw that this is only available for phones that have Camera2Api. The real problem is that my phone is a Moto G5S Android 8.1 SDK 27 and it still uses the old camera API.

The library decides whether or not I can use the Camera2Api features based on the SDK and, as I have 27, the library allows me to try to use the features and break when I use the react-native-camera prop useCamera2Api.

Has anyone else had this problem? I'm finding very strange new cellphones with an old api. How can I check now whether or not cell phones actually have access to Camera2Api?

The weirdest? According to the Camera2 Probe app, I would like to set up manual focus. However, debugging the app, I realized that at no time does it use the Camera2 class, but Camera1 instead. If I put the prop useCamera2Api, like I said, the app breaks because my Hardware Level Support Category is Legacy.

SimonErm:
Which targetSDK are you using? I think this could cause the fallback. I am facing the crash on useCamera2Api prop too and going to fix it this week.

PR #1928, opened 2018-11-19
Fix for useCamera2Api

SimonErm:
Fixes the issue which lead to a crash by using useCamera2Api Prop, which was caused by setPictureSize() before view was startet and not initialized. Furthermore this PR fixes the stopPreviewAfterCapture Key to pauseAfterCapture which caused the Camera2 to freeze after takePicture, even if pauseAfterCapture was set to false.
#1837
Maybe related: #1910

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera2.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera2.java
@@ -1053,8 +1053,8 @@ class Camera2 extends CameraViewImpl implements MediaRecorder.OnInfoListener, Me
                         public void onCaptureCompleted(@NonNull CameraCaptureSession session,
                                 @NonNull CaptureRequest request,
                                 @NonNull TotalCaptureResult result) {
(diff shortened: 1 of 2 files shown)


n1ru4l:
Just out of curiosity: What are the differences between Camera1 and Camera2? Will Camera2 solve the performance issues that Camera1 has?

sibelius:
you can read more about Camera2 api here

https://github.com/google/cameraview

Camera1 is deprecated, and we should use Camera2 api when dealing with android 5 or more

the problem is how to make a code that works both on Camera1 and Camera2

SimonErm:
@n1ru4l Thank you for this information :-)

fxcosta:
> Camera1 is deprecated, and we should use Camera2 api when dealing with android 5 or more
> 
> the problem is how to make a code that works both on Camera1 and Camera2

@sibelius Not exactly. My android is 8 and do not have camera2 api... its strange, but here u can see whats happen: #1837

PR #1932, opened 2018-11-19
fix(ios): only mute recording if mute property is set to false

### LINKED ISSUE #1864: [Video Record] has no sound
# Bug Report
**To Do First**
- [x] Did you try master?

**Related Modules**
RNCamera

**Platforms**
iOS

**Versions**

- Android:
- iOS: 11
- react-native-camera: 1.3.0
- react-native: 0.56.1
- react: 16.4.1

Record video with no sound even set `captureAudio=true`, have tested with VLC player in the PC and also with `react-native-video `no sound at all.

**Expected Behaviour**
The sound should be recorded.

**Steps to Reproduce**
1. Record video with `captureAudio=true` and allow the permission.
2. Video has only image with no sound.

n1ru4l:
Closes #1864

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -522,7 +522,7 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
         }
     }
 
-    [self updateSessionAudioIsMuted:!!options[@"mute"]];
+    [self updateSessionAudioIsMuted:[options[@"mute"] boolValue]];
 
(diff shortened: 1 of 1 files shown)


mcorner:
I believe this changed the default to mute on iOS if you don't specify anything.  We lost audio in recordings when we upgraded and mute: false fixes it.  The default should probably go back to true.

n1ru4l:
@mcorner sorry to hear that. Could you please send a pull request to fix this issue?

Issue #1725, opened 2018-08-02
RNCamera Android Orientation


Tsiniloiv:
So, RNCamera on Android will frequently save pictures in the wrong orientation. This seems to happen especially when the app's orientation has been locked - in my case, my app is locked to landscape orientation, and pictures will be saved correctly in one orientation, and upside down in the other orientation.

I'm very much not an expert on this, but I've been digging through the code and other posted issues, and near as I can tell this happens because RNCamera is finding its orientation by listening for onOrientationChanged events. (I don't know if locking the orientation blocks these events from firing or what, but like I said, very much not an expert)

The fixOrientation option seems to take the final image, read its real exif orientation, then processes the image to match that, I think? The big problem is that it's slow as hell, especially compared to RCTCamera.

So I figured I'd check in case I've missed a solution or workaround somewhere - has anybody figured out how to fix this, or come up with a workaround that doesn't multiply the amount of time it takes to take a picture?

If not, I _think_ I may have found where the camera orientation change happens. Please correct me if I'm wrong: The RNCameraView implements Google's CameraView, and just straight up calls super.takePicture(). CameraView seems to be a very nice and neat and convenient wrapper around the Camera API, so that's nice, except it bases its orientation on listening for onOrientationChanged events.

So, in theory, adding a check for the device's real orientation to CameraView's takePicture() function might solve the orientation issue without needing to reprocess every single picture.

EDIT: Yeah, I was mostly wrong, slightly right. CameraView uses yet another wrapper around the Camera API, and that's where the API is really called. Camera1.java takePictureInternal(). Finally, something I recognize from the old version.

Heading to lunch now, but I'm gonna try that after. Figured I'd put this up. Am I on totally the wrong path here? Has anybody solved this, so I don't have to waste time being wrong or duplicating someone else's solution?

SimonErm:
https://github.com/react-native-community/react-native-camera/issues/1662 i think it's the same like on ios. We should use accelerometer or gyroscope to get orientation. I think the native Android camera does the same.

Tsiniloiv:
I don't know anything about iOS coding, unfortunately, but it's a relief to know my train of though isn't terribly wrong.

I did some digging on the Android side and turns out I was mostly wrong, slightly right. CameraView uses yet another wrapper around the Camera API, and that's where the API is really called. The orientation listener thing is solid - it's to rotate the display when the display is supposed to rotate, I think.

[...]

PR #1941, opened 2018-11-22
Picture orientation from sensor

JulienUsson:
Hello,

The camera take device orientation from sensor and no more from the exif data.
I had orientation problems with an app locked in portrait.

Thanks,
Julien Usson

### DIFF
--- a/android/src/main/java/org/reactnative/camera/RNCameraView.java
+++ b/android/src/main/java/org/reactnative/camera/RNCameraView.java
@@ -43,6 +43,7 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
   private Promise mVideoRecordedPromise;
   private List<String> mBarCodeTypes = null;
   private Boolean mPlaySoundOnCapture = false;
+  private SensorOrientationChecker mSensorOrientationChecker;
 
   private boolean mIsPaused = false;
   private boolean mIsNew = true;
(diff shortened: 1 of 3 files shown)


n1ru4l:
Hey @JulienUsson thank you for your contribution!

Could you please provide a test plan? Like: `hold phone with example app like this, expect image to be like this`. Before and after your changes. That would help enormously!

JulienUsson:
With an app locked in portrait
```
<activity android:name=".MainActivity"
    android:configChanges="orientation"
    android:screenOrientation="portrait"/>
```

When you take a picture in landscape with `this.camera.takePictureAsync({ fixOrientation: true })`

Before : the saved image was not rotated.

After : the picture is saved with the correct orientation.

SimonErm:
I think this is a workaround. There already is a OrientationChecker(https://github.com/react-native-community/react-native-camera/blob/master/android/src/main/java/com/google/android/cameraview/DisplayOrientationDetector.java), but i think it's not working correctly and that's causing the CameraApi to set wrong exif data sometimes or rotate the picture wrong.

JulienUsson:
My SensorOrientationChecker.java comes from the RCTCamera code because orientation works well with it.
My PR doesn't work on all device like in RCTCamera after doing some tests with other devices.

JulienUsson:
This new version, use exif data and sensor data and now the picture is correctly oriented in all devices.

n1ru4l:
Does this issue address https://github.com/react-native-community/react-native-camera/pull/1941 ?

JulienUsson:
This PR correct the issue #1725

Teal-labs-developer:
@JulienUsson, just one little concern. I think that changing rotating bitmap after picture is captured takes few seconds to process. So why didn't you use the setRotation method for Camera.Parameters?

SimonErm:
@n1ru4l i still think we should fix DisplayOrientationChecker and add an orientation option like we have on ios

n1ru4l:
@SimonErm I agree with you!

@Teal-labs-developer Do you mean when using the `fixOrientation` option?

Issue #1934, opened 2018-11-20
The camera view flickers with a dark blue mask when start/stop recording


lucaspang:
# Bug Report
**To Do First**
- [x] Did you try latest release?
- [x] Did you try master?
- [x] Did you look for existing matching issues?

**Related Modules**
RNCamera

**Platforms**
iOS

**Versions**
- iOS: 12.0
- react-native-camera: 1.4.2
- react-native: 0.57.2
- react: 16.5.2

**Description/Current Behaviour**
The camera area flickers with a dark blue mask for about 0.1 second whenever I called recordAsync and stopRecording

**Expected Behaviour**
No mask should be covering the camera view

**Steps to Reproduce**
The camera component: 
```
  <RNCamera
    ref={ref => {
      this.camera = ref;
    }}
    style={{
      flex: 1
    }}
    type={RNCamera.Constants.Type.back}
    flashMode={RNCamera.Constants.FlashMode.off}
  >
    {({ camera, status }) => {
      if (status !== "READY")
        return <View style={{ flex: 1, backgroundColor: "blue" }} />;
      return (
        <View
          style={{
            flex: 1,
            flexDirection: "row",
            justifyContent: "center",
            alignItems: "flex-end"
          }}
        >
          <TouchableOpacity style={{ height: 50 }} onPress={this.buttonClick}>
            <Text>{this.state.isRecording == true ? "Stop" : "Start"}</Text>
          </TouchableOpacity>
        </View>
      );
    }}
  </RNCamera>
```

The start recording function:
const { uri, codec = 'mp4' } = await this.camera.recordAsync()
const result = await CameraRoll.saveToCameraRoll(uri, 'video')

The stop recording function:
this.camera.stopRecording()

**Additionals**
Here are some screen cap to help you understand what is happening:
I am filming the following scene
<img width="382" alt="screen shot 2018-11-20 at 4 50 32 pm" src="https://user-images.githubusercontent.com/7286004/48761852-9e758280-ece4-11e8-94b4-9e0cdf6c136c.png">

When I called recordAsync and stopRecording, the view follows with a blue mask or something on the view, sometimes the recorded video captured the blue mask.
<img width="383" alt="screen shot 2018-11-20 at 4 50 58 pm" src="https://user-images.githubusercontent.com/7286004/48761940-d381d500-ece4-11e8-849b-20d969ea5c25.png">

**Edit (n1ru4l)**:

- Prettified code sample

n1ru4l:
[...]

PR #1942, opened 2018-11-22
[WIP] Adds defaultMuted prop to help with flickering

maxschmeling:
See https://github.com/react-native-community/react-native-camera/issues/1934#issuecomment-440963828 and https://github.com/react-native-community/react-native-camera/issues/1934#issuecomment-440966283

This is a work in progress.

- [x] Add `defaultMuted` prop and use it on initial session setup
- [x] Add `defaultMuted` documentation
- [x] Determine cause of flicker on first `stopRecording` and fix it
- [ ] Look into handling change of `defaultMuted` and `defaultVideoQuality` values

Refs #1934

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -606,7 +606,7 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
         }
 
         // Default video quality AVCaptureSessionPresetHigh if non is provided
-        AVCaptureSessionPreset preset = ([self defaultVideoQuality]) ? [RNCameraUtils captureSessionPresetForVideoResolution:[[self defaultVideoQuality] integerValue]] : AVCaptureSessionPresetHigh;
+        AVCaptureSessionPreset preset = ([self defaultVideoQuality]) ? [RNCameraUtils captureSessionPresetForVideoResolution:[self defaultVideoQuality]] : AVCaptureSessionPresetHigh;
 
         self.session.sessionPreset = preset == AVCaptureSessionPresetHigh ? AVCaptureSessionPresetPhoto: preset;
 
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -937,6 +937,7 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
 
     if ([self.session canAddOutput:movieFileOutput]) {
         [self.session addOutput:movieFileOutput];
+        [self updateSessionAudioIsMuted:self.defaultMuted];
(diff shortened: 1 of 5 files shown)


maxschmeling:
I've found that the flicker when calling `stopRecording` the first time is caused by the difference between these two lines:
https://github.com/react-native-community/react-native-camera/blob/b1d91eb91ab148356a2069e20b24c93068154b2f/ios/RN/RNCamera.m#L609 and https://github.com/react-native-community/react-native-camera/blob/b1d91eb91ab148356a2069e20b24c93068154b2f/ios/RN/RNCamera.m#L1003

The first one is in `startSession` and passes `[[self defaultVideoQuality] integerValue]]` to `RNCameraUtils.captureSessionPresetForVideoResolution` while the second is in `cleanupCamera` (called after the recording is saved) and passes in `[self defaultVideoQuality]`.

Changing them to both use `[self defaultVideoQuality]` seems to fix the flickering. I haven't confirmed yet that it's using the property video quality value though. Maybe they both need to be using `integerValue`? I'm thinking the way I did it is probably right, but I'll have to confirm that.

maxschmeling:
@n1ru4l I'm a little less comfortable putting the handling of the default value changes in this pull request. Are you OK with merging this as is? Or at least without those changes.

If you'd like, I can work on that in a separate pull request after this one is in.

ios/RN/RNCamera.m
```diff
@@ -606,7 +606,7 @@ - (void)startSession
         }
 
         // Default video quality AVCaptureSessionPresetHigh if non is provided
-        AVCaptureSessionPreset preset = ([self defaultVideoQuality]) ? [RNCameraUtils captureSessionPresetForVideoResolution:[[self defaultVideoQuality] integerValue]] : AVCaptureSessionPresetHigh;
```
n1ru4l: Please revert this line, it causes an instant app crash when using the `defaultVideoQuality` property after you changed this line.

maxschmeling: So is the other method the appropriate way to  use `defaultVideoQuality`? They need to be consistent, right? It definitely fixed the issue for me when stopping recording the first time.

n1ru4l:
We already have the `captureAudio` property. I am going to close this due to no recent activity. If someone wants to tackle this issue please implement it by using the `captureAudio` property.

maxschmeling:
I haven't been able to get back to this, but I will be soon. It's extremely important for making the camera feel smooth and responsive.

I do think the `captureAudio` approach makes more sense though. I'll look at that soon, but that would be a more clear API without adding additional properties.

PR #1952, opened 2018-11-26
Adds typing to CameraStatus constant

maxschmeling:
This provides the ability to use the `Status` type from consumers of `react-native-camera` like this:
```
type SomethingWithCameraStatus = {
  status: $Values<typeof RNCamera.Constants.CameraStatus>
};
```
It would be nice to export the `Status` type directly, but it seems like that's being intentionally avoided? Either way, this typing of `CameraStatus` is probably a good thing.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,7 @@
 // @flow
 
 import Camera from './Camera';
-import RNCamera, { type Status as CameraStatus } from './RNCamera';
+import RNCamera, { type Status as _CameraStatus } from './RNCamera';
 import FaceDetector from './FaceDetector';
 
 export type CameraStatus = _CameraStatus;
--- a/src/index.js
+++ b/src/index.js
@@ -4,7 +4,7 @@ import Camera from './Camera';
 import RNCamera, { type Status as CameraStatus } from './RNCamera';
 import FaceDetector from './FaceDetector';
 
-export type CameraStatus = CameraStatus;
+export type CameraStatus = _CameraStatus;
 export { RNCamera, FaceDetector };
 
 export default Camera;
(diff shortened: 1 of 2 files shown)


n1ru4l:
@sibelius Any reasons for not exporting CameraStatus?

sibelius:
we can export

n1ru4l:
@maxschmeling Feel free to also export it 😊

src/index.js
```diff
@@ -1,7 +1,10 @@
+// @flow
+
 import Camera from './Camera';
-import RNCamera from './RNCamera';
+import RNCamera, { type Status as CameraStatus } from './RNCamera';
 import FaceDetector from './FaceDetector';
 
+export type CameraStatus;
```
n1ru4l:
```suggestion
export type CameraStatus = CameraStatus;
```

You will have to change this

n1ru4l:
@maxschmeling The export is not valid syntax. Please change it, I will merge this pull request afterwards!

n1ru4l:
@maxschmeling The export is not valid syntax. Please change it, I will merge this pull request afterwards!

src/index.js
```diff
@@ -1,7 +1,10 @@
+// @flow
+
 import Camera from './Camera';
-import RNCamera from './RNCamera';
+import RNCamera, { type Status as CameraStatus } from './RNCamera';
```
n1ru4l:
```suggestion
import RNCamera, { type Status as _CameraStatus } from './RNCamera';
```

Sorry me fault

src/index.js
```diff
@@ -1,7 +1,10 @@
...
 import FaceDetector from './FaceDetector';
 
+export type CameraStatus = CameraStatus;
```
n1ru4l:
```suggestion
export type CameraStatus = _CameraStatus;
```

n1ru4l: this should work

maxschmeling:
That was my first time using suggested changes. I guess it makes sense that they don't go in as a group because they're individual suggestions. My bad. Pretty cool though.

Issue #1950, opened 2018-11-26
iOS: [UIApplication statusBarOrientation] must be used from main thread only


n1ru4l:
# Bug Report
**To Do First**
- [x] Did you try latest release?
- [x] Did you try master?
- [x] Did you look for existing matching issues?

**Related Modules**
RNCamera

**Platforms**
iOS

**Versions**

- Android:
- iOS:
- react-native-camera:
- react-native:
- react:

**Description/Current Behaviour**

Runtime Warning:

```- (UIInterfaceOrientation)getOrientationBy:(CMAcceleration)acceleration``` in `RNSensorOrientationChecker.m`

-[UIApplication statusBarOrientation] must be used from main thread only

**Expected Behaviour**
No runtime warning

**Steps to Reproduce**
Use rn-camera.

**Additionals**

<img width="1659" alt="bildschirmfoto 2018-11-26 um 11 57 36" src="https://user-images.githubusercontent.com/14338007/49010111-95683380-f172-11e8-9b06-4f9dcb88f5ca.png">

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate
> Want this issue to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

justinkiang:
I believe this issue is causing barcode recognition to not work when I turn the device landscape in an app that is locked to portrait

n1ru4l:
@justinkiang Wanna contribute a fix?

justinkiang:
That should do it

PR #1958, opened 2018-11-30
Fixing https://github.com/react-native-community/react-native-camera/…

justinkiang:
Fixes Issue #1950

### DIFF
--- a/ios/RN/RNCameraUtils.m
+++ b/ios/RN/RNCameraUtils.m
@@ -99,7 +99,11 @@
     CVImageBufferRef imageBuffer = CMSampleBufferGetImageBuffer(sampleBuffer);
     CIImage *ciImage = [CIImage imageWithCVPixelBuffer:imageBuffer];
     // set correct orientation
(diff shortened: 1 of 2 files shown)


n1ru4l:
@justinkiang Please convince me that using `dispatch_sync` is the correct approach here. I am not that experienced in objective c but I think dispatch_async would be the better approach?

reime005:
I think `dispatch_sync` makes more sense @n1ru4l, see [this line](https://github.com/react-native-community/react-native-camera/blob/2e680e28d90fded135799fd80520b842b1e6743a/ios/RN/RNCameraUtils.m#L106).

n1ru4l:
@justinkiang @reime005 Can you confirm this solves this issue:

>  I believe this issue is causing barcode recognition to not work when I turn the device landscape in an app that is locked to portrait
(https://github.com/react-native-community/react-native-camera/issues/1950#issuecomment-443014134)

I will try to review this ASAP

PR #1979, opened 2018-12-06
feat: add rncamera-example without face detection

n1ru4l:
I started migrating rncamera-example over to this repository. I think this will be a huge benefit in order to test new features!

**TODO**:

- Figure out way to download all the Frameworks (that have some megabytes) post install to avoid bloating this repository
- Reimplement FaceDetection (I currently removed everything, since it is insta crashing anyways)
- Add more information about the changes made in `rn-cli.config.js` in order/why react-native-camera is not a dependency in the `package.json` of the example app to not confuse people

**Follow Ups**:

- Add CocoaPods Example
- Freeze `rncamera-example` and point people over here


sibelius:
Can we have 2 or more examples?

One without FaceDetection
One with FaceDetection using postInstall
One with FaceDetection using Pod?

n1ru4l:
@sibelius Yes looking definitely forward  to adding those aswell!

n1ru4l:
@sibelius The recommended way to download all those framworks is using CocoaPods, however I think using react-native is hard enough. We should provide an easier way that is still automatic. I am thinking about providing a react-native-camera cli with commands for downloading the required libraries. Users of react-native-camera could then simply add the cli as a post install script!

sibelius:
We can add this example in this PR and add more examples in other PRs

We can check how rn-fbsdk does the linking part, we can definitely automate this steps, we can open an issue to improve this workflow

n1ru4l:
Also tested it on android, did not experience any new issues. Going to merge it!

Issue #1983, opened 2018-12-10
include deviceOrientation and orientation in RecordResponse


n1ru4l:
# Feature-Request

**Describe the Feature**

I am developing an app that records videos in the portrait mode. Therefore I force the orientation Portrait mode with the following code:

```js
const options = {
  ...remainingOptions,
  orientation: 1
};

const result = await camera.recordAsync(options);
```

However I need to know the actual deviceOrientation when the video recording started on the JS Thread.

**Possible Implementations**
Expose a new property `deviceOrientation` on the result returned by `camera.recordAsync(options)`.
This property includes the actual orientation of the device. Furthermore for integrity we can also add the `videoOrientation` property to the result.

@sibelius Are you okay with me implementing this feature?

```js
const options = {
  ...remainingOptions,
  orientation: 1
};

const result = await camera.recordAsync(options);
console.log(result.videoOrientation, result.deviceOrientation);
```

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate
> Want this feature to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

sibelius:
I think we can provide this

PR #1985, opened 2018-12-10
feat(rn-camera): deviceOrientation, videoOrientation, pictureOrientation

n1ru4l:
- [x] ios
- [x] android
- [x] documentation

Fixes https://github.com/react-native-community/react-native-camera/issues/1983

Features:

- implements `orientation` option on `recordAsync` and `takePictureAsync`
- adds field `deviceOrientation` and `videoOrientation` to `recordAsync` result
- adds field `deviceOrientation` and `pictureOrientation` to `takePictureAsync` result

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -477,20 +477,23 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
 - (void)recordWithOrientation:(NSDictionary *)options resolve:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRejectBlock)reject{
     [self.sensorOrientationChecker getDeviceOrientationWithBlock:^(UIInterfaceOrientation orientation) {
         NSMutableDictionary *tmpOptions = [options mutableCopy];
-        tmpOptions[@"orientation"]=[NSNumber numberWithInteger:[self.sensorOrientationChecker convertToAVCaptureVideoOrientation: orientation]];
+        if ([tmpOptions valueForKey:@"orientation"] == nil) {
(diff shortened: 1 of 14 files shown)


ios/RN/RNCamera.m
```diff
@@ -522,7 +531,9 @@ - (void)record:(NSDictionary *)options resolve:(RCTPromiseResolveBlock)resolve r
         }
     }
 
-    [self updateSessionAudioIsMuted:[options[@"mute"] boolValue]];
+    if ([options valueForKey:@"mute"] != nil) {
```
n1ru4l: I also fixed this regression so it prevents the flickering if no mute property is present in `options["mute"]`. Nevertheless we should implement `defaultMuted` asap.

n1ru4l:
@sibelius This is only for capturing the initial device orientation when starting the recording/taking the picture. This will not include orientation changes over time. Nevertheless that information could be quite useful for cases where you want to train a machine learning model, so we cloud implement it in a future pull request (once somebody needs that information).

n1ru4l:
I will now try to implement the android counterpart

sibelius:
If orientation changes over time, does the width and height of video changes as well? This could be enough to check device orientation

n1ru4l:
@sibelius I also checked it on android and ran into this bug: https://github.com/react-native-community/react-native-camera/issues/1988

sibelius:
have you tried to use PermissionsAndroid ?

try with latest version of react native

permissions android had some bug with promises before

n1ru4l:
@sibelius I don't understand your suggestion. When I record without rotation everything works fine. This should not be related to permissions  or am I missing out on something?

sibelius:
try to add some breakpoints on android studio to check where the promise is broke

n1ru4l:
I now implemented the same features on android + implemented the `orientation` option which was documented as implemented on both platforms but actually missing on android.

PR #1989, opened 2018-12-12
chore: format typescript typings with prettier

n1ru4l:


sibelius:
we can have this on lint-staged

we can also use tslint on .ts types

n1ru4l:
@sibelius sure I will add it

PR #1992, opened 2018-12-13
fix(android): use device orientation for images/videos instead of display orientation

### LINKED ISSUE #1053: Orientation of captured photo wrong
I implemented camera like in documentation.
If I take photo with camera and phone is landscape (lock button up) image orientation is good.
But if I do the same with lock button down image is turned upside down.
On android work fine this issue is on IOS.

Does anybody have any idea how to fix this?

UPDATE

If I add fixOrientation to true it fix issue and orientation is good but I see that this function is deprecated.
Even the portrait taken photos are landscape when show them in browser.
Why I need to call this function to get good orientation if in documentation it says that it is not needed any more? What could be the reason I am getting all photos in landscape.

### LINKED ISSUE #1725: RNCamera Android Orientation

n1ru4l:
This is my attempt for fixing #1725 based on the work of @Fsarmento (https://github.com/react-native-community/react-native-camera/issues/1725#issuecomment-441599916 https://github.com/react-native-community/react-native-camera/compare/master...Fsarmento:master)

This PR unifies the orientation used for the pictures/video with the `ios` implementation by using the device orientation instead of the display orientation.

**Related:**

Fixes #1725
Closes #1941
Fixes #1053

cc @sibelius @JulienUsson @SimonErm @Tsiniloiv

Could you please test this implementation? I tested all orientations (including when device is locked in portrait mode). But before merging I want at least one more person to test the implementation.

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -98,6 +98,8 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
 
     private int mDisplayOrientation;
 
+    private int mDeviceOrientation;
+
     private int mOrientation = Constants.ORIENTATION_AUTO;
 
     private float mZoom;
@@ -156,7 +158,7 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
             mMediaRecorder = null;
 
             if (mIsRecording) {
-                int deviceOrientation = displayOrientationToOrientationEnum(mDisplayOrientation);
+                int deviceOrientation = displayOrientationToOrientationEnum(mDeviceOrientation);
                 mCallback.onVideoRecorded(mVideoPath, mOrientation != Constants.ORIENTATION_AUTO ? mOrientation : deviceOrientation, deviceOrientation);
                 mIsRecording = false;
             }
@@ -436,7 +438,7 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
     void takePictureInternal(final ReadableMap options) {
         if (!isPictureCaptureInProgress.getAndSet(true)) {
 
-            if (options.hasKey("orientation")) {
+            if (options.hasKey("orientation") && options.getInt("orientation") != Constants.ORIENTATION_AUTO) {
                 mOrientation = options.getInt("orientation");
                 int rotation = orientationEnumToRotation(mOrientation);
                 mCameraParameters.setRotation(calcCameraRotation(rotation));
@@ -461,7 +463,7 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
                     }
 
                     mOrientation = Constants.ORIENTATION_AUTO;
-                    mCallback.onPictureTaken(data, displayOrientationToOrientationEnum(mDisplayOrientation));
+                    mCallback.onPictureTaken(data, displayOrientationToOrientationEnum(mDeviceOrientation));
                 }
             });
         }
@@ -503,9 +505,7 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
             return;
         }
         mDisplayOrientation = displayOrientation;
-        if (isCameraOpened() && mOrientation == Constants.ORIENTATION_AUTO) {
-            mCameraParameters.setRotation(calcCameraRotation(displayOrientation));
-            mCamera.setParameters(mCameraParameters);
+        if (isCameraOpened()) {
             final boolean needsToStopPreview = mShowingPreview && Build.VERSION.SDK_INT < 14;
             if (needsToStopPreview) {
                 mCamera.stopPreview();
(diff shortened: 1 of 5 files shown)


sibelius:
why do we need 3 variables to handle orientation?

n1ru4l:
@sibelius 

- `mScreenOrientation` is needed for correctly rendering the `CameraScene`(using `mDisplayOrientation` could distort/traverse the camera output visible to the user.

- `mDeviceOrientation` is used for getting the orientation of the video/image if `mOrientation` does not override it.

- `mOrientation` is set by the user if he passes the `orientation` option to `recordAsync`/`takePictureAsync` and overrides the orientation which would otherwise determined by `mDeviceOrientation`.

JulienMalige:
@n1ru4l I'm from the @JulienUsson's team.

On this branch, on **iOS**, I still have an issue. When I take a picture in **portrait**:
```
deviceOrientation:1  // OK
pictureOrientation:1 // OK
```
but my exif looks wrong (reversed):
```
...
PixelXDimension:4032
PixelYDimension:1864
...
```
When I save it to the Library, the orientation is wrong (portrait in landscape, landscape in portrait).

n1ru4l:
@FSarmento @matheuscorreia Would be awesome if you could also test the implementation

matheuscorreia:
Tested this on a Samsung SM-G930F (Android 8.0.0), and it presented some odd behaviours. These tests were made on picture mode only.

Back camera at 0° - Worked;
Back camera at 90° - Worked;
Back camera at 180° - Worked;
Back camera at 270° - Worked;

Front camera at 0° - Worked;
Front camera at 90° - Did not work; Picture turned out upside down;
Front camera at 180° - Worked;
Front camera at 270° - Did not work; Picture turned out upside down;

android/src/main/java/com/google/android/cameraview/Camera1.java
```diff
@@ -706,12 +718,12 @@ private int calcDisplayOrientation(int screenOrientationDegrees) {
...
      */
     private int calcCameraRotation(int screenOrientationDegrees) {
-        if (mCameraInfo.facing == Camera.CameraInfo.CAMERA_FACING_FRONT) {
-            return (mCameraInfo.orientation + screenOrientationDegrees) % 360;
-        } else {  // back-facing
-            final int landscapeFlip = isLandscape(screenOrientationDegrees) ? 180 : 0;
-            return (mCameraInfo.orientation + screenOrientationDegrees + landscapeFlip) % 360;
-        }
+       if (mCameraInfo.facing == Camera.CameraInfo.CAMERA_FACING_FRONT) {
+           return (mCameraInfo.orientation + screenOrientationDegrees) % 360;
```
sibelius: ?why front facing camera does not need landspaceFlip/

n1ru4l: Idk why but changing only applying it to the front facing instead of the back facing fixes this issue 🤣 At least on my Nexus 5X...

n1ru4l:
@matheuscorreia Could you please try again 😊

matheuscorreia:
Working perfectly now, @n1ru4l. Good job on this PR!

LGTM \o/

sibelius:
@n1ru4l great job on this

we've tested and it is working great

🚀

n1ru4l:
@Fsarmento did most of the work. Thank you :fire:

jgcmarins:
what is missing for merge?

Fsarmento:
You're welcome @n1ru4l, great work!

n1ru4l:
@Fsarmento I am totally for also adding `displayOrientation` to the possible options, maybe we should name it `useDisplayOrientation` do be more explicit?

Issue #1311, opened 2018-03-02
takePictureAsync never returns when children elements are updated


ospfranco:
*RNCamera*

### Steps to reproduce
1. <Camera> Component with children element inside
2. on takePictureAsync use setState to change one of the children element (ex. conditional rendering)
3. takePictureAsync will never return

### Expected behaviour
Picture should be taken normally and UI updated correctly

### Actual behaviour
takePictureAsync never returns

### Environment
- React-native 0.53
- Android (tested on android 7)

### react-native-camera
**1.0.1**

sibelius:
does this happens on expo camera module as well?

ospfranco:
hey sorry, missed the message, not really sure, didn't have time to test this on expo

pstanton:
#1373

hayo0914:
I'm using expo, and having same trouble.
Expo=v27.0.1
ReactNative=0.55.2

IssuehuntBot:
@0maxxam0 funded this issue with $20. [See it on IssueHunt](https://issuehunt.io/repos/33218414/issues/1311)

IssuehuntBot:
@thymikee has submitted a pull request. [See it on IssueHunt](https://issuehunt.io/repos/33218414/issues/1311)

IssuehuntBot:
@sibelius has rewarded $14.00 to @thymikee. [See it on IssueHunt](https://issuehunt.io/repos/33218414/issues/1311)

- :moneybag: Total deposit: $20.00
- :tada: Repository reward(20%): $4.00
- :wrench: Service fee(10%): $2.00

Issue #1862, opened 2018-10-17
Rendering in response to state changes causes camera to go black


petlys:
### Which implementation are you using

RNCamera

### Steps to reproduce

1. Open camera
2. Scan barcode which updates state
3. Conditionally render a component inside the camera view in response to the updated state

### Expected behaviour

Component should render as expected when state changes

### Actual behaviour

Only an issue on Android.

When `onBarCodeRead` is fired, I update a state which causes a component to render.
When that component renders, the camera screen goes black. 

```javascript
const styles = StyleSheet.create({
  container: {
    display: 'flex',
    flex: 1,
    justifyContent: 'center',
    flexDirection: 'column',
    alignItems: 'center'
  }
})

...

<RNCamera
  onBarCodeRead={() => this.setState({ isSuccess: true })}
  permissionDialogTitle="Permission to use camera"
  permissionDialogMessage="We need your permission to use your camera."
  style={styles.container}
>
  {this.state.isSuccess && <Text>Great success!</Text>}
</RNCamera>
```

### Environment

- React Native version: 0.48.1
- React Native platform + platform version: Android 7.0
- react-native-camera version: 1.3.0

Samsinite:
Getting this same error, on React 16.3.1, RN 0.55, android, and react-native-camera 1.3.0

Samsinite:
For some reason it seems to work if I have `flex: 0` styles on RNCamera initially. If I change it via the inspector after to `flex: 1` then it continues to work. Not sure what is going on...

petlys:
Thanks for sharing! I successfully worked around the issue by having `RNCamera` as its own component, and using absolute positioning to overlay it:

```javascript
<View style={{ flex: 1 }}
  <RnCamera style={{ flex: 1 }} />
  {this.state.isSuccess && <Text style={{ position: absolute; top: 0; left: 0; right: 0; bottom: 0 }}>Great success!</Text>}
</View>
```

`z-index` doesn't seem to work properly on Android, so you have to utilise the order of components to decide what goes on top of what.

Samsinite:
Thanks! I might try to help get this fixed this weekend and prefer to keep using facc for now as the app is paying attention to the status. Is there a way to grab the camera status if not using facc?

miracle2k:
I ran into the same problem, and struggled for quite some time with permissions (https://github.com/react-native-community/react-native-camera/issues/1752), which turned out to be a dead end.

[...]

PR #308, opened 2016-06-02
Patch Z-Order Android Problem

gchiocchio:
Fix #286 issue

PR #1995, opened 2018-12-13
fix(android): prevent shutting down camera when re-mounting children

thymikee:
Fixes https://github.com/react-native-community/react-native-camera/issues/1311

There's a bug where re-mounting children without transparent background shuts down the camera on Android (adb on LG G2 shows that `stopPreview` from `QCamera2HWI` gets called, but I couldn't set up debugging to see what's calling it, help appreciated).

By accident I've found a workaround, which involves wrapping children in a transparent view (not sure why `backgroundColor: 'transparent'` plays role here but it does).

As a manual test I've added a previously buggy behavior to example app – pressing zoom in / zoom out will mount/unmount a `Text` node which is a direct child of `RNCamera`.

### DIFF
--- a/android/src/main/java/org/reactnative/camera/RNCameraView.java
+++ b/android/src/main/java/org/reactnative/camera/RNCameraView.java
@@ -214,15 +214,6 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
     // React handles this for us, so we don't need to call super.requestLayout();
   }
 
-  @Override
-  public void onViewAdded(View child) {
-    if (this.getView() == child || this.getView() == null) return;
-    // remove and read view to make sure it is in the back.
-    // @TODO figure out why there was a z order issue in the first place and fix accordingly.
-    this.removeView(this.getView());
-    this.addView(this.getView(), 0);
-  }
-
   public void setBarCodeTypes(List<String> barCodeTypes) {
     mBarCodeTypes = barCodeTypes;
     initBarcodeReader();
--- a/examples/basic/App.js
+++ b/examples/basic/App.js
@@ -203,6 +203,9 @@ export default class CameraScreen extends React.Component {
             )}
           </TouchableOpacity>
         </View>
+        {this.state.zoom !== 0 && (
+          <Text style={[styles.flipText, styles.zoomText]}>Zoom: {this.state.zoom}</Text>
+        )}
         <View
           style={{
             flex: 0.1,
@@ -273,6 +276,12 @@ const styles = StyleSheet.create({
   flipText: {
     color: 'white',
     fontSize: 15,
+  },
+  zoomText: {
+    position: 'absolute',
+    bottom: 70,
+    zIndex: 2,
+    left: 2,
   },
   picButton: {
     backgroundColor: 'darkseagreen',
--- a/src/RNCamera.js
+++ b/src/RNCamera.js
@@ -28,9 +28,6 @@ const styles = StyleSheet.create({
     textAlign: 'center',
     fontSize: 16,
   },
-  cameraStopPreviewTempFix: {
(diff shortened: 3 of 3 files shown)


n1ru4l:
I think we should fix the root cause of this issue instead of some temporary hack

sibelius:
Does this happen on Expo câmera module as well?

I think we can use this hacky for now

sibelius:
this is the same code of expo camera https://github.com/expo/expo/blob/eafa96e0d6/packages/expo-camera/android/src/main/java/expo/modules/camera/ExpoCameraView.java#L162

sibelius:
I think we can try to solve this using react native lifecycles

https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/bridge/LifecycleEventListener.java#L27

onHostResume
onHostPause
onHostDestroy

my guess is that onPause or onResume is not being called properly by react-native

n1ru4l:
by adding this code to `onViewAdded` the issue is also "solved". Anyway why is the view removed an readded? When you start zooming the first time this behaviour occures. 0 -> 0.1. However on further renders this issue is not present anymore.

```java
   if (hasCameraPermissions()) {
      start();
    }
```

Maybe your solution is the right solution 😕

thymikee:
> Anyway why is the view removed an readded?

Without this method overrriden the React injected UI (children) flash for a split of second and once the component mounts, the camera is put on top. So the solution is to remove and add views with 0 index, because that's supposedly how you can be sure about the view (camera) can be on the very bottom, and React UI is on still on top. Not sure why this happens, maybe because of how the camera screen lays out? 
Here's reference to where it was added: https://github.com/react-native-community/react-native-camera/pull/308

> However on further renders this issue is not present anymore

The camera instance is still present and `isCameraOpen` reports false after the first children re-render (changing zoom from 0 to 0.1)

I'm not experience in Android dev, so it's all new to me and I'm just trying to get my head around it 😅

n1ru4l:
**react-native-linear-gradient** avoids passing in children into the NativeView

https://github.com/react-native-community/react-native-linear-gradient/blob/master/index.android.js#L84

https://github.com/react-native-community/react-native-linear-gradient/blob/master/index.ios.js

Other libraries from `react-native-community` also do not have this issue since they do not allow passing children to the "NativeView" and force users to utilize absolute positioning.

thymikee:
So, maybe we should follow this guidance? Should be possible to remove that hack from 2yrs ago altogether.

n1ru4l:
@thymikee I agree.

I would encourage an implementation like this one: https://github.com/react-native-community/react-native-linear-gradient/blob/master/index.android.js#L84

thymikee:
Gonna work on it over the weekend then! Thanks for helping out

n1ru4l:
@thymikee Thank you for getting this finally fixed 🎉

n1ru4l:
@sibelius Do you know any project that implemented passing children to the native component directly (without the issues we are experiencing)?

sibelius:
we can check on github https://github.com/search?l=Java&q=ViewGroupManager&type=Code

maybe someone in the react native core could help us on this

cc @axe-fb @cpojer

thymikee:
@n1ru4l addressed your feedback, so it still works well and we got rid of a hacky behavior on Android.

miracle2k:
The pull request will fix this, but for completeness sake, I want to point to https://github.com/react-native-community/react-native-camera/issues/1862#issuecomment-433497819 which I believe is the reason why the issue exists when children are added inside the native view.

n1ru4l:
@sibelius Do you want to investigate more or accept this solution? I am fine with this "workaround".

sibelius:
I’m fine with this workaround, we can accept a native solution later on if someone fixes it

PR #1996, opened 2018-12-13
fix(rn-camera): remove lodash dependency

### LINKED ISSUE #1993: remove lodash dependency
# Feature-Request

**Describe the Feature**
Remove `lodash` as a dependency. No need to pull in the whole library.

**Possible Implementations**
We can use `Object.entries(props).map()` instead. 

**Related Issues**
NONE

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate
> Want this feature to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

n1ru4l:
Closes #1993

### DIFF
--- a/src/RNCamera.js
+++ b/src/RNCamera.js
@@ -1,7 +1,6 @@
 // @flow
 import React from 'react';
 import PropTypes from 'prop-types';
-import { mapValues } from 'lodash';
 import {
   findNodeHandle,
   Platform,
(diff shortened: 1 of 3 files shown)


src/RNCamera.js
```diff
@@ -461,7 +468,7 @@ export default class Camera extends React.Component<PropsType, StateType> {
     }
   }
 
-  _convertNativeProps(props: PropsType) {
+  _convertNativeProps({ children, ...props }: PropsType) {
```
n1ru4l: @sibelius I excluded `children` from the props here? Will this result in a performance improvement? IDK if the children would be passed to the native instance as a string or something like that

sibelius: Not sure either

n1ru4l: I mean it wont hurt not passing it because it is not intended to be passed down 😉

PR #2003, opened 2018-12-15
Feature: add onStatusChange callback

ossareh:
Hey there,

I needed an evented way to be able to tell that access to the camera has been denied to prevent changing state during renders. This branch adds an optional callback `onStatusChange` which is called with an object: `{ status: CameraStatus }`.

Prior to this change the FaCC pattern was the only way to tell whether the user had denied access to the camera. In my case the photo frame is quite small (about 250 x 250) and I needed a way to update the screen around the photo frame (things like a header, providing the user some context, etc) if there had been a denial for permission. Doing so in the FaCC results in me calling `setState()` during `render()`.

I also took the opportunity to add a missing property to the takePhotoAsync() method in the typescript types definitions. Typically I'd have done that as a separate PR - but it was so small I opted to throw it into this one. LMK if you'd rather I broke it up into two PRs.

I haven't come across any guidance on how to submit a PR to this project, so I've gone with my best intuition. If I've missed some step or done something out of keeping with the governance of this project: a) sorry!, b) please feel free to point me in the right direction.


ossareh:
Will do. I'll get these landed in the next day or two.

@n1ru4l do you think the event should also fire on mount if permissions have been denied? I'm not entirely sure where to implement that just yet - but it turned up in testing on our side. I'd have thought that `onMountError()` would fire if, during mount, we detected that we did not have access.

n1ru4l:
@ossareh Could you please rebase on `master` branch? (https://gist.github.com/ravibhure/a7e0918ff4937c9ea1c456698dcd58aa)

ossareh:
@n1ru4l absolutely. I'll do that and get the doc change in today.

docs/RNCamera.md
```diff
@@ -305,6 +305,10 @@ Function to be called when native code emit onCameraReady event, when camera is
 
 Function to be called when native code emit onMountError event, when there is a problem mounting the camera.
 
+#### `onStatusChange`
+
+Function to be called when native code emits [CameraStatus](#status) events.
```
n1ru4l: Please document the arguments onStatusChange will receive

src/RNCamera.js
```diff
@@ -465,6 +467,12 @@ export default class Camera extends React.Component<PropsType, StateType> {
     }
   };
 
+  _onStatusChange = () => {
+    if (this.props.onStatusChange) {
+      this.props.onStatusChange({ status: this.getStatus() });
```
n1ru4l: can you please also add `recordAudioPermissionStatus : this.state.recordAudioPermissionStatus`?

n1ru4l:
@ossareh wrote:
> Ultimately once that is deprecated getStatus() could instead return a more complex type which encodes values for both permissions.

`getStatus` is currently only used internally and not documented for outside usage, so we can make changes to it. The recommended way to access the info is `onStatusChange` or the FaCC component approach.

@ossareh wrote:
> Another option would be to change isAuthorized to be more like recordAudioPermissionStatus; that is - instead of using a boolean to keep track of the status just use the enum value directly.

I definitely would prefer this instead of the boolean value.

Feel free to address this in a new pull request.

src/RNCamera.js
```diff
@@ -534,6 +542,14 @@ export default class Camera extends React.Component<PropsType, StateType> {
     return isAuthorized ? CameraStatus.READY : CameraStatus.NOT_AUTHORIZED;
   };
 
+  getRecordAudioPermissionStatus = (): RecordAudioPermissionStatus => {
```
n1ru4l: we do not need this method, the default value of `this.state.recordAudioPermissionStatus` `RecordAudioPermissionStatusEnum.PENDING_AUTHORIZATION`

src/RNCamera.js
```diff
@@ -465,6 +467,12 @@ export default class Camera extends React.Component<PropsType, StateType> {
     }
   };
 
+  _onStatusChange = () => {
+    if (this.props.onStatusChange) {
+      this.props.onStatusChange({ cameraStatus: this.getStatus(), recordAudioPermissionStatus: this.getRecordAudioPermissionStatus() });
```
n1ru4l:
```suggestion
      this.props.onStatusChange({ cameraStatus: this.getStatus(), recordAudioPermissionStatus: this.state.recordAudioPermissionStatus });
```

This should be fine!

ossareh:
@n1ru4l not sure if you saw this, if you did and you're busy then I totally get it - I just didn't want it forgotten about because I didn't @ you when I landed the changes :)

types/index.d.ts
```diff
@@ -141,6 +141,7 @@ export interface RNCameraProps {
   captureAudio?: boolean;
 
   onCameraReady?(): void;
+  onStatusChange?(event: { status: CameraStatus }): void;
```
n1ru4l: could you please also update this type definition?

n1ru4l: could you please also update this type definition?

ossareh:
I'm embarrassed that I missed that @n1ru4l  - thanks for catching it and your patience 👍🏼

Issue #1280, opened 2018-02-23
RNCamera ratio question


Jinran:
### Warning
RCTCamera is **DEPRECATED** on v1.0.0 follow our migration guide here https://github.com/react-native-community/react-native-camera/blob/master/docs/migration.md

### Which implementation are you using

RNCamera

### Expected behaviour
I need a camera view with changeless width & height,
eg: width: 200,height:100

### Actual behaviour
this is iphone7 device
![wechatimg20](https://user-images.githubusercontent.com/16407903/36573815-b1b0d2f0-187d-11e8-8437-d78cde3c309b.jpeg)
this is huawei P9 device
![wechatimg19](https://user-images.githubusercontent.com/16407903/36573844-cb07e2fc-187d-11e8-95c6-66c9f77e2125.jpeg)

why my camera streched like this and how can I set ratio to solve this problem in android with changeless width & height

### Environment
- **Node.js version**:
- **React Native version**:
- **React Native platform + platform version**: iOS 11.0+, Android 7.0, etc

### react-native-camera
**Version**: npm version or "master"

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate

Fconstant:
Android has a unique prop `ratio` which default value is `4:3`.
Try to get all supported ratios before, and then setting it to your desire.

Example:
```javascript
import React from 'react';
import { Platform } from 'react-native';
import { RNCamera } from 'react-native-camera';

const DESIRED_RATIO = "16:9";

class MyCameraComponent extends Component {
    state = {}

    prepareRatio = async () => {
        if (Platform.OS === 'android' && this.cam) {
             const ratios = await this.cam.getSupportedRatiosAsync();

             // See if the current device has your desired ratio, otherwise get the maximum supported one
             // Usually the last element of "ratios" is the maximum supported ratio
             const ratio = ratios.find((ratio) => ratio === DESIRED_RATIO) || ratios[ratios.length - 1];
             
             this.setState({ ratio });
        }
    }

    render() {
        return (
            <RNCamera
                ref={(cam) => this.cam = cam}
                onCameraReady={this.prepareRatio} // You can only get the supported ratios when the camera is mounted
                ratio={this.state.ratio}
            />
        );
}
```

Fconstant:
@Jinran Please, close this issue if this is not a problem anymore. Thanks :)

PR #2006, opened 2018-12-16
Fix getSupportedRatios is not a function

rhalff:
Function does not seem to be used in the example, yet I'm using it as a first step into trying this library out.

The Error Message was:
```
TypeError: undefined is not a function (evaluating 'this.camera.getSupportedRatios()')
```

The correct function seems to have Async appended as shown in https://github.com/react-native-community/react-native-camera/issues/1280#issuecomment-368644836

### DIFF
--- a/examples/basic/App.js
+++ b/examples/basic/App.js
@@ -42,7 +42,7 @@ export default class CameraScreen extends React.Component {
   };
 
   getRatios = async function() {
-    const ratios = await this.camera.getSupportedRatios();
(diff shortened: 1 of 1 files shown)


n1ru4l:
@rhalff Are you interested in implementing the actual feature into the example? Otherwise I think we should remove the unused code.

rhalff:
@n1ru4l no not really interested to implement the feature at the moment as I'm still getting familiar with the code.

I also didn't notice the example was initially committed just a few days ago, which makes this pull request a bit premature I guess.

PR #2007, opened 2018-12-16
Feature: isRecordingInterrupted video promise property

reime005:
I've added a simple flag to detect whether a video recording has been interrupted by minimizing the app. The property will be included in the promise return properties as documented.


sibelius:
@reime005 sorry for the delay

can you fix conflicts and we can merge it

PR #2014, opened 2018-12-18
Feature: Add method refreshAuthorizationStatus allowing to refresh authorization status of Camera

darekg11:
Hello,

First of all thanks for maintaining this library and effort that was put it into. We are using this library in one of our projects and at one pointed I have came across obstacle that seemed to me as impossible to resolve without forking.
After forking, I have added my own functionality and decided that it might be worth to share it, **please note that this PR might resolve a situation which can already be handled some other way but I have failed to come up with different solution - if you aware of any other possible fixes, please let me know.**

### What was happening and why I have thought that this change of code was necessary for me:
We had a view where we mounted `RNCamera` and children by using `FaCC` which worked great during typical happy path: `All permissions accepted by the user`.
After awhile we have begun testing all other paths, mostly the situation where things are happening in such order:  
1. User opens the view
2. No permissions have been granted yet
3. RNCamera is mounted for the first time, ask user for permissions
4. Decline permissions
5. Camera is showing `NotAuthorizedView` so it is all nice
6. Have a separate button that show Android Permissions Dialog once more, hoping that user will accept permissions this time around
7. User accepts permissions
8. Status of `RNCamera` returned in `FaCC` is still not `READY`

So at that point user eventually accepted the permissions but there was no way to make `RNCamera` notice that since `isAuthorized` is only set during `componentDidMount` life cycle method and since the view has not remounted that component then it won't spot the difference in permissions.
Of course closing the view and opening it once more, make recording work because `RNCamera` was mounted again.

At that point, I started to look through code to see if there is a way to update component `isAuthorized` - I haven't found a way except for remounting which could not work very well for us.
So I have added that new method - `refreshAuthorizationStatus` and called it after having permissions granted via:
`await this.camera.refreshAuthorizationStatus()`, `this.camera` being a `ref` set with:  
```
ref={cam => {
    this.camera = cam
}}
```
After that, `state` was updated therefore `Status` in `FaCC` was also correctly set and camera could start recording without having to remount it.

### DIFF
--- a/docs/RNCamera.md
+++ b/docs/RNCamera.md
@@ -528,6 +528,11 @@ The promise will be fulfilled with an object with some of the following properti
 
 - `iOS` `codec`: the codec of the recorded video. One of `RNCamera.Constants.VideoCodec`
 
+#### `refreshAuthorizationStatus: Promise<void>`
+
+Allows to make RNCamera check Permissions again and set status accordingly.  
+Making it possible to refresh status of RNCamera after user initially rejected the permissions.
+
 #### `stopRecording: void`
 
 Should be called after recordAsync() to make the promise be fulfilled and get the video uri.
--- a/src/RNCamera.js
+++ b/src/RNCamera.js
@@ -504,20 +504,43 @@ export default class Camera extends React.Component<PropsType, StateType> {
     this._isMounted = false;
   }
 
-  async componentDidMount() {
+  async arePermissionsGranted() {
     const { hasCameraPermissions, hasRecordAudioPermissions } = await requestPermissions(
       this.props.captureAudio,
       CameraManager,
       this.props.permissionDialogTitle,
       this.props.permissionDialogMessage,
     );
+    const recordAudioPermissionStatus = hasRecordAudioPermissions
+      ? RecordAudioPermissionStatusEnum.AUTHORIZED
+      : RecordAudioPermissionStatusEnum.NOT_AUTHORIZED;
+    return { hasCameraPermissions, recordAudioPermissionStatus };
+  }
+
+  async refreshAuthorizationStatus() {
+    const {
+      hasCameraPermissions,
+      recordAudioPermissionStatus,
+    } = await this.arePermissionsGranted();
     if (this._isMounted === false) {
       return;
     }
 
-    const recordAudioPermissionStatus = hasRecordAudioPermissions
-      ? RecordAudioPermissionStatusEnum.AUTHORIZED
-      : RecordAudioPermissionStatusEnum.NOT_AUTHORIZED;
+    this.setState({
+      isAuthorized: hasCameraPermissions,
+      isAuthorizationChecked: true,
+      recordAudioPermissionStatus,
+    });
+  }
+
+  async componentDidMount() {
+    const {
+      hasCameraPermissions,
+      recordAudioPermissionStatus,
+    } = await this.arePermissionsGranted();
+    if (this._isMounted === false) {
+      return;
+    }
 
     this.setState({
       isAuthorized: hasCameraPermissions,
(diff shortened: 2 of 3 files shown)


src/RNCamera.js
```diff
@@ -322,6 +322,22 @@ export default class Camera extends React.Component<PropsType, StateType> {
     return await CameraManager.getAvailablePictureSizes(this.props.ratio, this._cameraHandle);
   };
 
+  async refreshAuthorizationStatus() {
```
n1ru4l: can you please change `componentDidMount` to call this method? It would be better to not have duplicated code.

docs/RNCamera.md
```diff
@@ -513,6 +513,11 @@ The promise will be fulfilled with an object with some of the following properti
 
  - `iOS` `codec`: the codec of the recorded video. One of `RNCamera.Constants.VideoCodec`
 
+ #### `refreshAuthorizationStatus: void`
```
n1ru4l: this method returns a `Promise` that resolves to `void` (`Promise<void>`).

types/index.d.ts
```diff
@@ -284,6 +284,7 @@ export class RNCamera extends Component<RNCameraProps & ViewProperties> {
 
   takePictureAsync(options?: TakePictureOptions): Promise<TakePictureResponse>;
   recordAsync(options?: RecordOptions): Promise<RecordResponse>;
+  refreshAuthorizationStatus(): void;
```
n1ru4l: this method returns a `Promise` that resolves to `void` (`Promise<void>`).

darekg11:
@n1ru4l 
In case of Android devices it depends on few factors:
- Android devices before 6.0 displays the permissions before installing the app and user have to agree to all or reject all of them. In this case if user rejects them, permissions have to be adjusted via the settings
- Android devices after and 6.0 allow to show permission dialog before actual usage of resources making it possible to allow user to reject permissions but as developer you also might request them at any given time when you want to making it possible for user to eventually accept them later on
On top of that, on those devices user might also select `Never ask again` making this permissions rejected ultimately and therefore forcing user to change it via the Setting as in point 1)

In  case of IOS (not relevant to this PR but wort mentioning):
- It works similar to Android devices pre 6.0. App gets only one possibility to show the permissions dialog window and if user rejects, it has to be manually set via the Settings.

Thanks for review, I agree with all feedback. I wanted to reuse that function but for the time being I was more concerned that maybe You or some other maintainer know about some work around that wouldn't require this PR. But if that is not the case, I will update the code accordingly today in the evening or tomorrow.
Thanks.

darekg11:
@n1ru4l Heyo, I applied changes and rebased on the top of the `master`. Let me know if anything is missing and if you would like to have those commits squashed.

n1ru4l:
@darekg11 I will try to review this tomorrow!

darekg11:
Sure.

src/RNCamera.js
```diff
@@ -322,6 +322,27 @@ export default class Camera extends React.Component<PropsType, StateType> {
...
+      this.props.permissionDialogMessage,
+    );
+    return isAuthorized;
+  }
+
+  async refreshAuthorizationStatus() {
+    if (this._isMounted === false) {
+      return;
+    }
+    const isAuthorized = await this.areAndroidPermissionsGranted();
```
n1ru4l: you should move the is mounted check after the async call.

darekg11:
Hey @n1ru4l I have force-pushed cleaner 1 commit version. Let me know if this looks good for merging. I have tested this on Android device.

 
What is really interesting is that after recent changes done to master, Camera actually reacted to permissions changes by itself but only video worked and I got warning yellow box that I can't record Audio due to missing permissions. With a call to `refreshAuthorizationStatus` before actual recording, it all works together (both audio and video) 👌

darekg11:
Just wondering if you guys plan to merge this any time soon?

sibelius:
@n1ru4l  anything missing here?

darekg11:
Thank you very much 🙇

PR #2016, opened 2018-12-19
fix: RCTCameraManager crash on ios

TMomemt:
`AVMetadataMachineReadableCodeObject.stringValue` is a nullable property,  so it will cause a crash when it is set to the 'event' dictionary without nil detection.

### DIFF
--- a/ios/RCT/RCTCameraManager.m
+++ b/ios/RCT/RCTCameraManager.m
@@ -979,7 +979,7 @@ didFinishRecordingToOutputFileAtURL:(NSURL *)outputFileURL
 
   for (AVMetadataMachineReadableCodeObject *metadata in metadataObjects) {
     for (id barcodeType in self.barCodeTypes) {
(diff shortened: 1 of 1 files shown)


TMomemt:
It crushed when there is  no any barcode in camera,  but I'm sorry I can't reproduce this too..

This is our crush log on Fabric
image

TMomemt:
In IOS 11 Apple maybe improve the algorithm of recognition, causing a normal picture to be recognized as a barcode.

 So they add a `nullabel` to the `stringValue` property, see [this](http://codeworkshop.net/objc-diff/sdkdiffs/ios/11.0/AVFoundation.html).

n1ru4l:
Do you experience the sam issue when using `RNCamera`?

TMomemt:
> Do you experience the same issue when using `RNCamera`?

No,  we don't use `RNCamera` yet.

PR #2045, opened 2019-01-10
Fix the git tag format in the podspec

matt-oakes:
The format of the git tag which the podspec file specified did not match the ones which were being published.

Previously, it was using the raw version number from the `package.json`, however, the git tags had a `v` prefix.

This lead to errors saying the "branch could not be found" when attempting to use the library through Cocoapods.

This fix still reads the version number from `package.json`, however, it now prefixes it with a `v` to get the real git tag values. For example, `v1.6.4`.

### DIFF
--- a/react-native-camera.podspec
+++ b/react-native-camera.podspec
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
   s.license        = package['license']
   s.author         = package['author']
   s.homepage       = package['homepage']
-  s.source         = { :git => 'https://github.com/react-native-community/react-native-camera', :tag => s.version }
+  s.source         = { :git => 'https://github.com/react-native-community/react-native-camera', :tag => "v#{s.version}" }
 
   s.requires_arc   = true
   s.platform       = :ios, '9.0'


n1ru4l:
I have no problems using this library with cocoapods without the prefixed `v`. Could you please specify under which circumstances this issue occurs?

matt-oakes:
I've put a example project which shows the bug here: https://github.com/matt-oakes/RNCameraCocoapodsErrorExample/tree/6a73aba723dc5d9b768a64221b731169714034f7

I created this by:

1. Running `react-native init`
2. Converting the React dependencies to use Cocoapods and ensuring that it could build correctly (it did).
3. Adding the `react-native-camera` dependency as per the README.

When I run `pod install`, I get this error:

```sh
❯❯❯ bundle exec pod install
Analyzing dependencies
Fetching podspec for `DoubleConversion` from `../node_modules/react-native/third-party-podspecs/DoubleConversion.podspec`
Fetching podspec for `Folly` from `../node_modules/react-native/third-party-podspecs/Folly.podspec`
Fetching podspec for `React` from `../node_modules/react-native`
Fetching podspec for `glog` from `../node_modules/react-native/third-party-podspecs/glog.podspec`
Fetching podspec for `react-native-camera` from `../node_modules/react-native-camera`
Fetching podspec for `yoga` from `../node_modules/react-native/ReactCommon/yoga`
Downloading dependencies
Installing DoubleConversion (1.1.6)
Installing Folly (2016.10.31.00)
Installing React (0.57.8)
Installing boost-for-react-native (1.63.0)
Installing glog (0.3.5)
Installing react-native-camera (1.8.0)

[!] Error installing react-native-camera
[!] /usr/local/bin/git clone https://github.com/react-native-community/react-native-camera /var/folders/rh/zjc3x5b91k3gsw1bsl91nw200000gn/T/d20190111-7250-vqejge --template= --single-branch --depth 1 --branch 1.8.0

Cloning into '/var/folders/rh/zjc3x5b91k3gsw1bsl91nw200000gn/T/d20190111-7250-vqejge'...
warning: Could not find remote branch 1.8.0 to clone.
fatal: Remote branch 1.8.0 not found in upstream origin
```

As you can see, the error is because it's looking for a tag (branch) named `1.8.0` and can't find it, because this git repo has `v1.8.0` instead.

n1ru4l:
The following works for me:

```
pod 'react-native-camera', :path => '../node_modules/react-native-camera/react-native-camera.podspec'
```

Also you are using `:podspec` instead of `:path`. The readme states you should use `:path`.

matt-oakes:
@n1ru4l Ah! That will be it. I think I was following the examples from above in the podfile and missing the slight difference.

It still stands that the podspec is technically incorrect and this PR fixes that. I know it doesn't really matter if users follow the instructions, however, it might be best to correct the podspec anyway just incase.

n1ru4l:
Yeah you are right it is still incorrect. I am going to merge this anyway. I just wanted to understand how this is causing issues. Thank you for your quick responses 😊

matt-oakes:
No problem. Thanks for maintaining the library!

Issue #2019, opened 2018-12-19
Version 2 Discussion


n1ru4l:
# Version 2 Discussion

Version 2 should be all about removing deprecated dependencies and APIs and fixing bad APIs.

Roadmap Overview: https://github.com/react-native-community/react-native-camera/projects/1

Please link all crucial issues that should be part of v2 in this Thread.

n1ru4l:
@sibelius we can develop v2.0 in a new branch and use semantic-release to create v2.0 pre-releases (https://github.com/semantic-release/semantic-release/issues/563#issuecomment-448538232).

n1ru4l:
# Idea: Do not record audio by default

Recording audio requires additional permissions. A developer that only wants to scan a qr code does not require recording audio.

Therefore it would be best if by default audio recording is disabled.

Only if the developer added the `captureAudio` property to the component (and permissions are granted). Also I think we should differentiate between the both cases:

1. `Camera allowed` and `Audio not allowed` 
2. `Camera allowed` and `Audio allowed`

In the first case the camera view is still  functional and the developer should decide whether the app user is forced to also allow audio permissions or is allowed to record without video.

Related: #2041

sfgghh:
Leave

n1ru4l:
## Idea: Migrate to TypeScript and provide flow type definitions

TypeScript is generally more widespread than flow. There is currently no check that ensures the validity of flow type annotations of `react-native-camera`.

n1ru4l:
## Idea: Use depfu for handling dependencies

https://github.com/marketplace/depfu

n1ru4l:
## Idea: Remove prop-types
Having TypeScript and flow type definitions should me more than enough.

sraka1:
@n1ru4l https://github.com/react-native-community/react-native-camera/projects/1#card-15903168 Regarding "[Android] remove Camera1 API and replace it completely with Camera2 API" doesn't that bump the minSdkVersion to 21 (meaning Android 5.0 or higher)? More than 10% of global Android users are still on a version below 5.0.

Bardiamist:
### Feedback
My project required to scan qr-code. It was implemented by another developer with `react-native-camera@1.13.1`.

Today I tried to migrate to V2 but faced the following problems:
[...]

Issue #1956, opened 2018-11-28
Front camera shows black screen and crashes app on iOS


albullington:
# Bug Report
**To Do First**
- [ Y ] Did you try latest release?
- [ Y ] Did you try master?
- [ Y ] Did you look for existing matching issues?

**Related Modules**
RNCamera

**Platforms**
iOS

**Versions**
- Android: N/A
- iOS: 12, testing production build on an iPhone SE
- react-native-camera: git+https://github.com/react-native-community/react-native-camera.git#e18d52d2045f7db82c11368ff51cc3882d002e75
- react-native: 0.57.1
- react: 16.5.0
- react-navigation: ^2.17.0

**Description/Current Behaviour**
1. When toggling to the front RNCamera from the back RNCamera, the screen appears black. If I press the button to take photos, the app crashes. 
2. If I set the initial RNCamera type prop to "front," the app immediately crashes when rendering the camera screen. 

**Expected Behaviour**
User should be able to toggle between the front and back camera. When toggling to the front camera, users should be able to see the camera, not a black screen. Users should be able to take a photo while the front camera is toggled on. 

**Steps to Reproduce**
1. Create a production build on iOS using "react-native run-ios --configuration Release."
2. Toggle from the back camera to the front camera OR set RNCamera prop to "front."

**Additionals**
![img_1913](https://user-images.githubusercontent.com/35536439/49187502-89b87080-f31c-11e8-94f5-d5caf7dbecd1.PNG)
This is what displayed after I toggled to the front camera via the upper right button. 

```
const flashModeOrder = {
    off: "on",
    on: "off"
};

class CameraScreen extends Component {
  constructor( { navigation }: Props ) {
    super();

    this.state = {
      cameraType: "back",
      flash: "off",
      flashText: "OFF",
    };

    this.toggleCamera = this.toggleCamera.bind( this );
    this.toggleFlash = this.toggleFlash.bind( this );
  }

  takePicture = async () => {
    if ( this.camera ) {
      this.camera.takePictureAsync()
        .then( data => this.savePhotoToGallery( data ) )
        .catch( ( err ) => {
          this.setState( {
            error: err.message
          } );
        } );
    }
  }

  toggleFlash() {
    const { flash } = this.state;

    this.setState( {
      flash: flashModeOrder[flash],
      flashText: flashModeOrder[flash].toUpperCase()
    } );
  }

  toggleCamera() {
    const { cameraType } = this.state;

    this.setState( {
[...]

PR #2048, opened 2019-01-11
feat: allow camera scene when audio permissions are denied

### LINKED ISSUE #2047: Black screen at first render and null object reference  on android
# Bug Report
**To Do First**
- [ x ] Did you try latest release?
- [ x ] Did you try master?
- [ x ] Did you look for existing matching issues?

**Related Modules**

RNCamera

**Platforms**

Android

**Versions**

- Android: 6.0.1
- iOS:
- react-native-camera: 1.8.0
- react-native: 0.57.8
- react: 16.6.3

After creating a new app: react-native init appexample,
In the app folder, install the package: npm install react-native-camera --save,
And do automatic link: react-native link,
Add permission requirement to AndroidManifest.xml.

However, when running the app on device, I get a black screen instead of the camera preview.
Upon clicking on SNAP, I get:

"Attempt to invoke virtual method 'double java.lang.Double.doubleValue()' on a null object reference"
     getDouble
     ReadableNativeArray.java:114

**Expected Behaviour**

Be able to see camera preview and take picture.

**Steps to Reproduce**

As above.

**Additionals**

Environment (at work):
Linux Mint 19.1 - Tara
Node 11.3

Got same issue on other app, I have been reading and trying every possible solution and iteration for the last two days. One thing that kindda worked was using react-native-camera@1.2.0.

Just reproduced the issue at home, using:
Linux Mint 18.2 Cinnamon
Node 10.14.1

Downgrading to react-native-camera@1.2.0 solved the issue.

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate
> Want this issue to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

### LINKED ISSUE #2051: RNCamera **will** crash on iOS if all three Privacy keys are not included in Info.plist
# RNCamera crashes on iOS if all three Privacy keys are not included in Info.plist

Upon reading the README I came across:

```
<key>NSCameraUsageDescription</key>
<string>Your message to user when the camera is accessed for the first time</string>

<!-- Include this only if you are planning to use the camera roll -->
<key>NSPhotoLibraryUsageDescription</key>
<string>Your message to user when the photo library is accessed for the first time</string>

<!-- Include this only if you are planning to use the microphone for video recording -->
<key>NSMicrophoneUsageDescription</key>
<string>Your message to user when the microphone is accessed for the first time</string>
```

Naive me, I wasn't planning on doing any recording, so I didn't include `NSMicrophoneUsageDescription`. My app kept crashing on render and I went digging down a number of rabbit holes. Frustrated and I went back to square one and added all three, and boom, it works. 

I would **highly** recommend removing the comments that suggest the last two keys are optional. In my experience they most certainly were not optional.

n1ru4l:
[@n1ru4l wrote](https://github.com/react-native-community/react-native-camera/issues/2019#issuecomment-452688195):
> Furthermore I think we should differentiate between the both cases:
>
> - `Camera allowed` and `Audio not allowed`
> - `Camera allowed` and `Audio allowed`
>
> In the first case the camera view is still functional and the developer should decide whether the app user is forced to also allow audio permissions or is allowed to record without video.

**What this PR does:**

- Prevent the camera view from **not being rendered** if camera permissions are granted but audio permissions are not granted. 
- Prevent crashes when trying to record audio without audio permissions being granted.
- Show meanigful warnings for users that forgot to add permissions to `Info.plist`/`AndroidManifest.xml`

**How to test this?**

```
yarn add react-native-camera@https://github.com/n1ru4l/react-native-camera.git#d77c9c3caddf6ff0e900b1c9e70e3fd7d550e9e5
```

**TODO**:

- [x] Check if keys where defined in `Info.plist` before calling iOS API's (https://github.com/react-native-community/react-native-camera/issues/1956#issuecomment-453715321)
- [x] Check if permissions are defined in `AndroidManifest.xml` before asking (possible crash)
- [x] Add documentation for new FaCC property `recordAudioPermissionStatus`

**Affected Issues:**

Closes #2051
Closes #2047


src/RNCamera.js
```diff
@@ -399,6 +404,14 @@ export default class Camera extends React.Component<PropsType, StateType> {
       options.mute = true;
     }
 
+    if (
+      (options.mute || captureAudio) &&
+      recordAudioPermissionStatus !== RecordAudioPermissionStatusEnum.AUTHORIZED
+    ) {
+      // eslint-disable-next-line no-console
+      console.warn('Recording with audio not possible. Permissions are missing.');
```
n1ru4l: @sibelius Do you know a smart way to strip those warnings in the release bundle? Is there a way to mark them?

sibelius: if (__DEV__)

But you should strip all console log, there is a Babel plugin to do this

sibelius: check this https://www.npmjs.com/package/babel-plugin-transform-remove-console

n1ru4l: Shouldn't the metro bundler have a strip for `DEV` in release builds by default?

sibelius: not sure if __DEV__ behavior is well documented

n1ru4l:
@HighSoftWare96 What does the crashlog offer?

HighSoftWare96:
@n1ru4l 
I'm not able to install this anymore... the packager cannot find the module "./handlePermissions", something is wrong?

n1ru4l:
@HighSoftWare96 Seems like you should clear your react-native packager cache. YOu should start the package with `yarn react-native start --reset-cache`

HighSoftWare96:
@n1ru4l Right now I'm using npm but I've already tried with `npm start -- --reset-cache` 🙌

src/RNCamera.js
n1ru4l: There is a paragraph here: https://facebook.github.io/react-native/docs/performance#profiling

I wrapped all if statements that would result in `console.warn` with `if (__DEV__) {}`

ios/RN/RNCameraManager.m
```diff
@@ -371,11 +371,31 @@ + (NSDictionary *)faceDetectorConstants
 
 RCT_EXPORT_METHOD(checkVideoAuthorizationStatus:(RCTPromiseResolveBlock)resolve
                   reject:(__unused RCTPromiseRejectBlock)reject) {
-    __block NSString *mediaType = AVMediaTypeVideo;
-    
-    [AVCaptureDevice requestAccessForMediaType:mediaType completionHandler:^(BOOL granted) {
-        resolve(@(granted));
-    }];
+    if ([[NSBundle mainBundle].infoDictionary objectForKey:@"NSCameraUsageDescription"] != nil) {
```
n1ru4l: @sibelius What is your opinion on this change?

I noticed that many users report app crashes and have issues with providing xcode logs and similar.

IMHO this approach is more helpful for new users as it shows a warning with whats wrong instead of crashing the app.

sibelius: I prefer this approach

Better to provide a fix path for users

n1ru4l:
@sibelius wrote
> Why remove common handle permission function?

Because they now differ as the `RNCamera` implementation returns an object with two boolean properties instead of one boolean value

ios/RN/RNCameraManager.m
n1ru4l: I try to do the same for android once I have access to an android device and can test it

n1ru4l:
@sibelius Do you know why on my Nexus 5X with Android `8.1.0` there is no Dialog being shown for Permissions when running the example app?

Also I do not need to add `<uses-permission android:name="android.permission.CAMERA"/>` to the `AndroidManifest.xml` (which is not in the example app anyway). However the camera view still works.

I only have to add `<uses-permission android:name="android.permission.RECORD_AUDIO"/>` otherwise the call in the `requestPermissions` function will resolve with `NEVER_ASK_AGAIN`. However after adding it, it will instantly resolve with `GRANTED` instead of showing a dialog to the user.

n1ru4l:
Ok I understood it partially. The `AndroidManifest.xml` located in `react-native-camera/android/src/main/AndroidManifest.xml` is "merged" with the one of the `SampleApp` upon building, which explains why I do not have to specify `<uses-permission android:name="android.permission.CAMERA"/>` in the `AndroidManifest.xml` of the consumer of `react-native-camera`.

However I still do not understand why I do not see a dialog...

android/src/main/java/org/reactnative/camera/CameraModule.java
```diff
@@ -369,4 +373,22 @@ public void execute(NativeViewHierarchyManager nativeViewHierarchyManager) {
             }
         });
     }
+
+    @ReactMethod
+    public void checkIfRecordAudioPermissionsAreDefined(final Promise promise) {
```
n1ru4l: I added a new method similar to the one on ios to check if the key is defined in the `AndroidManifest.xml` before asking for the permissions.

luissmg:
@asanz8 Thanks for testing! I will not test it then @n1ru4l

src/RNCamera.js
```diff
@@ -11,11 +11,61 @@ import {
...
+  }
+
+  if (Platform.OS === 'ios') {
+    hasCameraPermissions = await CameraManager.checkVideoAuthorizationStatus();
+  } else if (Platform.OS === 'android') {
+    const cameraPermissionResult = await PermissionsAndroid.request(
+      PermissionsAndroid.PERMISSIONS.CAMERA,
+      params,
+    );
+    hasCameraPermissions = cameraPermissionResult === PermissionsAndroid.RESULTS.GRANTED;
```
ChangeSomeCode: Hello! Why you use different "requestPermissions" method for Camera.js and RNCamera.js? Can you add `|| cameraPermissionResult === true` like in Camera.js? Because react-native return boolean-type result for devices before Android 6.0

thx!

PR #2050, opened 2019-01-11
fix(android): use the correct camera orientation to fix image detection

### LINKED ISSUE #1714: [Android] text recognition fails on Portrait mode

RNCamera 1.1.4

### Steps to reproduce
Text Recognition is not working when in Portrait Mode. I tested with a blank page with the following two lines: 
**Hello World**
**Hello World Again**
And it only detects Single or double characters on each event, like 'O', 'IC'. 

When I rotate to landscape mode, it works perfectly. 

Heres is how I handle the event and render it:
`onTextRecognized = ({textBlocks}) => this.setState({ detectedTexts: textBlocks.map(b => b.value) })`
```
renderDetectedText() {
    return (
      <View style={[styles.facesContainer,{left: 10, top:"50%"}]}>
        <Text style={styles.flipText}>{this.state.detectedTexts.join("\n")}</Text>
      </View>
    )
  }
```
The RNCamera element takes the full screen.

### Environment
- **React Native version**: 0.55.4
- **React Native platform + platform version**: Installed on a physical Android 7.0 Device

### LINKED ISSUE #2046: [Android] `useCamera2API` and barcode scanning isn't working
# Bug Report
**To Do First**
- [x] Did you try master?

**Related Modules**

**Platforms**
Android

**Versions**
- Android:
  * Nexus 5X @ Android 8.1.0
  * Samsung S6 Edge(SMG-925A)  @ Android 7.0
- react-native-camera: 1.8.0
- react-native: 16.6.3
- react:  0.57.8

**Description/Current Behavior**
Camera is setup as
```
      <RNCamera
        ref={ref => {
          this.camera = ref;
        }}
        style={{
          flex: 1,

        }}
        useCamera2Api={true}
        permissionDialogTitle={'Permission to use camera'}
        permissionDialogMessage={'We need your permission to use your camera phone'}
        onBarCodeRead={this.onBarcode}
     / >
```

**
When pointing the camera at the barcode with the `Camera2API`  enabled there are no detected barcodes.

**Expected Behaviour**
The barcodes should be detected.

**Steps to Reproduce**
1. Set a `onBarCodeRead` to a function
2. Set `useCamera2Api={true}

**Additionals**
I have spun out  the example camera app  into a new example repo with a project created by  `react-native init`. 

https://github.com/vinayan3/CameraTest

For ease of reproducing this bug I've added in buttons which let you toggle which `CameraApi` is used. When you toggle the API or the barcode detection the camera will be unmounted, 5 second wait, and remounted.  Here is a screenshot of the toggles I put in:

screenshot_20190110-115742

matt-oakes:
### The issue
[This code](https://github.com/react-native-community/react-native-camera/blob/master/android/src/main/java/org/reactnative/camera/RNCameraViewHelper.java#L278-L284) in the library is taking in the `rotation` from the `CameraView` and converting it to the image rotation. It is, however, making some assumptions about the camera orientation which is why it is failing on some devices.

The `rotation` from the `CameraView` callback corresponds to the display orientation. Portrait is *usually* `0`, however, that's device dependent. The camera on the device also has an orientation relative to the display orientation. You can read about it [here](https://www.captechconsulting.com/blogs/android-camera-orientation-made-simple).

### The fix
* Update CameraView library to allow us to get the orientation of the camera for both the deprecated camera API ([here](https://developer.android.com/reference/android/hardware/Camera.CameraInfo.html#orientation)) and the new one ([here](https://developer.android.com/reference/android/hardware/camera2/CameraCharacteristics.html#SENSOR_ORIENTATION)).
* Update the calculation in the library to use this orientation correctly

Fixes #1714
Should fix #2046 (untested)

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -500,6 +500,11 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
     }
 
     @Override
+    int getCameraOrientation() {
+        return mCameraInfo.orientation;
+    }
+
+    @Override
     void setDisplayOrientation(int displayOrientation) {
         if (mDisplayOrientation == displayOrientation) {
             return;
--- a/android/src/main/java/com/google/android/cameraview/Camera2.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera2.java
@@ -231,6 +231,8 @@ class Camera2 extends CameraViewImpl implements MediaRecorder.OnInfoListener, Me
 
     private int mFlash;
 
+    private int mCameraOrientation;
+
     private int mDisplayOrientation;
 
     private int mDeviceOrientation;
@@ -615,6 +617,11 @@ class Camera2 extends CameraViewImpl implements MediaRecorder.OnInfoListener, Me
     }
 
     @Override
+    int getCameraOrientation() {
+        return mCameraOrientation;
+    }
+
+    @Override
     void setDisplayOrientation(int displayOrientation) {
         mDisplayOrientation = displayOrientation;
         mPreview.setDisplayOrientation(mDisplayOrientation);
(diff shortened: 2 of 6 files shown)


n1ru4l:
@matt-oakes wrote
> The rotation from the CameraView callback corresponds to the display orientation. Portrait is usually 0, however, that's device dependent. The camera on the device also has an orientation relative to the display orientation.

Probably unrelated but could you please check if those constants are correct for all devices? https://github.com/react-native-community/react-native-camera/blob/ed4f5bf9182e568dba75872b9cf0e046438589ef/android/src/main/java/org/reactnative/camera/Constants.java#L43-L47

https://github.com/react-native-community/react-native-camera/blob/401c485db342fe6c7bd8580057c82b5ff412170d/android/src/main/java/com/google/android/cameraview/Constants.java#L42-L46

matt-oakes:
@n1ru4l I believe those are just used inside the camera library to decide how much rotation to apply. For example here: https://github.com/react-native-community/react-native-camera/blob/513bfe860b614ee827b2999ae09a8b2558c4253b/android/src/main/java/com/google/android/cameraview/Camera1.java#L408-L436

These will be the same for all devices as these angles are the same as the display rotation which is standard across all devices.

Chimaa123:
This change breaks face detection.

telmen:
Yes. After reverting this change, face detection works fine.

Issue #2041, opened 2019-01-09
"camera not authorized" in version 1.8.0 (Android)


matheusmonte:
# Bug Report
**To Do First**
- [X ] Did you try latest release?
- [ X] Did you try master?
- [ X] Did you look for existing matching issues?

**Related Modules**
RNCamera

**Platforms**
Android (Not tested in iOs)

**Versions**

- Android: 8.1
- iOS: not tested
- react-native-camera: 1.8.0
- react-native: "0.57.5"
- react: 16.6.1

**Description/Current Behaviour**
npm install in react-native-camera and try to use. The app will request camera permission, you tap "OK", the components show "camera not authorized"

**Expected Behaviour**
Show camera frame after get permission

**Steps to Reproduce**
Use RNCamera,
first app time, try to get permissions
try to use after this

**Additionals**
On system settings, show my app as permited
> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate
> Want this issue to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

renalpraba:
![photo_2019-01-09_11-53-00](https://user-images.githubusercontent.com/39117076/50877606-29314e80-1405-11e9-892c-fe0eacb87fc6.jpg)

Found this using v1.8.0 , v1.7.0 & 1.6.6
No problem in v1.6.5

syadykin:
Same issue, 1.6.1 works fine.

adimshev:
Same issue

renalpraba:
@n1ru4l I'm using [react-native-qrcode-scanner](https://github.com/moaazsidat/react-native-qrcode-scanner) (require react-native-camera) with its sample setup getting this error when using react-native-camera 1.6.6 or later, no problem on 1.6.5

adimshev:
```js
<RNCamera
	style={styles.camera}
	onCameraReady={this.onCameraReady}
	onMountError={this.onMountError}
	onBarCodeRead={
		waiter /*|| alert*/ ? undefined : this.onBarCodeRead
	}
	pendingAuthorizationView={null}
/>
```

adimshev:
Permissions are automatically requested when first rendering the RNCamera.

n1ru4l:
@adimshev Are both `Camera` and `Audio` recording permissions permitted?

adimshev:
I add into Manifest only Camera permissions (for barcode scanner).

[...]

PR #2052, opened 2019-01-14
Update RNCamera.md with NOT_AUTHORIZED problem solve

matheusmonte:
Add props captureAudio={false} in the sample and a desclaime in Status area to help people to solve NOT_AUTHORIZED message after give permission to system in version 1.8.0


matheusmonte:
Update documentation to avoid issue #2041

n1ru4l:
I am currently working on making microphone permissions optional: #2048

Would be awesome if you could test if this solves the problem.

n1ru4l:
This issue was resolved in #2048

You can now have insufficient `RECORD_AUDIO` permissions but still record videos. If there is still something missing in the documentation please open a new issue/pr.

PR #2055, opened 2019-01-16
feat(android): add videoBitrate option for recordAsync

andrewschenk-linx:
While VideoQuality can be set, this option does not give developers anyway to fine-tune the resulting file size of a recorded video.  720p recording will use a 12 Mbps bitrate on a Galaxy S8+, even though 720p should be closer to 5 Mbps.

Introducing a new optional recording option helps to solve this problem.
targetBitrate takes an integer value (e.g. 1000 x 1000 x 5 aka 5Mbps).

Example of calling it from React Native

```
const recordingOptions = {
          maxDuration: 60*20,
          mute: false,
          quality: RNCamera.Constants.VideoQuality['720p'],
          targetBitrate: 1000*1000*5 // 5 Mbps
};
const data = await this.camera.recordAsync(recordingOptions);
```

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -853,11 +853,14 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
         mMediaRecorder.setOutputFile(path);
         mVideoPath = path;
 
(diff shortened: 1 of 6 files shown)


andrewschenk-linx:
It falls back to default behavior, as if this code was never implemented.

andrewschenk-linx:
Added documentation and types.

PR #2056, opened 2019-01-17
wrap captureStillImageAsynchronouslyFromConnection with try/catch

KryDos:
hi everyone, 
my app is crashing from time to time and I've noticed it is happening due to unhandled exception raised somewhere inside of `captureStillImageAsynchronouslyFromConnection` method. 

The only solution I could come up with is simple try/catch block where catch is calling `reject` function so the code execution passed further to JavaScript level and can be handled there. 

The PR diff can look pretty big but the only change is try/catch block as I've mentioned. 

The issue can be found here - https://github.com/react-native-community/react-native-camera/issues/1956
You can find crash log posted by @gaurav321garg, which is very similar to crash log I had. 

With this fix app is not crashing BUT I'm now getting Black Screen instead of actual stream from camera for some reason (the frequency of this black screen is the same as frequency of crash that I had).

Let me know if I can do anything to improve this PR.

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -378,93 +378,102 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
 
     AVCaptureConnection *connection = [self.stillImageOutput connectionWithMediaType:AVMediaTypeVideo];
     [connection setVideoOrientation:orientation];
-    [self.stillImageOutput captureStillImageAsynchronouslyFromConnection:connection completionHandler: ^(CMSampleBufferRef imageSampleBuffer, NSError *error) {
-        if (imageSampleBuffer && !error) {
-            if ([options[@"pauseAfterCapture"] boolValue]) {
-                [[self.previewLayer connection] setEnabled:NO];
-            }
-
-            BOOL useFastMode = [options valueForKey:@"fastMode"] != nil && [options[@"fastMode"] boolValue];
-            if (useFastMode) {
-                resolve(nil);
-            }
-            NSData *imageData = [AVCaptureStillImageOutput jpegStillImageNSDataRepresentation:imageSampleBuffer];
-
-            UIImage *takenImage = [UIImage imageWithData:imageData];
-
-            CGImageRef takenCGImage = takenImage.CGImage;
-            CGSize previewSize;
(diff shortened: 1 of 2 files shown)


sibelius:
why does this happens?

KryDos:
@sibelius in my case it's pretty easy to reproduce if I'm just spamming (quickly pressing) the "take picture" button (the button is taking picture obviously). This button is turning off when it pressed first time, but if I'm fast enough I can press it twice. 

Also before "take picture" button I have some kind of "Show camera" button. In case I press "Show camera" and then quickly press "take picture" buttons then it's very good chance to get the crash. 

Unfortunately I don't know the actual reason of why is this happening. In "user-land" my users usually do not spam the "take picture" button. But sometimes they still getting crash for some reason. There is no logs I can get from that crashes unfortunately.

sibelius:
@n1ru4l what do you think about this?

KryDos:
I'm sorry guys, I understand I have to provide you with reproducible example but I can't make it (can't even start).

It's probably related to the way how this library is used in my application and the error is not in the library itself. But at some point the exception is thrown from inside the library and I can't handle it on JS level, so that's why I had to add that try/catch part (without deep understanding what is actually going on). 

Today I found that sometimes `_cameraHandle` is null and the whole app is crashing again because it cannot be null. So I've added another check today.

sibelius:
why `_cameraHandle` can be null?

could you open an issue on expo camera, so they can help us on this one?

what happens when you throw when cameraHandle is null?

KryDos:
@sibelius when I throw the exception (from JS file) when cameraHandle is null it is handled by my code and just shows an alert message. 

Why `_cameraHandle` can be null mystery for me. Is cameraHandle referencing the RNCamera component, right? 

About the expo camera, could you please help to understand what exact issues I have to open there. I don't really understand how Expo is related.

sibelius:
this module is based on expo camera module, they are very close to each other

sibelius:
can you rebase the last time, so we can merge it?

KryDos:
@sibelius sure, it's rebased. 

Let me know if I can do anything else.

sibelius:
I'd like to use try catch as the last resource, much camera is hard so let's do this

tks

PR #2068, opened 2019-01-24
upgrading to sdk 28 and androidx

gla5001:
`AndroidX` is the replacement for Google Support Libraries. Android P is the transition phase for this. The next version of Android is likely to support only AndroidX. This library fails on android builds that use the new compileSdkVersion. This PR makes the transition to AndroidX.

The only changes in this PR are namespaces, package names in import statements, and Gradle dependencies.

The changes in this PR were auto-generated by Android Studio using instructions from [here](https://developer.android.com/jetpack/androidx/migrate).


gla5001:
Build failure since new licences have not been accepted.

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'android'.
> Failed to install the following Android SDK packages as some licences have not been accepted.
     build-tools;28.0.3 Android SDK Build-Tools 28.0.3
     platforms;android-28 Android SDK Platform 28
  To build this project, accept the SDK license agreements and install the missing components using the Android Studio SDK Manager.
  Alternatively, to transfer the license agreements from one workstation to another, see http://d.android.com/r/studio-ui/export-licenses.html
  
  Using Android SDK: /opt/android-sdk-linux
```

sibelius:
Fix is coming

gla5001:
@sibelius Fix is coming for the build failure or for the AndroidX support? Do I need to do anything?

sibelius:
use

- image: entria/react-native-android:0.1.86

on config.yml of circleci

sibelius:
try to run `./gradlew check` to understand why it is not working

sibelius:
can you try to run basic example and see if it will work?

gla5001:
sure

gla5001:
@sibelius the example seems to work after the upgrade. I did need to make a bundle via `react-native bundle --platform android --dev false --entry-file index.js --bundle-output android/app/src/main/assets/index.android.bundle --assets-dest android/app/src/main/res` first.

see below

example

android/build.gradle
```diff
@@ -51,7 +46,7 @@ dependencies {
   implementation "com.google.zxing:core:3.3.0"
   implementation "com.drewnoakes:metadata-extractor:2.9.1"
   implementation "com.google.android.gms:play-services-vision:${safeExtGet('googlePlayServicesVersion', '15.0.2')}"
-  implementation "com.android.support:exifinterface:${safeExtGet('supportLibVersion', '27.1.0')}"
-  implementation "com.android.support:support-annotations:${safeExtGet('supportLibVersion', '27.1.0')}"
-  implementation "com.android.support:support-v4:${safeExtGet('supportLibVersion', '27.1.0')}"
+  implementation "androidx.exifinterface:exifinterface:1.0.0-beta01"
```
n1ru4l: why is this beta? is it not stable yet?

gla5001: Good catch. Thats what the Android Studio migration tool migrated the dependency to. Looking at [this](https://developer.android.com/jetpack/androidx/migrate), it looks like the stable release versions are out. I'll update with another commit.

gla5001:
@n1ru4l Unfortunately, this will break existing android users if they upgrade. They will need to migrate their projects to use compileSdk 28, build tools 28, update dependencies in `app/build.gradle` if they are pointing to the old support library versions, and use androidx by setting 

```
android.enableJetifier=true
android.useAndroidX=true 
```

in the gradle.properties file in their project. FWIW they will eventually have to upgrade their projects since the Android Support lib is now depreciated in favor of androidx.

Does that answer your questions? How would you like me to handle this?

sibelius:
can we add a .md file inside docs/ with a migration path to android 28 and androidx ?

gla5001:
sure. I can do that

kassim:
the reason why a migration is necessary is because of the nature of React Native modules being a distribution of the source code 

you could avoid the necessity of the migration if the relevant Android code was distributed in a packaged (aar or jar) form - as the Jettifier modifies such dependencies to AndroidX package naming

if you see a project such as https://github.com/react-native-community/react-native-text-input-mask - where the native implementation is actually held in a different project but is just set as a dependency - this kind of implementation prevents the need of migration between Support and AndroidX package naming

sibelius:
@kassim so what are your recommendation?

gla5001:
@kassim wouldn't the `react-native-camera-android` dep eventually need to be upgraded anyways if new features from exifinterface or annotation wanted to be used? Since the support library is deprecated all new dev is happening in the AndroidX packages?

I'll let @sibelius and @n1ru4l weigh in but maybe thats worth another PR to make it more flexible

kassim:
So of options:
1. as this PR does, force users to update their projects to AndroidX to be able to use React Native Camera - works for me but this lib has thousands of users and some of whom probably won't be happy
2. maintain 2 branches, one AndroidX and one Support packages - not so fun for the one who must maintain, thought its what some Android libraries do
3. a separate repository for Android code, have it dependent on the old Support package and have this project set it as a dependency (using Jitpack perhaps?) - negative is as @gla5001 says, you won't have the latest features at some point, also you're maintaining 2 repositories for 1 project which isn't nice either


perhaps option 2 then option 1 after a reasonable amount of time has passed

gla5001:
also, in regards to option 1. They potentially wont be happy if they upgrade, otherwise they should be fine.

darekg11:
Just opinion from someone who relies heavily on this library in production project:  

Maintaining 2 branches for now seems like much more reasonable solution for people like me who are already using this lib and would like to get most recent versions and at the same time make sure that other 3rd party libs will work correctly (not sure if upgrading to compileSDK 28 and AndroidX wouldn't break any other dependency).

At the same time, I know that this will be a pain in the ass for maintainers.

n1ru4l:
Maintaining two branches is definitely not a solution.

What is the state of using `AndroidX` for other packages? Does react-native use it?

gla5001:
I do not believe react-native uses it as of it, but its coming see https://github.com/facebook/react-native/issues/23112

sibelius:
can we wait until androidX lands on react-native core?

docs/AndroidXMigration.md
```diff
@@ -0,0 +1,101 @@
...
+## Migration via Android studio
+
+You can follow the instructions [here](https://developer.android.com/jetpack/androidx/migrate) to migrate you existing
+project.
+
+## Manual
+
+#### Update the android section `android/app/build.gradle` to
+
+```javascript
```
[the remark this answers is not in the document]
gla5001: ill update it

ankitbansal25:
Hi,
I am using this library in my project. I am also facing the same issue. 
When we can expect these changes to be merged.

Thanks and Regards,
Ankit

hsjoberg:
@sibelius AndroidX support has landed in 0.60.0-rc.0, which breaks react-native-camera.
Perhaps a good idea would be to move forward with this PR.

sibelius:
Good, let’s fix the conflicts and ship it

gla5001:
I was able to get a rebase in and resolve all the conflict. The build now passes. 

I was also able to do a little bit of testing on the example/basic app and everything seems to function as it should. I'm pretty tied up for the weekend and all of this coming week, so if somebody wants to test a little more extensively, feel free too.

banshee:
Personally, I would have liked the major version to change; a change like this that breaks builds for existing projects is usually a major version bump, not a minor.  So 3.0.0, not 2.10.

mikehardy:
A change like this looks like a clear major bump to me

I would integrate this script https://gist.github.com/rozPierog/9c1f63d39c4e5cece537e5b03d0f43f1 from https://github.com/react-native-community/discussions-and-proposals/issues/129#issuecomment-498776701 here in the library, and in the breaking-change documentation mention people not converted to AndroidX should use the react-native-camera-support.aar in their gradle dependency. 

Not sure *exactly* how they need to refer to it...but you already landed your AndroidX conversion so you're motivated to figure it out I think :-) - should be easy?

Note that I have worked on that issue, proposed that AAR de-jetified style, and reviewed that script, and my day-job app uses react-native-camera. So I will happily test integrate it following the proposed documentation but I don't have time to make the PR here sorry...

sibelius:
create an issue here https://github.com/react-native-community/react-native-camera/issues/2296, so more people can collaborate in this matter

and then we can do a major bump

PR #2078, opened 2019-01-29
Update `captureAudio` default value in docs.

loris42:
The current docs says the default `captureAudio` value is `true` whereas it is actually `false`.
See [https://github.com/react-native-community/react-native-camera/blob/master/src/Camera.js#L166](https://github.com/react-native-community/react-native-camera/blob/master/src/Camera.js#L166)


n1ru4l:
@loris42 On `RNCamera` the `captureAudio` property has the default value `false`. https://github.com/react-native-community/react-native-camera/blob/57bc327d847ccb7ea90ded51a5a82168388b1349/src/RNCamera.js#L324

PR #2087, opened 2019-02-02
[Android] Expose rawValue from google vision barcode scanner

gaykov:
Hello,

I've faced a pretty interesting issue with Google Vision scanner. When scanning a barcode, that contains, for example, VCard (contact) information in the following format:
```
BEGIN:VCARD
FN:Robert Pau Shou Chang
N:Pau;Shou Chang;Robert
SORT-STRING:Pau
END:VCARD
```
`onGoogleVisionBarcodesDetected` will receive the following object:
```
{
  barcodes: [
    {
      data: "Robert Pau Shou Chang",
      type: "QR_CODE"
  ],
  ...
}
```

The reason is that currently, `data` field is populated from `displayValue`, which is "user friendly code value", as stated here: https://developers.google.com/android/reference/com/google/android/gms/vision/barcode/Barcode

I think that in many cases users will expect to get a raw scanned value and now it's impossible to get it from Google Vision.

The solution is to get `data` from `rawValue` field, which is "Barcode value as it was encoded in the barcode." from documentation.

But this change will breack some existent code that relies on data, so here I've added rawValue as an additional parameter to the codes object.

Please let me know you throuhgts on it. If it's ok, I'll add documentation to this PR as well.

Thanks!

### DIFF
--- a/android/src/main/java/org/reactnative/camera/events/BarcodesDetectedEvent.java
+++ b/android/src/main/java/org/reactnative/camera/events/BarcodesDetectedEvent.java
@@ -72,6 +72,7 @@ public class BarcodesDetectedEvent extends Event<BarcodesDetectedEvent> {
       Barcode barcode = mBarcodes.valueAt(i);
       WritableMap serializedBarcode = Arguments.createMap();
       serializedBarcode.putString("data", barcode.displayValue);
+      serializedBarcode.putString("rawValue", barcode.rawValue);
       serializedBarcode.putString("type", BarcodeFormatUtils.get(barcode.format));
       barcodesList.pushMap(serializedBarcode);
     }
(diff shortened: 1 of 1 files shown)


android/src/main/java/org/reactnative/camera/events/BarcodesDetectedEvent.java
```diff
@@ -72,6 +72,7 @@ private WritableMap serializeEventData() {
       Barcode barcode = mBarcodes.valueAt(i);
       WritableMap serializedBarcode = Arguments.createMap();
       serializedBarcode.putString("data", barcode.displayValue);
+      serializedBarcode.putString("rawValue", barcode.rawValue);
```
n1ru4l: This is already implemented on iOS. 
Check out: https://github.com/react-native-community/react-native-camera/blob/22533ed8e8c47f17ed7ca28f5879064fc2572a7f/ios/RN/RNCamera.m#L884

Please use the same name `rawData` instead of `rawValue`. What happens if the barcode does not contain a raw value? Will putString result in `rawData` being `null` or will the application crash?

cozmo: Also implemented on android already (I think?) - https://github.com/Clever/react-native-camera/blob/a3afef1ab25da9337f3d9cf0ef9259063348b6e5/android/src/main/java/org/reactnative/camera/events/BarCodeReadEvent.java#L77

It's not immediately clear to me what the difference between thee 2 files are, but we should make sure value returned is the same in each

n1ru4l:
@gaykov As @cozmo pointed out this is already possible.

sibelius:
@gaykov  can you send a pr updating typescript def?

PR #2095, opened 2019-02-06
[Android] Expose barcode bounds and source dimension from google vision barcode scanner

shashankkushwah:
The current google vision barcode scanning is only exposing barcode data and type. I added the bounds and source dimensions in the scan event.

Also, implemented barcode scanning in basic example.

### DIFF
--- a/android/src/main/java/org/reactnative/camera/RNCameraView.java
+++ b/android/src/main/java/org/reactnative/camera/RNCameraView.java
@@ -432,7 +432,7 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
 
     SparseArray<Barcode> barcodesDetected = barcodesReported == null ? new SparseArray<Barcode>() : barcodesReported;
 
-    RNCameraViewHelper.emitBarcodesDetectedEvent(this, barcodesDetected);
+    RNCameraViewHelper.emitBarcodesDetectedEvent(this, barcodesDetected, sourceWidth, sourceHeight, sourceRotation);
   }
 
   public void onBarcodeDetectionError(RNBarcodeDetector barcodeDetector) {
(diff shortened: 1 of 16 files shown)


sibelius:
can you update typescript types?

n1ru4l:
@shashankkushwah You added `react-native-gesture-handler`, however I cannot see any usages.

shashankkushwah:
> can you update typescript types?

@sibelius I am quite new to js, can you please explain what do you mean by that?



> @shashankkushwah You added `react-native-gesture-handler`, however I cannot see any usages.
@n1ru4l I added react-navigation for easy navigation on basic example app. react-native-gesture-handle comes as part of it, might be getting used in iOS back swipe.

shashankkushwah:
@sibelius @n1ru4l on another note. 

The sourceWidth and sourceHeight values in onFramePreview doesn't match with the actual CameraView width and height.

`public void onFramePreview(CameraView cameraView, byte[] data, int width, int height, int rotation)`

this is what the scanning result is for the screen below 

device-2019-02-07-093644

scan result

```
{
  "target": 267,
  "barcodes": [
    {
      "bounds": {
        "origin": {
          "y": "698",
          "x": "320"
        },
        "size": {
          "height": "6",
          "width": "656"
        }
      },
      "type": "CONTACT_INFO",
      "data": "8873845100011B37"
    },
    {
      "bounds": {
        "origin": {
          "y": "869",
          "x": "443"
        },
        "size": {
          "height": "41",
          "width": "534"
        }
      },
      "type": "CONTACT_INFO",
      "data": "887384FF00011B37"
    }
  ],
  "sourceRotation": 90,
  "sourceHeight": 1200,
  "sourceWidth": 1600,
  "type": "barcode"
}
```

The cameraview dimension are 1440x1875. am I missing something?

sibelius:
upgrade the types here

https://github.com/react-native-community/react-native-camera/blob/master/types/index.d.ts

sibelius:
it would be better to update example in another pull request

good enough for now

tks

PR #2101, opened 2019-02-11
[WIP] Add usePersistentStorage

aleksanb:
Do not delete this branch! It's needed for master.

### DIFF
--- a/android/src/main/java/org/reactnative/camera/CameraModule.java
+++ b/android/src/main/java/org/reactnative/camera/CameraModule.java
@@ -242,7 +242,11 @@ public class CameraModule extends ReactContextBaseJavaModule {
   @ReactMethod
   public void takePicture(final ReadableMap options, final int viewTag, final Promise promise) {
     final ReactApplicationContext context = getReactApplicationContext();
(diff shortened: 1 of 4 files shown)


n1ru4l:
Why not use the `CameraRoll` module instead? https://facebook.github.io/react-native/docs/cameraroll

Do we really need this functionality inside `react-native-camera`? cc @sibelius

aleksanb:
Oh whops, I meant to create this PR against our own fork of the repo. I might clean this up and create a proper PR to react-native-camera.

Our use case is the following: We're a self-service car rental company and require people to take photos directly from the camera to prevent them from being taken beforehand in the case of damage on cars. The default storage mechanism for react-native-camera is temporary storage. Sometimes people don't have internet while returning / picking up cars, so we're required to store them offline for upload later. Previously we'd use the normal storage for react-native-camera, but as that's temp storage, sometimes the pictures would just disappear due to the OS having cleaned them up. To prevent the risk of such a disk space reclaiming happening on our photos, we created the possibility of storing images directly to persistent storage, which prevents the photos from being wiped randomly.

n1ru4l:
Ok I will close this for now!

PR #2106, opened 2019-02-13
Thumbnail android

thearad:
Added in android thumbnail changes, local device is sending thumbnails to public s3, files generated on 2/13 are the tests.


thearad:
JK

PR #2117, opened 2019-02-19
build(windows): update min sdk version

Hirurgo:
It's time to update TargetPlatformMinVersion in order to be in accordance with the react-native-windows
and prevent this kind of errors

image

https://github.com/Microsoft/react-native-windows/blob/master/ReactWindows/ReactNative/ReactNative.csproj#L15


Hirurgo:
Incredibly fast work! Thanks you!)
It was a little bit faster than PR with camera for windows =)

Issue #2085, opened 2019-01-31
Missing NSMicrophoneUsageDescription key (#2)


404sand808s:
Hey, sorry for opening another issue with this - tried to comment on the other thread but it was locked to me - but I think the issue may not be resolved. My app is still being automatically rejected because it wants me to have NSMicrophoneUsageDescription in info.plist. I also have put captureAudio={false} in my use of the RNCamera component :)

 Link to original issue [here](https://github.com/react-native-community/react-native-camera/issues/2069) I'm still getting the NSMicrophoneUsageDescription issue when I upload my app to the App Store, and I'm on the latest version of react-native-camera (1.10.10).

Just wanted to put this out there in case it's not just me!

n1ru4l:
Related: 
- https://github.com/react-native-community/react-native-camera/pull/2048
- https://github.com/react-native-community/react-native-camera/pull/2070


Because of this code Apple Rejects the app if `NSMicrophoneUsageDescription` is not provided. This sucks since the code will never be called if you set the property `captureAudio` to `false` this code will never be called.

```objc
[[AVAudioSession sharedInstance] requestRecordPermission:^(BOOL granted) {
  resolve(@(granted));
}];
```

I already thought about a way to fix it, however this will be a breaking change: 

By default `captureAudio` is `false`.

If you want to record audio you have to set `captureAudio` to `true` and also add `#define REACT_NATIVE_CAMERA_ENABLE_RECORD_AUDIO YES` to your `AppDelegate.h`.

This would involve some more setup code from the library consumers, however it will allow us to retain the meaningful warnings while not being rejected by Apple.

What is your opinion on this?

**Workarounds:**

- Add `NSMicrophoneUsageDescription` to `Info.plist`
-  Change the code according to this comment: https://github.com/react-native-community/react-native-camera/issues/2069#issuecomment-457507508 (E.g. with [patch-package](https://www.npmjs.com/package/patch-package))

adimshev:
Same trouble, version - 1.11.1

n1ru4l:
@adimshev Can you please test if #2119 resolves this issue?

adimshev:
I already added the NSMicrophoneUsageDescription, uploaded it to App Store and switched to another project. So I can't. Sorry.

n1ru4l:
Well I do not have a project where I can test this on. Once someone can we can resolve this issue.

adimshev:
I need run `react-native link` again?

[...]

PR #2119, opened 2019-02-20
fix: only call requestRecordPermission in debug mode

n1ru4l:
@adimshev Could you please test if this resolves the issue you are experiencing in #2085?

```bash
# yarn
yarn add react-native-camera@git+https://git@github.com/react-native-community/react-native-camera.git#fix-2085
# npm
npm install --save react-native-camera@git+https://git@github.com/react-native-community/react-native-camera.git#fix-2085
```


bonkowski:
I just verified that the fix works. My app was removed from TestFlight because we did not explain why we needed the microphone, which we didn't. After using this fix, it was accepted.

idrisssakhi:
tested for me i use the qrcode scanner and this solved my issue

ldgarcia:
The fix works for me too. Thanks!

sipotat:
Also for me. Thank!

PR #2135, opened 2019-03-07
feat(android): support inverted qrcode

zjkhiyori:
I found the lib could not recognized the QRCode which inverted(generic QRCode is white background and filling with black, the inverted is black background and filling with white)

but iOS System Camera can recognized


zjkhiyori:
inverted QRCode

generic QRCode

zjkhiyori:
> So many try catches, is there a better way?

Emm... cause of `decodeWithState` design
```
public Result decodeWithState(BinaryBitmap image) throws NotFoundException {
    // Make sure to set up the default state so we don't crash
    if (readers == null) {
      setHints(null);
    }
    return decodeInternal(image);
}
```
I have no idea now, can anyone give some advise

sibelius:
good enough for me

PR #2147, opened 2019-03-15
fix: setting barCodeTypes prop

apz77:
JS enum values weren't correctly converted into appropriate strings for java code.


sibelius:
@dov11 does this make sense?

sibelius:
@apz77 check this comment https://github.com/react-native-community/react-native-camera/issues/2155#issuecomment-475012199

dov11:
> @dov11 does this makes sense?

Not really, e.g. if user sets `barCodeTypes={[RNCamera.Constants.BarCodeType.qr]}`, The value of `barCodeTypes.getString(i)` in line 117 will be "QR_CODE" - a value of the Map `CameraModule.VALID_BARCODE_TYPES`, not a key. So `formatString` will always be null, because [get method expects a key](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#get-java.lang.Object-).

Also aren't changes to RNCameraView.java already in, after PR #2145?

sibelius:
gonna close this now, reopen with more information and how to test it

tks

PR #2162, opened 2019-03-23
Upgrade to React Native 0.59.1

reime005:
Upgrade to React Native 0.59.1


sibelius:
Can you fix CI status?

examples/basic/android/gradle.properties
```diff
@@ -18,3 +18,4 @@
 # org.gradle.parallel=true
 
 android.useDeprecatedNdk=true
+android.enableAapt2=false
```
friederbluemle: Just checking: Is this needed here? The default template of RN 0.59.1 doesn't have it. (`useDeprecatedNdk` is _not_ present as well). Also don't remove the newline character from the last line.

friederbluemle: Based on this warning in the build log there is definitely no need to add this line here:

```
WARNING: The following project options are deprecated and have been removed: 
android.enableAapt2
This property has no effect, AAPT2 is now always used.
```

reime005: sure, was just a try

sibelius:
any guess why this is failing?

sibelius:
💯

PR #2172, opened 2019-03-27
fix(android): fix code style to avoid try catch nest #2135

zjkhiyori:
Hi @sibelius , i think change to this will be better


zjkhiyori:
add ref #2135

Issue #2179, opened 2019-03-31
Build Failed


shtengel:
Build is failing when adding react-native-camera as a dependency in build.gradle (app),
saw it happen to someone else here but no one replied... anyone know how to fix it ?
i tried to exclude com.facebook from react-native-camera but it does nothing 

# Bug Report
**To Do First**
- [ V] Did you try latest release?
- [ V] Did you try master?
- [V ] Did you look for existing matching issues?

**Related Modules**

**Platforms**

**Versions**
- Android: 28
- react-native-camera: 2.1.1
- react-native: 0.51.1
- react: 16.0.0

**Description/Current Behaviour**
AGPBI: {"kind":"error","text":"Program type already present: com.facebook.infer.annotation.Mutable","sources":[{}],"tool":"D8"}

Execution failed for task ':app:transformDexArchiveWithExternalLibsDexMergerForX86Debug'.
  Program type already present: com.facebook.infer.annotation.Mutable

**Steps to Reproduce**
Add react-native-camera as a dependency to gradle

shtengel:
commenting out `implementation 'com.facebook.infer.annotation:infer-annotation:+'` makes it to build succesfully

sibelius:
Where?

shtengel:
works if i change it to `implementation com.facebook.infer.annotation:infer-annotation:10.0.0`
in `react-native-camera->build.gradle`

Strate:
Can confirm that issue exists, and confirm that commenting out `implementation 'com.facebook.infer.annotation:infer-annotation:+'` fixes issue

sibelius:
can somebody send a pull request that solve this?

Strate:
@sibelius just with commenting line? I am not sure that exactly happens, for example I don't use mlkit, maybe this is related.

tadeumx1:
@sibelius I can work on this

shtengel:
If you comment out then you can build but then runtime exception is thrown ( in my case)

PR #2182, opened 2019-04-01
Fix Build

tadeumx1:
Because of issue #2179 I am changing the version of com.facebook.infer.annotation to fix the build

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -73,7 +73,7 @@ dependencies {
   def googlePlayServicesVisionVersion = safeExtGet('googlePlayServicesVisionVersion', safeExtGet('googlePlayServicesVersion', '17.0.2'))
 
   implementation 'com.facebook.react:react-native:+'
-  implementation 'com.facebook.infer.annotation:infer-annotation:+'
+  implementation 'com.facebook.infer.annotation:infer-annotation:10.0.0'
   implementation "com.google.zxing:core:3.3.3"
   implementation "com.drewnoakes:metadata-extractor:2.11.0"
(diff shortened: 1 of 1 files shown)


tadeumx1:
I tried with all versions of this avaiable https://mvnrepository.com/artifact/com.facebook.infer.annotation/infer-annotation but nothing build

Strate:
@tadeumx1 why you change version to `10.0.0`, but https://mvnrepository.com/artifact/com.facebook.infer.annotation/infer-annotation has only 0.* versions?

tadeumx1:
First I changed to this version because of the commentary of @shtengel on issue #2179 but after I realized that the package, doesnt have this version, so I found this list of versions available to this package, and I tried all of them but nothing build

shtengel:
I probably took 0.10.0.0 and forgot the initial :) 
My issue was that it failed to build when gradle has com.facebook.infer.annonations:+
So I checked 0.10.0.0 and then I managed to build. I was changing the build.gradle of react-native-camera

shtengel:
@tadeumx1  i changed it to 0.10.0
`  implementation 'com.facebook.infer.annotation:infer-annotation:0.10.0'`
but maybe there is a deeper issue..

tadeumx1:
Thanks for the help, @shtengel I think that I tried with this version of com.facebook.infer.annotation, but I am going to make a commit with this

shtengel:
Thank you for fixing this issue so fast ! :)

sibelius:
fixed here https://github.com/react-native-community/react-native-camera/pull/2194

tks

PR #2204, opened 2019-04-11
fix: picture sizes not updated when ratio changed

hushicai:
I think it is safe to update pictureSizes every time the `adjustCameraParameters` method called.


sibelius:
Does this happen on camera2?

agrass:
Hello, I think this change is generating problems in some phones. After update to this version I'm getting this error: "Fatal Exception: java.lang.RuntimeException: setParameters failed", may be is related with this: https://stackoverflow.com/a/33580067/4042982

hushicai:
In fact, without this,i also encounter `setParameters failed` before.

Antonio Grass <notifications@github.com>于2019年5月29日 周三上午2:31写道：

agrass:
You are right, I undo this change and still with the error

n1ru4l:
I am locking this conversation. If you do still experience this problem and cannot find another issue that is about that problem, feel free to open a new issue.

PR #2207, opened 2019-04-12
[iOS] Video Interrupted on incoming call fix

hisankaran:
with this `isRecordingInterrupted` will reflect when recording is paused by incoming call


hisankaran:
Great, thanks!!

cristianoccazinsp:
Hey @hisankaran . I wonder what was the original issue with this? Is there really a reason top stop video recording / make the app go to background when audio is interrupted? Considering that when you take a call, you are most likely making the app go to background anyways, and that will already stop the video recording, what was the precise reason for this change?

Can we simply stop audio instead? This way, recording could continue (unless the app fully goes to background)

PR #2225, opened 2019-04-23
Extend interface with _cameraRef and _cameraHandle

Noitidart:


sibelius:
why is this needed?

Noitidart:
It's to make the type more accurate for TypeScript. TypeScript complains without it. I was using `this.ref._cameraRef.measure` (`<RNCamera ref={this.ref} />`) to get the measurements of the view to figure out the `autoFocusRegionOfInterest` width/height.

Noitidart:
Wow super fast collaboration! thank you!

PR #2245, opened 2019-04-29
Add prop of autoFocusPointOfInterest

Noitidart:


Noitidart:
Super fast thank you!

PR #2250, opened 2019-05-04
save files in custom directory

SychevSP:
In my application I need to save photos and videos to different directories.
I patched react-native-camera to make it possible to specify custom paths to folders.
One can construct path to directory with a library such as RN-Fetch-Blob and pass it as an option to takePictureAsync or recordAsync.
If the change seems beneficial I am happy to share it with the community.

### DIFF
--- a/android/src/main/java/org/reactnative/camera/CameraModule.java
+++ b/android/src/main/java/org/reactnative/camera/CameraModule.java
@@ -254,7 +254,9 @@ public class CameraModule extends ReactContextBaseJavaModule {
   @ReactMethod
   public void takePicture(final ReadableMap options, final int viewTag, final Promise promise) {
     final ReactApplicationContext context = getReactApplicationContext();
-    final File cacheDirectory = mScopedContext.getCacheDirectory();
+    final File cacheDirectory = options.hasKey("saveInDirectory") ?
+      new File(options.getString("saveInDirectory")) :
+      mScopedContext.getCacheDirectory();
     UIManagerModule uiManager = context.getNativeModule(UIManagerModule.class);
     uiManager.addUIBlock(new UIBlock() {
       @Override
@@ -276,7 +278,9 @@ public class CameraModule extends ReactContextBaseJavaModule {
   @ReactMethod
   public void record(final ReadableMap options, final int viewTag, final Promise promise) {
       final ReactApplicationContext context = getReactApplicationContext();
-      final File cacheDirectory = mScopedContext.getCacheDirectory();
+      final File cacheDirectory = options.hasKey("saveInDirectory") ?
+        new File(options.getString("saveInDirectory")) :
+        mScopedContext.getCacheDirectory();
       UIManagerModule uiManager = context.getNativeModule(UIManagerModule.class);
 
       uiManager.addUIBlock(new UIBlock() {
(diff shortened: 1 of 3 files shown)


MateusAndrade:
This looks really good! Im going to give a try on this with Android. If someone can handle the test on IOS would be great, since i cant test this on the weekend.

sibelius:
can't you use this https://github.com/itinance/react-native-fs?

SychevSP:
If you want to save photos/ videos to a custom directory instead of cacheDir/Camera you can take the path returned by takePictureAsync or recordAsync and then feed this path to another library to move the file.  
It can be  https://github.com/itinance/react-native-fs or https://github.com/joltup/rn-fetch-blob or any other library that can move files.
But it is a two step solution: first take picture, then move it

In my app I wanted to do it in one step. So made this little tweak to react-native-camera.

android/src/main/java/org/reactnative/camera/CameraModule.java
```diff
@@ -254,7 +254,9 @@ public void execute(NativeViewHierarchyManager nativeViewHierarchyManager) {
   @ReactMethod
   public void takePicture(final ReadableMap options, final int viewTag, final Promise promise) {
     final ReactApplicationContext context = getReactApplicationContext();
-    final File cacheDirectory = mScopedContext.getCacheDirectory();
+    final File cacheDirectory = options.hasKey("saveInDirectory") ?
```
sibelius: can you move this to a common function?

ios/RN/RNCamera.m
```diff
@@ -417,7 +417,13 @@ - (void)takePicture:(NSDictionary *)options resolve:(RCTPromiseResolveBlock)reso
             NSMutableDictionary *response = [[NSMutableDictionary alloc] init];
             float quality = [options[@"quality"] floatValue];
             NSData *takenImageData = UIImageJPEGRepresentation(takenImage, quality);
-            NSString *path = [RNFileSystem generatePathInDirectory:[[RNFileSystem cacheDirectoryPath] stringByAppendingPathComponent:@"Camera"] withExtension:@".jpg"];
+            NSString *path = nil;
+            if (options[@"saveInDirectory"]) {
```
sibelius: move this logic to a common function

sibelius:
I think we can land this

let's add to new docs as well

MateusAndrade:
> I think we can land this
> 
> let's add to new docs as well

Would be nice apply this with our new docs using docusaurus. We can even modify  READM.MD showing this.

zackify:
Does this library currently store data in the documents folder of the app? If not, how would I use this PR to save photos there https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html

sibelius:
@zackify could you test if this PR solve your use case?

zackify:
What I’m trying to say is that I don’t understand where the temporary photos are currently stored, or if this PR will let me put them in a persistent place. 

Basically I need to store photos in my app, but not to the camera roll. But I want them to stay on the phone until the app is deleted

fabriziobertoglio1987:
@zackify the photos are stored in [cache](https://developer.android.com/studio/debug/device-file-explorer), but as explained above you can use other libraries to save the file to a specific location.

@SychevSP are you still working on this? Or should I fork it and make the required changes?

SychevSP:
I have a more recent version of this feature which I forked in December. 
I currently use in one of my projects
https://github.com/SychevSP/react-native-camera/commits/save-in-dir

Issue #2246, opened 2019-05-01
Face detection not working on android front camera


Chimaa123:
# Bug Report

**To Do First**

- [x] Did you try latest release?
- [x] Did you try master?
- [x] Did you look for existing matching issues?

**Platforms**

Android

**Versions**
all version after 1.8.1

- Android:
- react-native-camera:

  

**Description/Current Behaviour**
Face detection is not working on Android.

**Expected Behaviour**
Face detection should work

**Steps to Reproduce**

**Additionals**
I found that this issue raised cause of #f7fb3e3 commit (https://github.com/react-native-community/react-native-camera/commit/f7fb3e361d4f1b773ee93bea8336f276e84b692c) which is released in v1.8.1.

> Love react-native-camera? Please consider supporting our collective: 👉 https://opencollective.com/react-native-camera/donate
> Want this issue to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

sibelius:
duplicated of https://github.com/react-native-community/react-native-camera/issues/2221

Chimaa123:
#2221 is only on IOS, but this issue is on android

RWOverdijk:
@Chimaa123 What do you mean recovering? I'm having the same issue

RWOverdijk:
I missed the release. Verifying it works now.

RWOverdijk:
@n1ru4l Closer, but still no. Not it's inverted horizontal causing inverted readings.

Issue #2221, opened 2019-04-20
v2.6.0 face detection doesn't work


michaelcurtis:
# Bug Report

**To Do First**

- [ Y ] Did you try latest release?
- [ Y ] Did you try master?
- [ Y ] Did you look for existing matching issues?

**Platforms**

iOS

**Versions**

- XCode: 10.2 (10E125)
- iOS: 12.2
- react-native-camera: 2.6.0 (tried master too)
- react-native: 0.59.5
- react: 16.8.3
  

**Description/Current Behaviour**

The onFacesDetected callback is fired endlessly with an empty faces object. The value looks like this: { type: 'face', faces: [], target: 23 }. It does this whether or not faces are in the camera field of view. 

**Expected Behaviour**

The onFacesDetected callback to be fired: 1. when faces are detected and 2. with a populated faces object.

**Steps to Reproduce**

1. Create new react native app
2. yarn add react-native-camera@2.6.0
3. Update Podfile
4. pod install
5. Set up Firebase (confirmed it connected in Firebase when the app runs)
6. Compile and run

***Podfile***
```
target 'CameraTest' do
  pod 'React', :path => '../node_modules/react-native', :subspecs => [
    'Core',
    'CxxBridge',
    'DevSupport',
    'RCTText',
    'RCTNetwork',
    'RCTWebSocket',
    'RCTAnimation',
  ]

  pod 'yoga', :path => '../node_modules/react-native/ReactCommon/yoga'
  pod 'DoubleConversion', :podspec => '../node_modules/react-native/third-party-podspecs/DoubleConversion.podspec'
  pod 'glog', :podspec => '../node_modules/react-native/third-party-podspecs/glog.podspec'
  pod 'Folly', :podspec => '../node_modules/react-native/third-party-podspecs/Folly.podspec'

  pod 'Firebase/Core'
  pod 'react-native-camera', path: '../node_modules/react-native-camera', subspecs: [
    'FaceDetectorMLKit'
  ]

  target 'CameraTestTests' do
    inherit! :search_paths
  end
end

post_install do |installer|
  installer.pods_project.targets.each do |target|
    if target.name == "React"
      target.remove_from_project
    end
  end
end
```

***AppDelegate.m***
```
#import "AppDelegate.h"

#import <React/RCTBridge.h>
#import <React/RCTBundleURLProvider.h>
#import <React/RCTRootView.h>

@import Firebase;

@implementation AppDelegate

[...]

Issue #1962, opened 2018-12-02
Face recorgnition


shohailahm:
# Question

**To Do First**
- [ ] Take a look in the [README](https://github.com/react-native-community/react-native-camera/blob/master/README.md)
- [ ] Take a look in the [docs](https://github.com/react-native-community/react-native-camera/blob/master/docs/RNCamera.md)
- [ ] Take a look in the [QA](https://github.com/react-native-community/react-native-camera/blob/master/docs/QA.md)

**Ask your Question**

**Tags**

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate
> Want this issue to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

Onfaces detected is getting called when faces comes infront of camera but doesnt fire event when the faces go out of camera?

n1ru4l:
Duplicate of #1963

Issue #1325, opened 2018-03-07
Android front camera vertical Flip problem


condorman:
**Steps to reproduce**

- react-native init testRnCamera
- npm install --save git+https://git@github.com/react-native-community/react-native-camera
- follow configuration for AndroidManifest.xml

**Actual behaviour**
```
<RNCamera ref={(cam) => this.camera = cam }
                     style = {styles.preview} 
                     type={RNCamera.Constants.Type.front}
                     flashMode={RNCamera.Constants.FlashMode.off}
                     ratio={"16:9"} />

```
**Test function** 
```
this.camera.recordAsync()
.then(data => { console.log(data); })
.catch(err => console.error(err));
setTimeout(()=>{ this.camera.stopRecording(); },4000);
```

**Problem**
Video camera preview works correctly, functions and writing of video file on device is working properly but video is vertically inverted.

**Environment**
Nodejs version: 8.9.4
React Native version: 0.54.0:
Phisical Device: Android 7.0 (Samsung s7+)
react-native-camera: Master version

sibelius:
This is the default behavior, check native camera app

condorman:
thank you, the default behavior of video recorded is a VERTICAL mirror? it seems strange.. the video produced is upside down. I cant' find information for switch mirror on doc: https://github.com/react-native-community/react-native-camera/blob/master/docs/RNCamera.md

thank you, i try to debug JAVA library

sibelius:
maybe it is a bug instead, misread your issue

condorman:
I find same problem on android cameraview.
[https://github.com/google/cameraview/issues/84](url)

And same problem on Expo:
[https://github.com/expo/expo/issues/1344](url)

At this point I hope this problem is solved on external library.

sibelius:
I think u can try this prop `forceUpOrientation`

condorman:
Thank you, i try but not work, i think is a prop for IOS in TakePictureOptions. My problem is the saved video on device is Upside-down. i think is a problem of android.cameraview library :\

condorman:
I @sibelius, i find the problem in cameraview repository, with [this change ](https://github.com/react-native-community/cameraview/pull/3) the recorded video with FRONT camera in PORTRAIT mode work fine! my issues is solved, thank you!

rkostrab:
This PR still not resolved the issue. 
Using `Google Pixel` - Every record from front camera is flipped upside down.
My dependency is `Master`: 
```
[...]

PR #2257, opened 2019-05-09
Fix android front camera face detection

### LINKED ISSUE #2100: Front facing camera not detecting face
# Bug Report
**To Do First**
- [Yes] Did you try latest release?
- [No] Did you try master?
- [No] Did you look for existing matching issues?

**Related Modules**
RNCamera
FaceDetector

**Platforms**
Android

Front facing camera is not detecting face

**Expected Behaviour**
Front facing camera should detect face

**Steps to Reproduce**
Used example code provided for face detection

telmen:
## Intro

Android front camera face detection has been broken since this PR #2050 had landed I guess and released within **1.8.1** version.

Fixes #2100, #1962, #2246, #2221 maybe #1325

## Tested devices: 

- Pixel 2 (Android Q)
- Samsung S7 (Android 8.0)
- Nexus 5 (Android 6.0.1)

### DIFF
--- a/android/src/main/java/org/reactnative/camera/RNCameraViewHelper.java
+++ b/android/src/main/java/org/reactnative/camera/RNCameraViewHelper.java
@@ -235,7 +235,8 @@ public class RNCameraViewHelper {
 
   public static int getCorrectCameraRotation(int rotation, int facing, int cameraOrientation) {
     if (facing == CameraView.FACING_FRONT) {
-      return (360 - (cameraOrientation + rotation) % 360) % 360;
+      // Tested the below line and there's no need to do the mirror calculation
+      return (cameraOrientation + rotation) % 360;
     } else {
       final int landscapeFlip = rotationIsLandscape(rotation) ? 180 : 0;
       return (cameraOrientation - rotation + landscapeFlip) % 360;


sibelius:
Does this cause any regression?

sibelius:
reverting this https://github.com/react-native-community/react-native-camera/pull/2050

@matt-oakes what do you think of this changes?

matt-oakes:
It's not a full refer, it looks more like a correction. If it's been tested on multiple devices and with other forms of image detection (not just face detection) then it should be ok to merge. I'm not actively using this library much anymore so I can't test it easily.

RWOverdijk:
I think the mirror was needed. I use this module for a small game that checks which eye is closed, and with this PR the results are the opposite of what iOS does.

sibelius:
@RWOverdijk what do you suggest?

Maybe let the user define orientation strategy

RWOverdijk:
@sibelius I don't know, but I'd at least have it be consistent between android and iOS I think.

It's on its side for me on android. Less than ideal. I'm using the facial recognition to register facial expressions which is now a bit wacky. Is there a way to set it straight for me? Even temporarily? I tried simply switching the values on android but I can't do that without quite a bit of math and it's a super hack 😄

sibelius:
Open an issue so we can think in a fix to everybody

EskelCz:
@sibelius In my case the face recognition on android is acting very strange, alternating between empty face array and non empty, basically on every call of onFacesDetected. Needless to say, tested in good light conditions, with a perfectly still shot of my face.
On iOS it's completely stable and reliable. I can record a video if it helps.
Seems like a bug.

Version 2.7.0, on real device Xiaomi Redmi 6A

sibelius:
it is working well for @andreseiji, can you open an issue so we can dig more on this?

andreseiji:
@EskelCz When I first tried the RNCamera, it was on version 2.5.1. In this version, I've had a behaviour similar to what you are describing on Android, where the faces array was switching between empty and having faces detected.

After this PR was merged, I updated the lib to 2.6.0 and the array was stable. However, using the Android front camera, the reading was off due to an orientation issue. I have fixed the orientation issue, and version 2.7.0 worked fine, except for another issue, as @RWOverdijk states, the issue that remains is the inverted y-axis reading.

Currently I am using version 2.9.0 and haven't seen the issue you are describing anymore, since 2.51. Anyway, if you open an issue with more details, I can surely help trying to figure out what could be going on.

EskelCz:
@andreseiji @sibelius I take it back, I somehow missed two versions. you guys are moving really fast! :) at 2.9.0 it's indeed stable. thank you!

PR #2269, opened 2019-05-15
Adding image height and width if skipProcessing is enabled

psk200:
Adding image height and width if skipProcessing is enabled


psk200:
thank you

PR #2288, opened 2019-05-29
fix setParameters error

agrass:
Hi, 
This is a solution to setParameters error, it seems that that there are a lot of border cases when We set the parameters of the Camara, because depending the device it could change and raise a RuntimeError. I have an error in only one specific device Galaxy Tab A 7.0.

This fix will not change any logic, it continue working the same way. The only difference is that is catching RuntimeException at the moment of setting the parameters.
 
I was searching for the best solution, but adding try catch it was the only solution I found, because I can't replicate this in all the devices: https://stackoverflow.com/questions/39878068/exception-java-lang-runtimeexception-setparameters-failed

Let me know what do you think guys.

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -275,7 +275,12 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
         }
         if (mCameraParameters != null && mCamera != null) {
             mCameraParameters.setPictureSize(mPictureSize.getWidth(), mPictureSize.getHeight());
-            mCamera.setParameters(mCameraParameters);
+            try{
+              mCamera.setParameters(mCameraParameters);
+            }
+            catch(RuntimeException e ) {
+              Log.e("Camera1", "setParameters failed", e);
+            }
         }
     }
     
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/google/android/cameraview/Camera1.java
```diff
@@ -275,7 +275,12 @@ void setPictureSize(Size size) {
         }
         if (mCameraParameters != null && mCamera != null) {
             mCameraParameters.setPictureSize(mPictureSize.getWidth(), mPictureSize.getHeight());
-            mCamera.setParameters(mCameraParameters);
+            try{
+              mCamera.setParameters(mCameraParameters);
+            }
+            catch(RuntimeException e ) {
+              Log.e("Camera1", "setParameters failed", e);
```
sibelius: better keep the same tag for this log

Log.e("RNCamera")

agrass: you are right, I'm going to change that.

agrass: Actually it use "CAMERA_1::", should I conserve that?

agrass:
Changed log names, let me know if there are something else that I need to change.

PR #2385, opened 2019-07-23
fix(RNCameraViewHelper): Check for TAG_ORIENTATION when setting exif …

### LINKED ISSUE #1053: Orientation of captured photo wrong

CareerFairPlus:
…attributes so Integer can be used

# Summary 
Fixes #1053 on Android by writing out correct EXIF data for picture orientation (using Integer type instead of Double type).  Affects RNCameraViewHelper.

## Test Plan

Used http://exif.regex.info/exif.cgi for before and after tests of images

I was able to reproduce this issue using a Google Pixel 2 XL phone.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    N/A     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/org/reactnative/camera/RNCameraViewHelper.java
+++ b/android/src/main/java/org/reactnative/camera/RNCameraViewHelper.java
@@ -321,7 +321,12 @@ public class RNCameraViewHelper {
           exifInterface.setAttribute(key, Boolean.toString(exifMap.getBoolean(key)));
           break;
         case Number:
-          exifInterface.setAttribute(key, Double.toString(exifMap.getDouble(key)));
+          // This is to fix an issue with using doubles for Orientation
+          if (key.equals(ExifInterface.TAG_ORIENTATION)) {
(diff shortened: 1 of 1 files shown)


fellmann:
This is only a partial fix, and still produces invalid EXIF data. Please consider https://github.com/react-native-community/react-native-camera/pull/2396 instead!

CareerFairPlus:
@fellmann You are correct.  Your PR is a much better, more complete fix.

PR #2410, opened 2019-08-04
feat: get format for android

TVMD:
get format when barcode detected on android


sibelius:
can you add this to type definition as well?

TVMD:
> can you add this to type definition as well?

Sure, i'll do it

PR #2423, opened 2019-08-12
feat: (iOS Only) Discussion - auto-expose on point of interest && trigger on subject change

zyasir:
# Summary

This PR is for discussion of potential API to solve the "motivation" below. As the PR is right now, it changes pre-existing behavior (adds autoexposure when autoFocusPointOfInterest is set). I did not update readme or changelong yet as I expect this API to change based on feedback from you all.

Motivation: On the focused point, I also needed auto exposure. I also needed to let go of the auto-focused area once subject changes.

Feature: Auto exposure on the `autoFocusPointOfInterest`. Trigger when subject changes (so I can null out the `autoFocusPointOfInterest`).
How: I modified current `updateAutoFocusPointOfInterest` to also auto expose. I then added the delegate of onSubjectMonitor to trigger a callback. Subject monitor only triggers if is focused on something (`autoFocusPointofInterest` is set), as if no point of focus I assume there is no subject.

* To test that auto exposure is also happening, set an `autoFocusPointOfInterest`, then move the camera from a dark area to bright area, and you should see the camera dim.
* To test the subject monitor, add a prop `onSubjectAreaChanged={e => console.log(e)}`
* To test clearing the focus point on subject area change, inside this callback (2nd bullet, clear set the `autoFocusPointOfInterest` to `null`).

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/RN/RNCamera.h
+++ b/ios/RN/RNCamera.h
@@ -91,5 +91,6 @@
 - (void)onText:(NSDictionary *)event;
 - (void)onBarcodesDetected:(NSDictionary *)event;
 - (bool)isRecording;
+- (void)onSubjectAreaChanged:(NSDictionary *)event;
 
 @end
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -33,6 +33,9 @@
 @property (nonatomic, copy) NSDate *startFace;
 @property (nonatomic, copy) NSDate *startBarcode;
 
+@property (nonatomic, copy) RCTDirectEventBlock onSubjectAreaChanged;
+@property (nonatomic, assign) BOOL isFocusedOnPoint;
+
 @end
 
 @implementation RNCamera
@@ -84,6 +87,8 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
                                                    object:nil];
         self.autoFocus = -1;
 
+        self.isFocusedOnPoint = false;
+
     }
     return self;
 }
@@ -120,6 +125,13 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
 {
     if (_onTextRecognized && _session) {
         _onTextRecognized(event);
+    }
+}
+
+- (void)onSubjectAreaChanged:(NSDictionary *)event
+{
+    if (_onSubjectAreaChanged) {
+        _onSubjectAreaChanged(event);
     }
 }
 
@@ -218,6 +230,37 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
     [device unlockForConfiguration];
 }
 
+- (void)defocusPointOfInterest
+{
+    AVCaptureDevice *device = [self.videoCaptureDeviceInput device];
+
+    if (self.isFocusedOnPoint) {
+
+        self.isFocusedOnPoint = false;
+
+        device.subjectAreaChangeMonitoringEnabled = NO;
+        [[NSNotificationCenter defaultCenter] removeObserver:self name:AVCaptureDeviceSubjectAreaDidChangeNotification object:device];
+
+        CGPoint prevPoint = [device focusPointOfInterest];
+
+        CGPoint autofocusPoint = CGPointMake(0.5f, 0.5f);
+        CGPoint exposurePoint = CGPointMake(0.5f, 0.5f);
+
+        [device setExposurePointOfInterest: autofocusPoint];
+        [device setFocusPointOfInterest: exposurePoint];
+
+        [device setFocusMode:AVCaptureFocusModeContinuousAutoFocus];
+        [device setExposureMode:AVCaptureExposureModeContinuousAutoExposure];
+
+        [self onSubjectAreaChanged:@{
+            @"prevPointOfInterest": @{
+                @"x": @(prevPoint.x),
+                @"y": @(prevPoint.y)
+            }
+        }];
+    }
+}
+
 - (void)updateAutoFocusPointOfInterest
 {
     AVCaptureDevice *device = [self.videoCaptureDeviceInput device];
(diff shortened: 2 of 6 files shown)


YasirZillow:
Example:


```
import React, { useState } from 'react';
import RNCamera from 'react-native-camera';

function Example() {
    cosnt [autoFocusPointOfInterest, setAutoFocusPointOfInterest] = useState({ x: 0.5, y: 0.5 });

    return (
        <RNCamera
            autoFocusPointOfInterest={autoFocusPointOfInterest}
            onSubjectAreaChanged={() => {
                setAutoFocusPointOfInterest(null);
            }}
        />
    );
}
```

sibelius:
is this related to this PR https://github.com/react-native-community/react-native-camera/pull/2419?

YasirZillow:
Thanks @sibelius for such a fast reply! From my look, I think it is not related. This is "automatic adjustment of exposure based on subject, continuously as the he moves the camera". That PR seems to just set the exposure to a number and it remains at that.

YasirZillow:
I will definitely have to update my PR if that is merged. 

Features are different however, I set here `setExposurePointOfInterest` to adjust itself based on the point is at. Just like auto focus at point.

sibelius:
does that PR makese sense to you ?

YasirZillow:
I don't have epxerience with the Android side of things.

But for iOS I have some experience. They are manually setting exposure, and it stays at that level. Setting autofocus point removes that level. I asked some questions in the review.

sibelius:
is it a good idea to merge this PR with the other one?

YasirZillow:
I think it's different, his is maintain at level manually. Mine is an option while focused. I'm talking to him about it in the review, we're still working it out. Thanks very much for bridging us.

sibelius:
@durac what are your thoughts on this?

durac:
I can't really say a lot here since I have no knowledge about the iOS side, but on Android it is the default behaviour to also set auto exposure on the autoFocusPointOfInterest at least in my understanding.

YasirZillow:
Note to self: Based on the feedback in #2419 lets remove the auto exposure in the update auto focus point. I and the PR #2419 were treating this as "tap to focus" but it is not.

cristianoccazinsp:
I'm curious if this would mean that setting the focus area would also cause the camera to adjust brightness/exposure to properly match that area just like the native camera app? 

I can't tell if https://github.com/react-native-community/react-native-camera/pull/2419 was supposed to allow that or not, but I'm not able to achieve that behaviour with the current code.

zyasir:
Hi @cristianoccazinsp - when I fix this PR, you can set a prop:
```
const point = { foo foo foo }; 

<RNCamera 
  autoExposurePointOfInterest={point}  
  autoFocusPointOfInterest={point}
/>
```

And it will give you the same feature you see on tap to focus on native camera.

cristianoccazinsp:
Is there anything blocking this?  I've commented on the other PR that adding the native camera behaviour could be included very easily by passing an additional parameter to autoFocusPointOfInterest. 
The subject change event, however, is more advanced.

cristianoccazinsp:
@zyasir do you think the changes are done and it is just a merge issue? I'm willing to give these changes a try on my own branch.

zyasir:
@cristianoccazinsp I have to change a lot to match the style and discussion points of the other PR. Currently this PR does not offer `autoExposurePointOfInterest`, it does auto exposure on same `autoFocusPointOfInterest`. You can use it, it will work. I use it in production.

cristianoccazinsp:
Not entirely sure what are all the updates you need to do. If the only goal
is to also auto expose on the POI, I've suggested a very small change
(passing a third argument to autoFocusPointOfInterest) that would solve the
issue right away.

However, your PR goes beyond that and also allows you to reset the POI on
subject change, I think that change is also very valuable.

I might be able to take your changes and propose a new PR based on the
previous changes that have already been made, and some of the updates/fixes
I'm still waiting on to get merged. Doesn't seem like it would have a lot
of conflicts.

zyasir:
I just need to follow the discussion we had in the other PR. We don't want to mix autoFocus and autoExpose, to be true to React style we want it separate. 

PR is very usable, just need to follow expected style.

cristianoccazinsp:
I've taken your code and merged it into my fork. I don't really think a new property for auto exposure on point of interest is needed since you will most likely be using the same focus point (it would be really odd to focus on one point and expose on another).

For that reason, I really believe a simple flag to use it or not should be more than enough.

https://github.com/react-native-community/react-native-camera/pull/2514

zyasir:
That's a great idea. However a problem comes up when considering - https://github.com/react-native-community/react-native-camera/pull/2419/files - adds prop of `exposure` level. If we auto focus, while exposure level is set, then the react prop is now lying to us. As exposure is no longer what the prop says it is (if it was set of course).

cristianoccazinsp:
I think that is a very tricky scenario. If you test the native camera, you
can adjust the brightness, but tapping to focus will immediately focus and
change your exposure. When that happens, you need to manually adjust the
exposure again.

I think if someone is implementing both manual exposure and tap to focus,
the user will need to re-set the exposure variable. when that happens, it
will override whatever was set by the tap to focus flag.

Lastly, if a user wants full control over exposure, then he can just not
set the autoExpose flag when setting the point of interest and will need to
control exposure by himself . I think all these cases can be handled by the
developer, and the event that you've implemented when the focus area
changes is also very useful to adjust this if needed.

zyasir:
Yes this was the discussion we had over in PR #2419 - it's a tricky discussion. I also wanted to to auto focus and auto expose at same time. But that chat over there told us:

1. The react way, is to "respect props".
2. We are not offering "tap to focus" feature. We are offering `autoFocusPointOfInterest`. "tap to focus" feature is what the native cameras are doing, and we want to give users the props so they can do that. If we want to offer them an easy way to enable this, that would be nice, but we don't want to break away from "1" which is respect the props.

cristianoccazinsp:
Probably can be closed due to 
https://github.com/react-native-community/react-native-camera/pull/2514

zyasir:
Thanks @cristianoccazinsp! Did they add in the subject did change monitor too?

cristianoccazinsp:
Yes, the event is there. You might want to give it a try to confirm it
works as you expected.

zyasir:
Oh awesome!! Will give it a shot!

zyasir:
@cristianoccazinsp thanks for writing that out and doing that. It definitely does superceed this. That PR is awesome. We had hoped for a `exposurePointOfInterest` prop, as we didn't want to mix in exposure on focus, as per discussion in #2419. I do see people having problems if they set `exposure` and then they use the `autoExposure` here, but they should figure out that using both at same time as a dev he should identify that is obviously odd to use both at same time. He will have to remove the exposure prop and add it back if he used autoExposure in between. 

But awesome job thank you! Closing out this PR.

Issue #2421, opened 2019-08-09
takePictureAsync does not return if focus changed


cristianoccazinsp:
# Bug Report

**To Do First**

- [x] Did you try latest release? Yes
- [x] Did you try master? Yes
- [x] Did you look for existing matching issues? Yes

**Platforms**

Android

**Versions**

- Android: 8 and 9
- react-native-camera: 3.1.0
- react-native: 0.60.4
- react: 16.8.6
  

**Description/Current Behaviour**
If I call setState to change ```autoFocusPointOfInterest``` while the camera is trying to focus to capture after a call too ```takePictureAsync```, the method never resolves and neither rejects, it basically becomes stuck.

**Expected Behaviour**
takePictureAsync should always resolve or reject.

**Steps to Reproduce**

The code to capture looks like:
```
 try{
        data = await this.camera.takePictureAsync(options);
      }
catch(err){
        Alert.alert("Error", "Failed to take picture: " + err.message);
}
```

if a setState is called to update the autoFocusPointOfInterest RNCamera property before the above method resolves, it will never resolve. There are no errors on the android console neither.

cristianoccazinsp:
Update: If flash is turned on, flash/light also gets stuck until the camera component is unmounted. Tested on both Pixel 2 and Moto G5. This makes tap to focus nearly impossible to use on Android.

cristianoccazinsp:
Related to this: https://github.com/react-native-community/react-native-camera/issues/2420

If we remove the call to autoFocus before capturing all issues go away.
https://github.com/react-native-community/react-native-camera/blob/master/android/src/main/java/com/google/android/cameraview/Camera1.java#L453

cristianoccazinsp:
In addition to the above, if we don't remove the autoFocus call, this is most likely related to the additional calls to autoFocus and cancel/reset auto focus. The capture event is most likely lost when this happens and the promise is never resolved. Following https://github.com/react-native-community/react-native-camera/issues/2420, we most likely want to just get rid of the auto focus call (perhaps use a setting for it), which would solve both issues.

cristianoccazinsp:
Related changes: https://github.com/react-native-community/react-native-camera/pull/2427

n1ru4l:
Resolved in #2427

PR #2427, opened 2019-08-14
Update focus handling on Camera1.java to fix #2420 and #2421.

### LINKED ISSUE #2420: autoFocusPointOfInterest not kept while capturing photo
# Bug Report

**To Do First**

- [x] Did you try latest release? Yes
- [x] Did you try master? Yes
- [x] Did you look for existing matching issues? Yes

**Platforms**

Android

**Versions**

- Android: 8 and 9
- react-native-camera: 3.1.0
- react-native: 0.60.4
- react: 16.8.6

autoFocusPointOfInterest on Android correctly changes the camera focus while previewing. However, after calling ```takePictureAsync``` the focus is lost, and the camera captures the picture with a completely different focus. Lastly, the auto focus coordinates seem to reset very quickly (3 seconds max)

I suspect the 3 seconds focus reset comes from here: https://github.com/react-native-community/react-native-camera/blob/master/android/src/main/java/com/google/android/cameraview/Camera1.java#L71

**Expected Behaviour**

Focus should be kept for longer, until the camera has moved enough like to trigger a new focus event. Capturing a photo should not attempt to re-focus.

cristianoccazinsp:
# Summary

First, it updates the take picture call to capture right away instead of trying to focus again. This will match closer to what the native camera does, and also what the Camera2 does.

Secondly, it removes all calls to resetFocus after capturing and after a given timeout. This was causing issues when using autoFocusPointOfInterest, and it really didn't make sense to use an internal timeout to reset the focus. If anything, the user should do this from the JS side. A new method could also be exposed to do this.

Lastly, it adds some error catching logic to the autoFocus call. There was a chance on some devices that calling autoFocus while the device is capturing a photo that it would fail.

## Test Plan

Tested on Google Pixel 2 (Android 9), Motorola G5 (Android 8), Samsung Grand Prime (Android 5?), and LG20 (android 6?). No issues so far, the camera captures way faster than before and never gets stuck while focusing.

An android device with a camera.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅    |

## Checklist

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -450,17 +450,13 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
         if (!mIsPreviewActive) {
             throw new IllegalStateException("Preview is paused - resume it before taking a picture.");
         }
-        if (getAutoFocus()) {
-            mCamera.cancelAutoFocus();
-            mCamera.autoFocus(new Camera.AutoFocusCallback() {
-                @Override
-                public void onAutoFocus(boolean success, Camera camera) {
-                    takePictureInternal(options);
-                }
-            });
-        } else {
-            takePictureInternal(options);
-        }
+
+        // UPDATE: Take picture right away instead of requesting/waiting for focus.
+        // This will match closer what the native camera does,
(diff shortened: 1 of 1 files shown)


cristianoccazinsp:
Any reviews? Complains? Thoughts? Hopefully this can be merged soon since it really improves the camera on Android.

Issue #2235, opened 2019-04-25
While capturing video, Android freeze if home/blocked/minimized


cristianoccazinsp:
# Bug Report

Android

**Versions**

- Android: 8.1 (Moto G5), 9 (Pixel 2)
- react-native-camera: 2.2.2
- react-native: 0.59.4
- react: 16.8.6
- react-navigation: 3.6.0

While capturing a video, even if stopRecording() is called before the app transitions to background/inactive, the app will totally freeze after resuming.

I'm starting the video capture as: ```result = await this.camera.recordAsync(options);```

Then, on the app state change events, if background or inactive, I call ```stopRecording()```

I can get a console print right after the stopRecording call; however, nothing ever outputs after the await call; looks like the promise never resolves.

These are the logs that appear after resuming. No error, no crash, just this.

```
04-25 10:30:53.159 30414-30420/com.zinspector_rn I/zygote: Thread[3,tid=30420,WaitingInMainSignalCatcherLoop,Thread*=0xab7b6200,peer=0x13380120,"Signal Catcher"]: reacting to signal 3
04-25 10:30:53.553 30414-30420/com.zinspector_rn I/zygote: Wrote stack traces to '[tombstoned]'
04-25 10:32:02.787 30414-30420/com.zinspector_rn I/zygote: Thread[3,tid=30420,WaitingInMainSignalCatcherLoop,Thread*=0xab7b6200,peer=0x13380120,"Signal Catcher"]: reacting to signal 3
04-25 10:32:03.076 30414-30420/com.zinspector_rn I/zygote: Wrote stack traces to '[tombstoned]'
```

cristianoccazinsp:
Some more logs. Also, this happens even if the device rotates.

 W/CameraSource: Timed out waiting for incoming camera video frames: 479533666669 us

cristianoccazinsp:
Crash report from bugsnag

ANR Application did not respond for at least 5000 ms 
    MediaRecorder.java:-2 android.media.MediaRecorder.stop
    Camera1.java:1034 com.google.android.cameraview.Camera1.stopMediaRecorder
    Camera1.java:523 com.google.android.cameraview.Camera1.stopRecording
    CameraView.java:557 com.google.android.cameraview.CameraView.stopRecording
    CameraModule.java:313 org.reactnative.camera.CameraModule$7.execute
    UIViewOperationQueue.java:583 com.facebook.react.uimanager.UIViewOperationQueue$UIBlockOperation.execute
[...]

PR #2467, opened 2019-09-05
Fix app freeze on app/host pause on Android Camera1 Impl.

cristianoccazinsp:
# Summary
Fix app freeze on app/host pause on Android. Fix is quite simple, recording has to be stopped before the camera preview.
Fix taken from: https://stackoverflow.com/questions/8914454/mediarecorder-stop-hanging-with-android-4-0-ics/18876200

Should fix https://github.com/react-native-community/react-native-camera/issues/2235

TODO: Review Camera2 implementation since it might have the same issue.

## Test Plan

Tested on Android 8.1 (Moto G5), Android 9 (Pixel 2), Android 10 (Pixel 2)

-

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -159,10 +159,7 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
 
     @Override
     void stop() {
-        if (mCamera != null) {
-            mCamera.stopPreview();
-            mCamera.setPreviewCallback(null);
-        }
+
         mShowingPreview = false;
         if (mMediaRecorder != null) {
             mMediaRecorder.stop();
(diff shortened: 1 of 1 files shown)


cristianoccazinsp:
Honestly, I've never managed to get Camera2 to work properly so I can't test it on a real device. For that reason I refrained myself from doing any change to Camera2.

However, I'm almost sure (just by looking at the code), that there might be a similar issue considering the recorder is the last one stopped (camera and everything else is stopped first).

https://github.com/react-native-community/react-native-camera/blob/4241b39a2206f2e98674c68b14254ea5da24dc07/android/src/main/java/com/google/android/cameraview/Camera2.java#L326

sibelius:
we can review camera2 in other pr, tks

PR #2471, opened 2019-09-06
Improve Android Camera1 error and concurrency handling.

cristianoccazinsp:
# Summary
Improve Android Camera1 error and concurrency handling.

NOTE: Please review carefully, creating this PR for some discussion first.

These changes include the following:
- use atomic boolean for capturing photo flag just like video
- add more exception catching and checks
- raise error instead of failing silently if can't capture photo - improve error handling here
- synchronize stop to avoid race conditions and crashes
- delay params updates (surface) if capturing or recording to avoid bugs
- do not allow video or photo capture if already doing video or photo

## Test Plan

Tested on Android 8.1 (Moto G5), Android 9 and 10 (Pixel 2)

-

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -86,7 +86,7 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
 
     private String mVideoPath;
 
-    private boolean mIsRecording;
+    private final AtomicBoolean mIsRecording = new AtomicBoolean(false);
 
     private final SizeMap mPreviewSizes = new SizeMap();
 
@@ -120,6 +120,8 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
 
     private boolean mIsScanning;
 
+    private boolean mustUpdateSurface;
+
     private SurfaceTexture mPreviewTexture;
 
     Camera1(Callback callback, PreviewImpl preview) {
@@ -127,11 +129,7 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
         preview.setCallback(new PreviewImpl.Callback() {
             @Override
             public void onSurfaceChanged() {
-                if (mCamera != null) {
-                    setUpPreview();
-                    mIsPreviewActive = false;
-                    adjustCameraParameters();
-                }
+                updateSurface();
             }
 
             @Override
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/google/android/cameraview/Camera1.java
```diff
@@ -86,7 +86,7 @@
 
     private String mVideoPath;
 
-    private boolean mIsRecording;
+    private final AtomicBoolean mIsRecording = new AtomicBoolean(false);
```
sibelius: How this works?

cristianoccazinsp: This change was so this flag is also consistent with the photo capture one which was already atomic. Basically, this is to prevent the variable from getting changed concurrently and end up setting up a record or capture twice.

cristianoccazinsp:
> Do you have a state machine of all possible camera states?

Not in a picture, but it is basically ready / not ready / recording / capturing.

I'm still waiting on some more tests from other devices that are known to be more problematic (Samsungs, Xiaomi). But I think these changes make the capture/record calls more similar to iOS' one. If you test on IOS, you will see you get nice errors when trying to capture or record if already doing something. On Android, however, it would either crash or fail silently before these changes.

cristianoccazinsp:
There's still one edge case I couldn't resolve (because there's no way to reject the photo capture promise from Camera1). If the app is paused while the camera is processing a photo (e.g., the stop method is called), the PictureCallback is never fired, and hence the promise is never resolved. However, when the app resumes it will at least reset the capturing flag and the camera will be usable (before the camera would plain simply become unusable and would need to be re-created).

The ideal fix would reject the promise if capturing a photo, but making the promise all the way down to the Camera1 class would mean changing a million interfaces and classes which I think it's not worth the effort for such an edge case.

sibelius:
can you have a Map that store all unresolved promises, and then reject them onResume?

cristianoccazinsp:
The thing is no promise object ever makes it up to the Camera1 class.  I probably don't have enough knowledge of the code, but the only available "communication" from Camera1 to the rest of the components is through the callback object, but sadly, it only has a onPictureTaken callback and not a onPictureError or something like that.

My other guess is that we could fire the picture callback with null data, but that also probably requires a bunch of changes in all the clases that use it. The other problem is that we don't really know if the capture failed, the camera API doesn't give you this info, it just never calls the capture callback, so we would need to rely on a "guess" on the stop method to fire such event.

cristianoccazinsp:
Tested on Xiaomi Mi mix 2, and Galaxy Grand Prim as well, no crashes so far.

Issue #2379, opened 2019-07-17
Ability to select back camera for multi-camera devices


Pauly-D:
# Feature-Request

**Description**
Ability to select which **back** camera to use. Particularly useful for all the multi-camera android devices with wide angle lenses. 

**Possible Implementations**
Since the implementation of [Android API Level 9](https://developer.android.com/training/camera/cameradirect#java) the Camera object supports the passing of an id to select which camera to use (eg. the wide angel lens). 

Which files need to be modified to take advantage of the new android API so that we can update the **type={RNCamera.Constants.Type.back}** for selection of different cameras on the back of android devices?

cristianoccazinsp:
Hey, any luck with this?

cristianoccazinsp:
I believe this is now possible with the above PRs and can be closed.

farhanmeo:
Unable to use Wide angle camera in android

cristianoccazinsp:
Sadly, newer android devices no longer support camera enumeration through the currently implemented method. For these devices, a new implementation path must be followed in order to be able to read from individual logical cameras.

I believe the Camera2 implementation may be able to support that, but it is way too bugged to use right now.

PR #2492, opened 2019-09-17
Android: Support to enumerate and select Camera devices

cristianoccazinsp:
# Summary
This update attempts to bring a very useful feature to support selecting from multiple back/front cameras (e.g., for phones with wide lenses cameras, or multi cameras). 

The update implements a ```getCameraIdsAsync``` methods to enumerate all available cameras, and adds a ```cameraId``` property to the RNCamera component to set it.

This should be 100% backwards compatible. A null cameraId value (the default), will work as it currently does by selecting the camera based on the type value. However, setting an ID will override the type value and select that camera.

I've implemented the change for both Camera1 and Camera2 APIs, but I could really use some extra pair of eyes to test and review the changes since I'm not 100% confident with the internals of the library and I might have missed some stuff.

Lastly, I'm unsure of the flash behaviour if the front facing camera is used. I believe it won't be possible to use a "front" flash right now.

For now, these changes will be Android only. It makes sense considering that only iPhone 11 will support multiple cameras. I also lack Objective-C knowledge and camera internals like to do this change.

Related issue: https://github.com/react-native-community/react-native-camera/issues/2379

## Test Plan
Tested on Pixel 2 (android 9 and 10), Moto G5 (Android 8), LG V20

Real device

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅    |

## Checklist

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -34,6 +34,7 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Properties;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -73,6 +74,7 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
     private Handler mHandler = new Handler();
 
     private int mCameraId;
+    private String _mCameraId;
 
     private final AtomicBoolean isPictureCaptureInProgress = new AtomicBoolean(false);
 
@@ -282,6 +284,26 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
     }
 
     @Override
+    void setCameraId(String id) {
+
+        if(_mCameraId != id){
+            _mCameraId = id;
+
+            // this will call chooseCamera
+            if (isCameraOpened()) {
+                stop();
+                start();
+            }
+        }
+
+    }
+
+    @Override
+    String getCameraId() {
+        return _mCameraId;
+    }
+
+    @Override
     Set<AspectRatio> getSupportedAspectRatios() {
         SizeMap idealAspectRatios = mPreviewSizes;
         for (AspectRatio aspectRatio : idealAspectRatios.ratios()) {
@@ -290,6 +312,23 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
             }
         }
         return idealAspectRatios.ratios();
+    }
+
+
+    @Override
+    List<Properties> getCameraIds() {
+        List<Properties> ids = new ArrayList<>();
+
+        Camera.CameraInfo info = new Camera.CameraInfo();
+
+        for (int i = 0, count = Camera.getNumberOfCameras(); i < count; i++) {
+            Properties p = new Properties();
+            Camera.getCameraInfo(i, info);
+            p.put("id", String.valueOf(i));
+            p.put("type", String.valueOf(info.facing));
+            ids.add(p);
+        }
+        return ids;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


cristianoccazinsp:
Pushed an additional update to remove the requirement from the camera to be running/active. We can still query for this info as long as it is instantiated.

android/src/main/java/com/google/android/cameraview/Camera1.java
```diff
@@ -693,14 +739,25 @@ public Size getPreviewSize() {
...
-                return;
+        if(_mCameraId == null){
+            for (int i = 0, count = Camera.getNumberOfCameras(); i < count; i++) {
+                Camera.getCameraInfo(i, mCameraInfo);
+                if (mCameraInfo.facing == mFacing) {
+                    mCameraId = i;
+                    return;
+                }
+            }
+            mCameraId = 0; // default to the only available camera
```
cristianoccazinsp: @sibelius  Can you comment on these lines? I wonder if we need to do something different here actually to handle the case where a camera is not found.  I think I overlooked this one and it should have been INVALID_CAMERA_ID instead. Thoughts?
Need to test on a device with no cameras, or just 1 camera.

sibelius: hard to find a device without cameras

what happen when you throw? can an user get this error using an ErrorBoundary?

`0` is always the first camera?

won't break the current behavior?

cristianoccazinsp: We can/should probably mimic what Camera2 does and raise an error if there are 0 cameras, but set the camera to the first one available if we didn't find one: https://github.com/react-native-community/react-native-camera/blob/master/android/src/main/java/com/google/android/cameraview/Camera2.java#L688

How does that sound? I believe the Camera1 code originally would just set the camera ID to be the invalid one and you would get no image at all in this case (if lets say you don't have a back camera). It's a very specific case.

I was thinking about something like this:

```
int count = Camera.getNumberOfCameras();
            if(count == 0){
                throw new RuntimeException("No camera available.");
            }

            for (int i = 0; i < count; i++) {
                Camera.getCameraInfo(i, mCameraInfo);
                if (mCameraInfo.facing == mFacing) {
                    mCameraId = i;
                    return;
                }
            }
            // no camera found, set the one we have
            mCameraId = 0;
            Camera.getCameraInfo(mCameraId, mCameraInfo);
```

sibelius: this part is weird

// no camera found, set the one we have
mCameraId = 0;

but it makes sense, you are looking for the camera facing the correct facing the user selected

cristianoccazinsp: The original code would simply set the camera to be the INVALID_ID one, but that behaviour is different from the Camera2 API which simply sets the first available camera if the code wasn't able to find the back/front facing camera. Do you think it makes sense to make both APIs consistent?

sibelius: sure

cristianoccazinsp: I've just added a commit for that.

sibelius:
tests are failing

cristianoccazinsp:
My bad, missed this when moving the file from my test folder to the repo. Should be good now.

cristianoccazinsp:
Doing more tests to the Camera2 API, please do not merge this yet since I've found a few more issues (some not even related to my changes, but will fix them since I'm at it)

cristianoccazinsp:
Did a few more fixes to Camera2.

sibelius:
@cristianoccazinsp is this good enough now?

cristianoccazinsp:
Haven't had issues or crashes so far, so I would say it is good enough. It
also seem to work fine in devices with wide angle lenses (LG V20 and V30)
for example.

Issue #2473, opened 2019-09-08
videoBitrate option for iOS


xHeinrich:
# Feature-Request

**Describe the Feature**
Ideally the same as the android option.

**Possible Implementations**
I believe this can be implemented using the following key around here: https://github.com/react-native-community/react-native-camera/blob/master/ios/RN/RNCamera.m#L601 though I'm no iOS dev so i'm not 100%.

https://developer.apple.com/documentation/avfoundation/avvideoaveragebitratekey?language=objc
https://developer.apple.com/documentation/avfoundation/avcapturemoviefileoutput?language=objc

**Related Issues**
https://github.com/react-native-community/react-native-camera/issues/687
https://github.com/react-native-community/react-native-camera/pull/2055

xHeinrich:
I've hacked this into my own app with some hard coding. If anyone wants to improve upon this and submit a PR feel free to do so.

`    if (@available(iOS 10, *)) {
        AVVideoCodecType videoCodecType = options[@"codec"];
        NSString *videoBitrate = options[@"videoBitrate"];
        NSLog(@"VIDEO BITRATE %@", videoBitrate);
        [self.movieFileOutput setOutputSettings:@{
          AVVideoCodecKey:AVVideoCodecH264,
          AVVideoCompressionPropertiesKey:
              @{
                  AVVideoAverageBitRateKey:videoBitrate
              }
          } forConnection:connection];
    }
`

sibelius:
fixed here https://github.com/react-native-community/react-native-camera/pull/2504

PR #2504, opened 2019-09-24
videoBitrate option for iOS

xHeinrich:
Add videoBitrate option for ios, setting video bitrate requires a codec to be set

# Summary

Android allows this, adding the same functionality to iOS

## Test Plan

Testing on real devices:
iPod Touch
iPad
iPhone 8+

Test with and without videoBitrate key set, support is ios10+

Check video output bitrate is the same or close to the videoBitrate key

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [yes] I have tested this on a device and a simulator
- [yes] I added the documentation in `README.md`
- [yes] I mentioned this change in `CHANGELOG.md`
- [yes] I updated the typed files (TS and Flow)
- [yes] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -76,8 +76,6 @@ interface RecordOptions {
   mute?: boolean;
   mirrorVideo?: boolean;
   path?: string;
-
-  /** Android only */
   videoBitrate?: number;
 
   /** iOS only */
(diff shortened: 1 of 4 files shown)


xHeinrich:
Closes #2473

Issue #1836, opened 2018-10-05
Quick Preview Capture Mode to Beat Slow onPictureTaken on Android


FullstackJack:
# Feature-Request

**Describe the Feature**
Despite recent modifications to improve image capture on Android, camera capture is still slow. For applications needing very fast capture before launching a new screen (i.e. Snapchat, Instagram), we should consider offering a capture mode to enable the use of PreviewCallback.

**Possible Implementations**
There is already an implementation of this for CameraKit-Android. Since this library is now included with Expo, we should consider porting that code over to this package instead of splitting the community over this issue.
https://github.com/CameraKit/camerakit-android/blob/master/camerakit-core/src/main/api16/com/wonderkiln/camerakit/Camera1.java#L450

**Related Issues**
There are lots of closed issues surrounding this, but none of them really fixed the problem. Let's hope this solves it.
https://github.com/react-native-community/react-native-camera/issues/942
https://github.com/react-native-community/react-native-camera/issues/1000
https://github.com/react-native-community/react-native-camera/issues/768

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate

n1ru4l:
@FullstackJack Are you interested in implementing this feature?

bcalik:
Its still so slow on Android despite messing with all RNCamera options. I hope someone will fix it.

n1ru4l:
@bcalik Are you interested in fixing it?

cesartotvs:
@n1ru4 Got it, I'm interested in fixing this issue.

cesartotvs:
@n1ru4l  Got it, I'm interested in fixing this issue.

FullstackJack:
I'm currently building a prototype with video player and focusing on that at the moment, but I'm interested in helping out when I have the time (unless others have already pitched in what I'm suggesting.

cesartotvs:
Just a quick note I will not be adding any new features, but only fixing the performance issues.

n1ru4l:
How would this be implemented?

You pass an option to `takePicture` and the first value of uri will be the "preview". Then afterwards we will fire a event that includes the url to the actual picture?

sibelius:
Having an event could make sense, we can have more events

n1ru4l:
Is it even possible get a preview frame and take photo at once? camerakit-android only supports the one or the other.

sibelius:
Not sure, we should try

[...]

PR #2505, opened 2019-09-25
Much faster picture taking on Android, configurable

Boris-c:
# Summary

Android's native camera takePicture() is very slow. I wasn't happy with it and many people complain too (see #1836 for instance).

Camera1 allows capturing frame previews. By processing frame previews we can avoid using takePicture() altogether.

The previews are in YUV and need to be transformed to RGB. The proposed implementation uses a render script for fast processing.

4 modes are proposed (as a RN prop, "camera1ScanMode"): "none", "eco", "fast" and "boost".

- none: use takePicture(), as currently done by react-native-camera
- eco: capture a frame preview when a picture is requested, then process it
- fast: capture all frame previews and process one when a picture is requested
- boost: capture and process all frame previews. When a picture is requested, send the last one

In my app, on my device, the benchmark is the following (average of 3 pictures):
- none: 1.7 s
- eco: 415 ms
- fast: 361 ms
- boost: 171 ms

These measures were done in React Native, from the call to `capture()` to the result being received after conversion to base64 and bridge transfer. The speed increase of the proposed change if seen at a lower level (Android `takePicture()` vs frame previews processing) is even more dramatic.

I use the library to get pictures in base64. My tests did not cover all possible uses of the library.

Note that the min SDK upgrade to version 17 is needed for the render script.

## Test Plan

I'll leave it to more experienced users of the library to test every possible use case.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -20,7 +20,7 @@ android {
   buildToolsVersion safeExtGet('buildToolsVersion', '28.0.3')
 
   defaultConfig {
-    minSdkVersion safeExtGet('minSdkVersion', 16)
+    minSdkVersion safeExtGet('minSdkVersion', 17)
     targetSdkVersion safeExtGet('targetSdkVersion', 28)
   }
 
--- a/android/src/main/java/org/reactnative/camera/CameraViewManager.java
+++ b/android/src/main/java/org/reactnative/camera/CameraViewManager.java
@@ -142,6 +142,11 @@ public class CameraViewManager extends ViewGroupManager<RNCameraView> {
     view.setUsingCamera2Api(useCamera2Api);
   }
 
+  @ReactProp(name = "camera1ScanMode")
+  public void setCamera1ScanMode(RNCameraView view, String camera1ScanMode) {
+    view.setCamera1ScanMode(camera1ScanMode);
+  }
+
   @ReactProp(name = "playSoundOnCapture")
   public void setPlaySoundOnCapture(RNCameraView view, boolean playSoundOnCapture) {
     view.setPlaySoundOnCapture(playSoundOnCapture);
--- a/android/src/main/java/org/reactnative/camera/RNCameraView.java
+++ b/android/src/main/java/org/reactnative/camera/RNCameraView.java
@@ -3,7 +3,9 @@ package org.reactnative.camera;
 import android.Manifest;
 import android.annotation.SuppressLint;
 import android.content.pm.PackageManager;
+import android.graphics.Bitmap;
 import android.graphics.Color;
+import android.graphics.Matrix;
 import android.media.CamcorderProfile;
 import android.media.MediaActionSound;
 import android.os.Build;
@@ -21,9 +23,11 @@ import org.reactnative.barcodedetector.RNBarcodeDetector;
 import org.reactnative.camera.tasks.*;
 import org.reactnative.camera.utils.RNFileUtils;
 import org.reactnative.facedetector.RNFaceDetector;
+import org.reactnative.camera.utils.YuvToBitmap;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.Field;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
@@ -37,6 +41,19 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
   private Promise mVideoRecordedPromise;
   private List<String> mBarCodeTypes = null;
   private Boolean mPlaySoundOnCapture = false;
+
+  public static final String CAMERA1SCANNONE = "none";
+  public static final String CAMERA1SCANECO = "eco";
+  public static final String CAMERA1SCANFAST = "fast";
+  public static final String CAMERA1SCANSUPERFAST = "boost";
+
+  private YuvToBitmap mYuvToBitmap = null;
+  private Matrix mRotationMatrix = null;
+  private int mLastRotation = -1;
+  private String mCamera1ScanMode = CAMERA1SCANECO;
+  private boolean mProcessingPreview = false;
+  private Bitmap mRotated = null;
+  private final Object COLLECTION_LOCK = new Object();
 
   private boolean mIsPaused = false;
   private boolean mIsNew = true;
@@ -66,6 +83,10 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
   private boolean mTrackingEnabled = true;
   private int mPaddingX;
   private int mPaddingY;
+
+  private synchronized void storeRotated(Bitmap rotated) {
+    mRotated = rotated;
+  }
 
   public RNCameraView(ThemedReactContext themedReactContext) {
     super(themedReactContext, true);
@@ -121,13 +142,20 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
       }
 
       @Override
-      public void onFramePreview(CameraView cameraView, byte[] data, int width, int height, int rotation) {
-        int correctRotation = RNCameraViewHelper.getCorrectCameraRotation(rotation, getFacing(), getCameraOrientation());
+      public void onFramePreview(CameraView cameraView, final byte[] data, final int width, final int height, final int rotation) {
+        final int correctRotation = RNCameraViewHelper.getCorrectCameraRotation(rotation, getFacing(), getCameraOrientation());
         boolean willCallBarCodeTask = mShouldScanBarCodes && !barCodeScannerTaskLock && cameraView instanceof BarCodeScannerAsyncTaskDelegate;
         boolean willCallFaceTask = mShouldDetectFaces && !faceDetectorTaskLock && cameraView instanceof FaceDetectorAsyncTaskDelegate;
         boolean willCallGoogleBarcodeTask = mShouldGoogleDetectBarcodes && !googleBarcodeDetectorTaskLock && cameraView instanceof BarcodeDetectorAsyncTaskDelegate;
         boolean willCallTextTask = mShouldRecognizeText && !textRecognizerTaskLock && cameraView instanceof TextRecognizerAsyncTaskDelegate;
-        if (!willCallBarCodeTask && !willCallFaceTask && !willCallGoogleBarcodeTask && !willCallTextTask) {
+        Promise pictureTakenPromise;
+        synchronized (COLLECTION_LOCK) {
+          pictureTakenPromise = mPictureTakenPromises.poll();
+        }
+        final boolean superFast = mCamera1ScanMode.equals(CAMERA1SCANSUPERFAST) && !mProcessingPreview;
+
+        if (!willCallBarCodeTask && !willCallFaceTask && !willCallGoogleBarcodeTask && !willCallTextTask
+                && pictureTakenPromise  == null && !superFast) {
           return;
         }
 
@@ -169,6 +197,57 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
           textRecognizerTaskLock = true;
           TextRecognizerAsyncTaskDelegate delegate = (TextRecognizerAsyncTaskDelegate) cameraView;
           new TextRecognizerAsyncTask(delegate, mThemedReactContext, data, width, height, correctRotation, getResources().getDisplayMetrics().density, getFacing(), getWidth(), getHeight(), mPaddingX, mPaddingY).execute();
+        }
+
+        if (pictureTakenPromise != null || superFast) {
+          checkScanning();
+          mProcessingPreview = true;
+
+          if (mYuvToBitmap == null) {
+            mYuvToBitmap = new YuvToBitmap(getContext());
+          }
+
+          final Promise aPictureTakenPromise = pictureTakenPromise;
+
+          Thread thread = new Thread() {
+            @Override
+            public void run() {
+              // Get RGB
+              mYuvToBitmap.refreshBitmap(data, width, height);
+
+              Bitmap rotated;
+              if (correctRotation == 0) {
+                rotated = mYuvToBitmap.getBimap();
+              } else {
+                // Rotate
+                if (mLastRotation != correctRotation) {
+                  mLastRotation = correctRotation;
+                  mRotationMatrix = new Matrix();
+                  mRotationMatrix.postRotate(correctRotation);
+                }
+                Bitmap bitmap = mYuvToBitmap.getBimap();
+                rotated = Bitmap.createBitmap(bitmap, 0, 0, bitmap.getWidth(), bitmap.getHeight(), mRotationMatrix, false);
+              }
+
+              mProcessingPreview = false;
+
+              // Resolve
+              synchronized (COLLECTION_LOCK) {
+                Promise promise = aPictureTakenPromise;
+                if (promise == null) {
+                  promise = mPictureTakenPromises.poll();
+                }
+                if (promise == null) {
+                  storeRotated(rotated);
+                } else {
+                  ReadableMap options = mPictureTakenOptions.remove(promise);
+                  final File cacheDirectory = mPictureTakenDirectories.remove(promise);
+                  resolveTakenPicture(rotated, options, promise, cacheDirectory);
+                }
+              }
+            }
+          };
+          thread.start();
         }
       }
     });
@@ -226,20 +305,79 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
     mPlaySoundOnCapture = playSoundOnCapture;
   }
 
+  public Object getImpl() {
+    try {
+      Field field = CameraView.class.getDeclaredField("mImpl");
+      field.setAccessible(true);
+      Object value = field.get(this);
+      field.setAccessible(false);
+      return value;
+
+    } catch (NoSuchFieldException e) {
+      throw new RuntimeException(e);
+
+    } catch (IllegalAccessException e) {
+      throw new RuntimeException(e);
+    }
+
+  }
+
+  private boolean usingCamera2() {
+    Object impl = getImpl();
+
+    if (impl == null) return false;
+
+    String className = impl.getClass().getName();
+    return className.contains("cameraview.Camera2");
+  }
+
+  private boolean useTakePicture() {
+    return usingCamera2() || mCamera1ScanMode.equals(CAMERA1SCANNONE);
+  }
+
+  private void resolveTakenPicture(Bitmap picture, ReadableMap options, final Promise promise, File cacheDirectory) {
+    ResolveTakenPictureAsyncTask task = new ResolveTakenPictureAsyncTask(
+            picture, promise, options, cacheDirectory, 0, RNCameraView.this
+    );
+
+    if (Build.VERSION.SDK_INT >= 11/*HONEYCOMB*/) {
+      task.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR);
+    } else {
+      task.execute();
+    }
+  }
+
   public void takePicture(ReadableMap options, final Promise promise, File cacheDirectory) {
-    mPictureTakenPromises.add(promise);
-    mPictureTakenOptions.put(promise, options);
-    mPictureTakenDirectories.put(promise, cacheDirectory);
+    boolean takePicture = useTakePicture();
+
     if (mPlaySoundOnCapture) {
       MediaActionSound sound = new MediaActionSound();
       sound.play(MediaActionSound.SHUTTER_CLICK);
     }
+
+    if (!takePicture && mCamera1ScanMode.equals(CAMERA1SCANSUPERFAST) && mRotated != null) {
+      resolveTakenPicture(mRotated, options, promise, cacheDirectory);
+      storeRotated(null);
+      return;
+    }
+
+    synchronized (COLLECTION_LOCK) {
+      mPictureTakenPromises.add(promise);
+      mPictureTakenOptions.put(promise, options);
+      mPictureTakenDirectories.put(promise, cacheDirectory);
+    }
     try {
-      super.takePicture(options);
+      if (takePicture) {
+        super.takePicture(options);
+      } else {
+        checkScanning();
+      }
     } catch (Exception e) {
-      mPictureTakenPromises.remove(promise);
-      mPictureTakenOptions.remove(promise);
-      mPictureTakenDirectories.remove(promise);
+      synchronized (COLLECTION_LOCK) {
+        mPictureTakenPromises.remove(promise);
+        mPictureTakenOptions.remove(promise);
+        mPictureTakenDirectories.remove(promise);
+      }
       throw e;
     }
   }
@@ -285,6 +423,23 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
   }
 
   /**
+   * Set camera1 scan mode (camera2 does not allow frame preview).
+   * - none: use native takePicture()
+   * - eco: run preview on picture request only
+   * - fast: preview each frame and process when picture is requested
+   */
+
+  public void setCamera1ScanMode(String scanMode) {
+    scanMode = scanMode.toLowerCase();
+    if (scanMode.equals(CAMERA1SCANECO) || scanMode.equals(CAMERA1SCANFAST) || scanMode.equals(CAMERA1SCANSUPERFAST)) {
+      mCamera1ScanMode = scanMode;
+    } else {
+      mCamera1ScanMode = CAMERA1SCANNONE;
+    }
+    checkScanning();
+  }
+
+  /**
    * Initialize the barcode decoder.
    * Supports all iOS codes except [code138, code39mod43, itf14]
    * Additionally supports [codabar, code128, maxicode, rss14, rssexpanded, upc_a, upc_ean]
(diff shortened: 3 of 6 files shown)


bkDJ:
There is a *world* of difference between <200ms and ~2 seconds in how a user experiences photo capture in an app. Thanks so much for the contribution! Could you please mention which device your benchmark was performed on?

Boris-c:
@bkDJ Values we’re obtained on a Nokia 7.1 Android one.

cristianoccazinsp:
This is very interesting. Is there a lot of overhead for having the preview running?

Also, was the measured time done with Camera1 implementation? Camera1 is quite fast right now, almost instant with the latest changes that removed unnecessary focus attempts.

Boris-c:
@cristianoccazinsp The overhead here comes from processing the frame preview (YUV to RGB and resizing). The overhead is minimal in eco and fast modes, bigger in boost.

The proposed implementation works with Camera1 only.

cristianoccazinsp:
@Boris-c  thanks for the update! My question about the overhead was mostly about the phone being usable (at least on average phones) while the preview is running. I have an application that overlays a bunch of stuff on top of the camera and it is critical for the UI/JS code to remain efficient even with the preview running.

About Camera1, I do understand it is Camera1 only. My question was, did you run the tests with master? Because I'm far from seeing a 1.7s captures even on cheap devices, more like 0.5-1s at most since it now captures right away what's on the preview.

I will give this branch a test if I have time later today.

Boris-c:
@cristianoccazinsp I did all my test using release 3.4.0. At what resolution do you get 0.5-1 second per capture?

cristianoccazinsp:
@Boris-c default resolution, and I'm even doing post processing with image resizing afterwards with another library.
Test comes from a Pixel 2 and Motorola moto g5. Perhaps the camera on Nokia is heavier by itself, hard to tell, but I'm definitely not getting those high numbers.

Boris-c:
@cristianoccazinsp I think no overhead can be perceived by the user in eco and fast modes

cristianoccazinsp:
I've been testing on a Moto G5 (a very, very slow device, I expect this to be much faster with prod build) with debug/dev mode on:

- 'none': 1.1 seconds in average
- 'eco': crash
- 'fast': crash
- 'boost': crash

Not sure where are the crashes coming from, did I do something wrong? All crashes are due to the same.

Crash log
```
2019-09-27 13:01:34.052 31138-31166/com.zinspector3.dev E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #2
    Process: com.zinspector3.dev, PID: 31138
    java.lang.RuntimeException: An error occurred while executing doInBackground()
        at android.os.AsyncTask$3.done(AsyncTask.java:353)
        at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:383)
        at java.util.concurrent.FutureTask.setException(FutureTask.java:252)
        at java.util.concurrent.FutureTask.run(FutureTask.java:271)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
        at java.lang.Thread.run(Thread.java:764)
     Caused by: java.lang.NullPointerException: Attempt to get length of null array
        at java.io.FileOutputStream.write(FileOutputStream.java:309)
        at org.reactnative.camera.tasks.ResolveTakenPictureAsyncTask.doInBackground(ResolveTakenPictureAsyncTask.java:75)
        at org.reactnative.camera.tasks.ResolveTakenPictureAsyncTask.doInBackground(ResolveTakenPictureAsyncTask.java:27)
        at android.os.AsyncTask$2.call(AsyncTask.java:333)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162) 
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636) 
        at java.lang.Thread.run(Thread.java:764) 
```

Boris-c:
It seems your crash happen in ResolveTakenPictureAsyncTask.java, in the section of the code that handles the option `skipProcessing`. I'm not sure because I don't know what version of that file you have. But if that's the case then try to remove `skipProcessing` from your options.

I developed that improvement last year, before that option was available. skipProcessing relies on receiving a byte array in `mImageData`, whereas my method sends a bitmap in `mBitmap`.

cristianoccazinsp:
Still getting the same crash with skipProcessing={false}.

I'm using your master branch directly from github, so I should be using exactly your code.

Boris-c:
image

Line 75 (from your crash report above) is in the section that handles skipProcessing. Are you sure you've deactivated skipProcess? Are you sure that it's the same crash?

Boris-c:
Oh, but I see, they did something wrong, they just check `mOptions.hasKey("skipProcessing")` instead of checking its value.

Try to remove the option instead of setting it to false!...

Boris-c:
It should be `mOptions.hasKey("skipProcessing") && mOptions.getBoolean("skipProcessing")` or something of the kind.

cristianoccazinsp:
You are absolutely right, that's actually a good catch/bug.

I now see where you were getting such high times, if skipProcessing is not used, the camera is definitely very slow, but it is quite fast with skipProcessing. It is really not Android's camera API fault's but rather all the processing done by the library itself.

New times, without skipProcessing:

none: 2.4 seconds (against 1.1, holy cow skipProcessing makes a huge difference)
eco: 0.9
fast: 0.7
boost: 0.5

I've noticed, however, picture quality is considerably lower with anything that's not set to "none". It is darker, blurrier, and lower resolution (a few hunder kbs lower)

Original:


Fast/Boost:



Are you willing to do fixes so it also works with skipProcessing? Have you also tried with back/front/wide lense cameras? Your code is not up to date to master so I can't test with the camera selection options.

cristianoccazinsp:
There's also another subtlety. When using any mode != 'none', on Android 10, the camera no longer makes the capture sound. I believe Android 10 is now copying iOS and any app that uses use of the camera API is forced to make a sound.

Testing on a Pixel 2 I've also got:

none / no skipProcessing: 1.1 seconds
none / skipProcessing: 1 second
boost / no skipProcessing: 0.25 seconds

Again, the picture quality is severely reduced with eco/fast/boost (mostly light and size).

I've gotta say it's a good trade off if you need very quick photos. The API might get a bit complicated with so many options (skip processing, fix orientation, scan mode).

Boris-c:
You can try

```
                if (mImageData != null) {
                    // Save byte array (it is already a JPEG)
                    fOut.write(mImageData);
                } else if (mBitmap != null) {
                    // Convert to JPEG and save
                    mBitmap.compress(Bitmap.CompressFormat.JPEG, getQuality(), fOut);
                }
```

... as a replacement of ...

```
                    // Save byte array (it is already a JPEG)
                    fOut.write(mImageData);
```

Boris-c:
As for the quality of the picture, no idea. I get very good pictures on my device.

cristianoccazinsp:
Most likely related that one mode takes the picture from the camera, and another one just captures from the preview. The preview is always lower quality than the final capture.

The change above fixes the crash with skipProcessing which is good.

Do you think you can add that fix to your branch, and rebase from master to include all the latest updates? 

I would also definitely mention the differences when using any of those scan modes. Picture quality might suffer, there will be no shutter sound performed automatically by Android, and there might be a slight overhead.

cristianoccazinsp:
Here's another "danger" of using the fast capture options:

cristianoccazinsp:
Another comparison using none and boost. You can see the boost mode pretty much skips any software processing from the camera. This is very noticeable on a pixel 2 that relies heavily on post processing.
Again, it's a very good trade off, but it should definitely be mentioned!

Original:
normal

Boost:
fast

Boris-c:
I didn't rebase, since I didn't put my changes in a branch. But I pulled from master.

Then I implemented the changes to get the image as base64 even in "skipProcessing" mode, and allowed deactivating file saving in that mode.

I also fixed the check of `skipProcessing`.

Regarding the missing shutter sound you mentioned, there is a prop, `playSoundOnCapture` that lets you require it.

Boris-c:
Last thing, we have an Android app used by hundreds of people (for research), who uploaded thousands of pictures of their food in various light conditions and with a whole range of devices, and we didn't notice any problems with picture quality in real life, using the implementation proposed in that PR.

cristianoccazinsp:
Thanks for the updates @Boris-c !. 

About picture quality, the difference is really not that significant. It becomes more obvious when using the focus feature since that makes a whole lot of changes to the lightning capture.

Again, I think the difference is not too significant, but if you test it is noticeable and will vary from device to device based on the camera features. I'm still not sure where the difference comes from, but is definitely there!

bneigher:
oh wow this is awesome

kperreau:
Very nice PR.
It works very much faster than the original.
The only things is the "**mirrorImage**" option is not working.

Boris-c:
@kperreau The `mirrorImage` option could easily be handled by patching the code.

But more generally, my personal opinion is that the set of options is a bit of a mess. I don't really get the `skipProcessing` option for instance. There should be a set of processing options, which, when all turned off, would let you "skip processing". And is getting EXIF data a processing? And so on.

Boris-c:
@kperreau I added mirrorImage handling to the PR

cristianoccazinsp:
> @kperreau The `mirrorImage` option could easily be handled by patching the code.
> 
> But more generally, my personal opinion is that the set of options is a bit of a mess. I don't really get the `skipProcessing` option for instance. There should be a set of processing options, which, when all turned off, would let you "skip processing". And is getting EXIF data a processing? And so on.

I agree with you, skip processing should be the default, and additional processing should happen with additional options. The whole processing / exif extraction logic should also be reviewed since it makes things so much slower. Changing this, however, would be difficult given how incompatible it would be with previous versions.

sibelius:
can we have processing as a string array?

so we can deal with this like as a pipeline of transformations of the raw image

Boris-c:
Especially since the library has become the community camera library of choice, I think a rewrite of the options would make sense. Even if it brings breaking changes, I don't see them as a major pain, since, from a user's perspective, it would just require readjusting the options (as opposed to a data or response format change, which might require a lot of re-coding).

But since I wasn't involved in the initial specs, and not fully understand their logic, I'm just suggesting...

By the way, I also have the same kind of improvement for iOS (not as huge though, but still quite noticeable), which I'll push to this PR soon. I should probably make another PR, but I don't have time and it's more or less the same logic.

cristianoccazinsp:
Hey @Boris-c , would you please review the following PR if you do any changes to iOS? I've made a bunch of fixes and changes to allow for better camera selection on iOS (to support wide and telephoto cameras), and discovered and fixed a bunch of bugs in the process.

Would be good to have those updates on made on top of the fixes and to also ensure they will work with all cameras: https://github.com/react-native-community/react-native-camera/pull/2514

Lastly, I wasn't able to test this PR on a wide-lense camera when using the cameraId property. Would it be there any reason for it to not work with this?

cristianoccazinsp:
> can we have processing as a string array?
> 
> so we can deal with this like as a pipeline of transformations of the raw image

That would probably work. The way the different settings work are also a bit confusing for me, but I would definitely help with it.
There's also the issue about some secret/undocumented settings (e.g., `pictureSize` for iOS) that probably could also use some help. Not to mention the unused code/references to the old RTC code that should also probably be removed, some cleanup here and there overall.

Boris-c:
@cristianoccazinsp I don't think my changes could affect/be affected by your changes, but that might require a deeper check. I'll give more details about my changes when I push them to this PR.

cristianoccazinsp:
@sibelius  it is not crashing.

However, I was taking a second look, would the default be "eco"? Does that mean that anyone updating will be migrated to this new capture behaviour? (https://github.com/react-native-community/react-native-camera/pull/2505/files#diff-6ee136227b7a6fdd96b7a213d5648f4eR461)

I don't think this should be default given all the caveats capturing like this can have.

Boris-c:
@cristianoccazinsp In our experience, with thousands of pictures taken by hundreds of Android users, the caveats did not show (users take pictures of their food, in everyday life conditions, little light, etc.).

If I was to decide, given the huge speed improvement and the minor real life risk of poorer quality pictures, I would leave "eco" as a default and would explain in the docs that if the best picture "quality insurance" is needed, they should use "none" instead, i.e. no speed optimization.

But that's just my humble opinion and humble experience.

cristianoccazinsp:
Perhaps I need to test on more devices, but looking at the pictures above, and when using focus and white balance settings, the quality difference is quite noticeable on a Pixel 2 (which is known for having a very good camera with a lot of software processing). I would definitely not like that my pictures end up with lower quality after an upgrade. If anything, the update notes should have a big warning about it.

On the app we're developing, lightning and quality is extremely important for the captured pictures since we use it to capture walls, paintings, floors, etc. So if we look at the pictures above, the one taken with "boost" definitely loses a lot of detail in regards to the lightning on the top right of the photo.

sibelius:
What is the best default to keep the current behavior?

Boris-c:
By the way, I have a similar improvement nearly ready for iOS. On my iPhone X, capture takes ~110ms instead of ~800.

cristianoccazinsp:
@sibelius in my opinion the default should be 'none' just so no one gets surprised when upgrading, but I can be convinced otherwise as long as it's is stated in the changelog/release notes.

@Boris-c I'm eager to see those changes on iOS as well! Also, can you make sure to update the types file (types/index.d.ts) so the new settings are documented when using an IDE?

cristianoccazinsp:
Here's another comparison on a moto G5, you can see the difference in lightning. Dark areas are pretty much black when using eco mode. While the original is able to keep a good enough contrast to show both light and dark areas. Both pictures where taken by focusing on the top right (white) corner. This also shows that it is not a subtle difference due to focus/exposure, because the white top right area looks exactly the same.

I know I'm looking at very fine (and most likely not important) details, but just want to get the point that pictures are indeed different!

orig vs eco

Boris-c:
It was already obvious from your previous pictures that some picture adjustment/postprocessing is skipped when capturing the preview. But in our case, speed is much more important than picture best-of-the-best quality (which is the reason why I investing time working on this aspect).

sibelius:
Can we merge and iterate from master?

Boris-c:
I just added the iOS equivalent of Android's "fast" mode. I made them 2 different props, since one might want the Android improvement only, for instance, and the modes are not exactly the same (none, eco, fast and boost on Android, none and fast on iOS), but you might prefer a unified setting.

cristianoccazinsp:
Do you know if telephoto and ultra wide cameras will also work with this mode? Those two cameras are very strict and most of the time crash when given unsupported settings.

Also, would this work when switching cameras? For instance, some settings are reset on camera / preset switching and must be re-enforced on the native side. Other cases, the camera code might just crash if the device is changed while the session has some unsupported settings. 

Have you observed the same quality changes as Android?

I will give these a try later today, but sadly I do not have an iPhone 11 for local development so it might take some time until it is tested by the other team members.

Boris-c:
I don't have experience with telephoto and ultra wide cameras, nor did I test the front/back camera switch.

But the code mainly hooks in `- (void)captureOutput: didOutputSampleBuffer: fromConnection:`, which was there already, taking care of text recognition, face recognition and so forth. So my guess is that everything that worked with those will work with my code.

I didn't do a side by side comparison like you did on Android. But I'll do one now and post it.

Boris-c:
Apparently, the same thing happens as on Android, i.e. the lack of post-processing is visible. It's currently very dark in my home, and the difference is noticeable when the picture is zoomed.

The good news is that the similarity on both platforms will make it easier for the documentation ;)

cristianoccazinsp:
That's good to know. Do you have any timing samples? Remember to use skipProcessing as well since that flag makes a huge difference.

cristianoccazinsp:
Hey @Boris-c , I've found a few issues while testing on an iPhone 7.

- "failing real check" is constantly written to the console output. NSLog statements will be written even when built for production. This can cause some performance problems.

> 2019-10-10 17:27:53.403369-0300 tmi3[1200:461242] failing real check
> 2019-10-10 17:27:53.438288-0300 tmi3[1200:461242] failing real check
> .... x 100

- Flash does not work, I'm guessing this is an obvious caveat when using such mode. This applies to Android as well.

- There's a black screen for half a second the first time a picture is taken, have you noticed that? If not, I can try making a video of it. I'm not sure what's causing it exactly, but it causes the first picture after the camera is mounted to be nearly full black. This can happen easily also when switching from back to front camera and vice versa, the first picture always comes up fully black. When using the front camera, the first picture is pretty much 100% black, but there's some image when using the back camera. See sample pictures below

- An "Already capturing" error can leave the camera stuck (remount is needed) after switching from back to front cameras multiple times, recording video, or changing the cameraScanMode property.

- Lastly, I'm guessing this is to the poor iphone 7 camera, but I really didn't notice a difference in photo quality. The only noticeable difference is that the shutter sound is not forced as when using the actual capture API.


Black photos on the first picture that is taken. Back and front cameras:
46967BD4-EE7F-4605-82F6-484415CEBEEB
33EEAFCA-CFE2-42FE-9BBB-0E5D34E9C904

Boris-c:
Hi @cristianoccazinsp, here are my comments:
- "failing real check". I did not implement a switching off of the session output of the sample buffer to the delegate, because as a whole, this is weirdly implemented for other functionality (text, face and barcode detectors), and I didn't need that (I know it should be there though!). Take the following code:
```
- (void)setupOrDisableTextDetector
{
    if ([self canReadText] && [self.textDetector isRealDetector]){
        if (!self.videoDataOutput) {
            self.videoDataOutput = [[AVCaptureVideoDataOutput alloc] init];
            if (![self.session canAddOutput:_videoDataOutput]) {
                NSLog(@"Failed to setup video data output");
                [self stopTextRecognition];
                return;
            }
            NSDictionary *rgbOutputSettings = [NSDictionary
                dictionaryWithObject:[NSNumber numberWithInt:kCMPixelFormat_32BGRA]
                                forKey:(id)kCVPixelBufferPixelFormatTypeKey];
            [self.videoDataOutput setVideoSettings:rgbOutputSettings];
            [self.videoDataOutput setAlwaysDiscardsLateVideoFrames:YES];
            [self.videoDataOutput setSampleBufferDelegate:self queue:self.sessionQueue];
            [self.session addOutput:_videoDataOutput];
        }
    } else {
        [self stopTextRecognition];
    }
}

- (void)stopTextRecognition
{
    if (self.videoDataOutput && !self.canDetectFaces) {
        [self.session removeOutput:self.videoDataOutput];
    }
    self.videoDataOutput = nil;
}
```

In `setupOrDisableTextDetector`, `stopTextRecognition` is called if we don't need to detect text or fail at adding the output, which in turn removes the video data output, if it exists, depending on wether we need to detect faces too. This is wrong, and the same logic appears in different places, with a lot of duplicated code. For instance, disabling text detection could make barcode detection (or my picture capture method) fail.

Should I rewrite the logic here, I'd rather have a function that checks the need for the video data output, does the job of setting up a new one or removing an existing one, and call it from different places. The code that logs "failing real check" probably shouldn't be there in the first place, since didOutputSampleBuffer should never be called unless we want it to be called.

- Flash not working: that's no surprise I guess. No idea if it can force programmatically turned on.

- Black screen: never had that, neither on Android nor on iOS

- "Already capturing": might be fixed when the first issue is fixed.

- Picture quality and shutter sound: There is definitely a small difference in quality on iPhone X. Shutter sound fixed in latest push.

cristianoccazinsp:
Hi @Boris-c ,

I'm not sure what's the best way to fix the "failing real check" issue. I personally don't quite understand the logic behind text/barcode recognition neither nor why those checks are needed. It would be great if you could fix it.

The black screen, really happens all the time. Can you try to reproduce it? Make sure you're using all the camera features (camera switching, zoom, white balance, touch to focus, video recording, etc.). I'm also using `skipProcessing: true` and `fixOrientation: false`.  I'm also not setting `defaultVideoQuality` nor `pictureSize`. The black screen looks very similar to the flicker that happens when recording video due to preset changes.

cristianoccazinsp:
@Boris-c sorry for the spam!.

I've taken a video that shows the issue of a black screen. Now that you've added sound, it is even stranger! Looks like the picture is taken twice, but it is not the code or me taking the picture twice since you can see only one picture is added, and the same code captures only once with "none" mode. Even stranger, **the issue doesn't happen when in landscape mode!**
You can see a video of it in action below.

"fast mode" on:
https://zinspectordev2.s3.amazonaws.com/usi/2/1570805547eec281ab6bfc4643bd1f40cd68f30714.mp4

"none" mode:
https://zinspectordev2.s3.amazonaws.com/usi/2/1570805627b9212c7925cc4ebf8e3256729b14ea19.mp4

Second issue. The pictures taken with this mode do not seem to be using the capture preset configuration very well. On the videos above, I'm also testing capturing a photo with the default settings, but also with the "High" `pictureSize` preset, which is the one used to capture video.
As you can see from the photos below, fast and none mode yields totally different picture sizes. Looks like "fast" does not capture exactly what's on the preview.


"fast mode" / default and "High" pictureSize (look at the lamp):
15708055091ff0ecac22cb48af813db682727d3d7f
157080552063207ab5cb324a21aeb7ed2558478571

"none mode" / default and high pictureSize:

15708056008d6632e70a084bc1b969f1c1161f23fc
1570805611639f9beaf7504acd91f8470bf322eccc


RN Camera code that was used:

```
<RNCamera
  ref={ref => {
    this.camera = ref;
  }}
  style={cameraStyle}
  cameraId={cameraId} // set to back camera after onStatusChange
  camera1ScanMode={this.props.settings.photoFastMode ? 'eco' : 'none'}
  cameraScanMode={this.props.settings.photoFastMode ? 'fast' : 'none'}
  pictureSize={this.state.pictureSize} // testing using "Photo" (defaukt) and "High" (video)
  ratio={this.state.aspectRatioStr} // used 4:3 while testing
  flashMode={flashMode} // none
  zoom={zoom} // 0
  maxZoom={MAX_ZOOM} // 8
  whiteBalance={WB_OPTIONS[wb]}
  autoFocusPointOfInterest={this.state.focusCoords} // undefined / none set
  onStatusChange={this.onCameraStatusChange}
  pendingAuthorizationView={
    <SafeAreaView style={styles.cameraLoading}>
      <Spinner color={style.brandLight}/>
    </SafeAreaView>
  }
  notAuthorizedView={
    <View>
      {cameraNotAuthorized}
    </View>
  }
>
  ...
</RNCamera>
```

Boris-c:
@cristianoccazinsp I think I also fixed the dark image issue, that was probably caused by the call to set video orientation (that triggers a light, white balance and focus check by the camera apparently).

I also merged 3.8.0

cristianoccazinsp:
@Boris-c  were you also able to review the other issue related to image cropping? Do you think the time difference might be really just that the original take picture does that image cropping first? If you were to add that to the fast mode, would it still be faster? I'm not sure what's the best to do there, but I'm guessing that both fast and none modes should take identical pictures.

cristianoccazinsp:
@Boris-c  Also, do you have an idea if these changes for Android might cause issues if the run/stop steps are moved to non UI threads? I'm testing some changes to improve the camera startup time and deal with ANRs when coming from background here: https://github.com/react-native-community/react-native-camera/compare/master...cristianoccazinsp:android-ui-thread?expand=1

Lastly, I'm also trying to test issues on iOS related to audio interruptions, do you think these changes might break when the session is updated due to audio interruptions? Changes here: https://github.com/react-native-community/react-native-camera/compare/master...cristianoccazinsp:ios-interruptions?expand=1

Boris-c:
I don't think the time difference comes from images not being cropped. My guess is that something bad is happening in Android's public camera API (since the Camera app itself performs much faster than `takePicture()`).

Avoiding the cropping of the preview frames to what the preview itself is showing is, in my opinion, impossible.

Regarding audio interruptions, no idea!...

cristianoccazinsp:
The cropping is for iOS, not android.

For Android, I think the capture speed is the same for the native camera
apps. The difference is that RN code waits for the final picture to be
returned, while native apps use all the 3 or 4 events provided by by
android and release the camera for a new picture as soon as the first event
fires, and let's the rest of the processing to work in background. At least
that's how the one from Google works.

Boris-c:
About Android, I was referring to the speed, in the sense that the huge time it takes to get the result for takePicture() is probably caused by some implementation peculiarity on Android’s side, not to the picture cropping.

And when I mean the Camera app, I mean the app provided with the system, which takes pictures much faster than what we can do (in native apps and RN apps, since the bottleneck is the native call to takePicture(), in RN as we’ll ).

Some while ago I benchmarked the whole process in detail, from RN’s capture call to the callback in RN. takePicture() is the bad guy in the chain. That’s how I came to the idea of capturing the frame preview.

nguyenhose:
is anyone in charge on this PR ? this also happened for my app now

sibelius:
does this solves this issue https://github.com/react-native-community/react-native-camera/pull/2577?

cristianoccazinsp:
> does this solves this issue #2577?

I believe that PR will make things a bit easier (since there's no skip processing to worry about, it was in fact @Boris-c  who suggested removing it), but this PR still makes capture much faster on Android since it skips the take picture calls altogether and uses a much different approach.

Still think this would be a very good addition, as long as it is not the default behaviour due to all the extra caveats it might add.

kperreau:
It would be nice if it could be merged into the master.

ShivanshJ:
Why is this not being merged? It's been over an year.

sibelius:
we need to fix conflicts

and have more people testing this

rafaelcavalcante:
@sibelius are there any discussions outside this pr for what should be tested before it goes do master? I'm willing to help.

louisobasohan:
looks like there are conflicts

sibelius:
the first thing would be to create a new PR without conflicts

pranatha:
Hi @Boris-c , why the width and height result of takePictureAsync in your version 
are different from the current version?
I use @pontusab/react-native-image-manipulator for cropping picture,
somehow for certain device, the cropping process failed by using your version because of the image size inconsistence.

Issue #1772, opened 2018-09-01
Too much zoom on iOS 


varburg:
### Which implementation are you using

*RNCamera*
 
Only on iOS
there have strange behaviour when I add zoom to RNCamera component.
The behaviour is something like "super zoom" i.e. if max zoom is 1x I see 10x.

react native version 0.56.0
xcode Version 9.4.1 (9F2000)

Code was copied from RNCamera docs. Just added zoom.
Can somebody guide me what is wrong in this code:
```js
'use strict';
import React, { Component } from 'react';
import {
  AppRegistry,
  Dimensions,
  StyleSheet,
  Text,
  TouchableOpacity,
  View
} from 'react-native';
import { RNCamera } from 'react-native-camera';

class BadInstagramCloneApp extends Component {
  render() {
    const zmm = 0.5;
    return (
      <View style={styles.container}>
        <RNCamera
            ref={ref => {
              this.camera = ref;
            }}
            zoom = {zmm}
            style = {styles.preview}
            type={RNCamera.Constants.Type.back}
            flashMode={RNCamera.Constants.FlashMode.off}
            permissionDialogTitle={'Permission to use camera'}
            permissionDialogMessage={'We need your permission to use your camera phone'}
        />
        <View style={{flex: 0, flexDirection: 'row', justifyContent: 'center',}}>
        <TouchableOpacity
            onPress={this.takePicture.bind(this)}
            style = {styles.capture}
        >
            <Text style={{fontSize: 14}}> SNAP </Text>
        </TouchableOpacity>
        </View>
      </View>
    );
  }

  takePicture = async function() {
    if (this.camera) {
      const options = { quality: 0.5, base64: true };
      const data = await this.camera.takePictureAsync(options)
      console.log(data.uri);
    }
  };
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    flexDirection: 'column',
    backgroundColor: 'black'
  },
  preview: {
    flex: 1,
    justifyContent: 'flex-end',
    alignItems: 'center'
  },
  capture: {
    flex: 0,
    backgroundColor: '#fff',
    borderRadius: 5,
    padding: 15,
    paddingHorizontal: 20,
    alignSelf: 'center',
    margin: 20
  }
});

AppRegistry.registerComponent('BadInstagramCloneApp', () => BadInstagramCloneApp);
```

victorileaf:
Same issue here as well 👍

aseem1992:
+1

[...]

PR #2510, opened 2019-09-26
Add cameraId feature to iOS and various bug fixes

cristianoccazinsp:
# Summary

Hello, this PR is mostly to add the cameraId feature from Android to iOS as well in order to be able to select from various cameras. While doing this, I've also found a bunch of other bugs that were also fixed. 

In addition, a second commit tries to fix the zoom issue by adding a new iOS only property: maxZoom. This property can be used to restrict the max zoom calculations.

Please review carefully and let me know of any possible issues. I'm currently using this code on a live app with a good amount of users so I will report back any issues soon.

**There's one "issue" I wasn't able to figure out why it is happening** (and it is not related to this change). The camera view instance seems to be created twice the first time it is used, but only one view is mounted/used. The reason of why RN is creating the view twice is still unknown to me, but the changes to event subscription should improve this in a way that events are not fired if not needed. **I would really appreciate some review from someone else to figure out why this happens, the easy to test is to simply add a log to the RNCamera.m initWithBridge method and observe how it is called twice.**. Fixing this would reduce some additional redundant code that is called and makes the view stay in memory.

Summary:

- Implement getCameraIds and cameraId property to manually select a camera device. This was previously implemented for Android and now added to iOS as well

- Adds a maxZoom property to control extraneous zoom values on iOS. Should help with https://github.com/react-native-community/react-native-camera/issues/1772

- Fix the internal preset being used to properly use a photo and video preset, and check for invalid presets on camera switch.
    - This fixes a bug that would happen after recording. After a recording is complete, the camera preset was set to 4k permanently, and attempting to select any camera that does not support that preset would result in an totally unusable camera. Now, the "Photo" preset is the default which should be the highest quality option for photos, and the previously used "High" preset will be used as video default if no value is provided.
    - This also adds proper support to setting the undocumented "pictureSize" property. Setting this property will change the default photo quality to be that one.

- Minor cleanup to the camera initialization so the same code is not called as many times redundantly (start session / device happening multiple times)
    - The reduced redundant calls to start session and session updates should make the camera startup slightly faster.

- Fix for warnings issued due to misuse of the "isRecording" property. This property was defined as both as a property and a method and was both getting set manually and retrieved with a function. It should now consistently only be retrieved with a function based on the real recording status. Should have no behaivour change, but removes a pesky warning.

- Fix an issue related to the camera getting stuck after a background resume by removing the use of the queue for those two events.

- Fixes a wrong event unsubscription (UIDeviceOrientationDidChangeNotification to UIApplicationDidChangeStatusBarOrientationNotification) from a previous change. Also move event subscriptions to superview changes for consistency and so they are not used/consumed if not needed. After testing, the View of RNCamera is instantiated twice (for some reason I couldn't find) but used only once. RN keeps a reference to the view in memory even after the camera is destroyed, resulting in these events getting fired and handled all the time during the app's lifetime.

## Test Plan

Tested on iPhone 7, iphone X, and iPad mini.

-

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist


cristianoccazinsp:
Note that I've also have another update for this to support ultra wide lenses (iOS 13). However, I haven't included it yet because it would need people to use XCode with the SDK for iOS 13 which is quite new. This is because the new ultra wide lense variable is available only on the new SDK:

https://github.com/cristianoccazinsp/react-native-camera/commit/44c7ee8ec0a3568b3932642d18b423a72a6e430e

scandinaro:
@cristianoccazinsp once you get the list of camera IDs for iOS how do you set the camera by id?

scandinaro:
>@cristianoccazinsp once you get the list of camera IDs for iOS how do you set the camera by id?

got it RNCamera prop `cameraId`

Tibfib:
I've been battling a lot of instability issues with react-native-camera (failures to take the picture, crashes) on iOS for the past week and decided to give @cristianoccazinsp's fork a chance and it's much improved. Hoping the maintainers have time to look at this in the near future and get it merged.

sibelius:
is `cameraId`required after this PR?

Tibfib:
> is `cameraId`required after this PR?

No. I haven't made any changes beside swapping out the package.

cristianoccazinsp:
It is not. cameraId set to null/undefined (default) will just do nothing
and use the type prop instead to select the camera as before. It is only
intended for advanced use.

PR #2514, opened 2019-09-27
Ios exposure poi

cristianoccazinsp:
# Summary

Taken from https://github.com/react-native-community/react-native-camera/pull/2423 and resolved conflicts. Also made the auto exposure feature to be optional with an additional iOS only flag on autoFocusPointOfInterest.

This PR builds on top of the previous PR (https://github.com/react-native-community/react-native-camera/pull/2510) that fixes several iOS bugs.

PS: I'm not sure if a PR on top of a PR can really work or not. Just in case, here's the summary of the previous PR.

--
Hello, this PR is mostly to add the cameraId feature from Android to iOS as well in order to be able to select from various cameras. While doing this, I've also found a bunch of other bugs that were also fixed.

In addition, a second commit tries to fix the zoom issue by adding a new iOS only property: maxZoom. This property can be used to restrict the max zoom calculations.

Please review carefully and let me know of any possible issues. I'm currently using this code on a live app with a good amount of users so I will report back any issues soon.

There's one "issue" I wasn't able to figure out why it is happening (and it is not related to this change). The camera view instance seems to be created twice the first time it is used, but only one view is mounted/used. The reason of why RN is creating the view twice is still unknown to me, but the changes to event subscription should improve this in a way that events are not fired if not needed. I would really appreciate some review from someone else to figure out why this happens, the easy to test is to simply add a log to the RNCamera.m initWithBridge method and observe how it is called twice.. Fixing this would reduce some additional redundant code that is called and makes the view stay in memory.

Summary:

- Implement getCameraIds and cameraId property to manually select a camera device. This was previously implemented for Android and now added to iOS as well
- Adds a maxZoom property to control extraneous zoom values on iOS. Should help with #1772
- Fix the internal preset being used to properly use a photo and video preset, and check for invalid presets on camera switch.
- This fixes a bug that would happen after recording. After a recording is complete, the camera preset was set to 4k permanently, and attempting to select any camera that does not support that preset would result in an totally unusable camera. Now, the "Photo" preset is the default which should be the highest quality option for photos, and the previously used "High" preset will be used as video default if no value is provided.
- This also adds proper support to setting the undocumented "pictureSize" property. Setting this property will change the default photo quality to be that one.
- Minor cleanup to the camera initialization so the same code is not called as many times redundantly (start session / device happening multiple times)
- The reduced redundant calls to start session and session updates should make the camera startup slightly faster.
- Fix for warnings issued due to misuse of the "isRecording" property. This property was defined as both as a property and a method and was both getting set manually and retrieved with a function. It should now consistently only be retrieved with a function based on the real recording status. Should have no behaivour change, but removes a pesky warning.
- Fix an issue related to the camera getting stuck after a background resume by removing the use of the queue for those two events.
- Fixes a wrong event unsubscription (UIDeviceOrientationDidChangeNotification to UIApplicationDidChangeStatusBarOrientationNotification) from a previous change. Also moved event subscriptions to superview changes for consistency and so they are not used/consumed if not needed. After testing, the View of RNCamera is instantiated twice (for some reason I couldn't find) but used only once. RN keeps a reference to the view in memory even after the camera is destroyed, resulting in these events getting fired and handled all the time during the app's lifetime.

--

## Test Plan
Test on iPhone 7

-

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist


cristianoccazinsp:
Can anyone help me or guide on how I can get all these changes into individual PRs? Just figured out it included all changes from the previous PR as well.

sibelius:
we need to merge the other  PRs first, and then merge this one

cristianoccazinsp:
Got it. Apologies for the spam, but things are moving quite fast here.

scandinaro:
@cristianoccazinsp once you get the list of camera IDs for iOS how do you set the camera by id?

cristianoccazinsp:
Just like Android, there's a `cameraId` prop that can be used and which will override the back/front value.

cristianoccazinsp:
Any ideas how we can make code that compiles fine for older SDKs? The new telephoto and ultra wide angle camera constants are only available on the latest SDKs, which means I cannot include some more updates for the new iphones without requiring an SDK upgrade.

sibelius:
what would be the min version?

can't we use pragma for this?

cristianoccazinsp:
Note that this is not for IOS versions (there's code for that as recommended by Apple already), but rather XCode SDK version. The new constants (AVCaptureDeviceTypeBuiltInUltraWideCamera, AVCaptureDeviceTypeBuiltInTripleCamera, AVCaptureDeviceTypeBuiltInTelephotoCamera, and the property isVirtualDevice) are only available on the latest XCode version.

sibelius:
@n1ru4l what do you think about this?

can we force users to use latest xcode version?

cristianoccazinsp:
Not sure if something like this is the preferred way of doing such sdk compile based decision:

```
if (@available(iOS 13.0, *)) {
    captureDeviceType = @[
        AVCaptureDeviceTypeBuiltInWideAngleCamera,
        AVCaptureDeviceTypeBuiltInTelephotoCamera
        #if __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
            ,AVCaptureDeviceTypeBuiltInUltraWideCamera
        #endif
    ];
}
else{
    captureDeviceType = @[AVCaptureDeviceTypeBuiltInWideAngleCamera, AVCaptureDeviceTypeBuiltInTelephotoCamera];
}
```

n1ru4l:
please keep it backward compatible 😕

cristianoccazinsp:
I think the above macro should keep compile issues due to old SDKs away.

cristianoccazinsp:
I've just pushed a bunch of other commits as a result of a lot of testing on Iphone X and 11 and camera switching. Iphone 11 Pro Max has a total of 4 cameras ( 3 on the back, 1 on the front). These latest updates show allow for consistently switching across all of them.

Some fixes also include a crash on white balance change if it is unsupported, missing camera initialized checks, etc. 

The macro should also make the code compatible with the previous SDK that lacks the ultra wide camera constant.

Still getting some more field testing and feedback but everything seems to be working fine.

cristianoccazinsp:
@sibelius yes. Give me a couple of minutes. This PR should be the final changeset since it includes all the required stuff for ios 13 wide lense support and a few other bug fixes that arised when testing iPhone's 11 ultra wide camera.

cristianoccazinsp:
Here's a small sample of what can be achieved with this on an iPhone 11 Pro. Enumerate all cameras, get their type, and switch between selfie, regular, telephoto (zoomed), and ultra wide. With the extra error checks there are also no crashes when trying to use unsupported features with the telephoto/ultrawide cameras (white balance is quite limited for the ultra wide camera, same goes for focus).





Sadly for Android, we cannot still get the camera type (ultra wide, telephoto, etc.) like we do with iOS until Camera2 is the only supported implementation.

sibelius:
can't we fix this for camera2 only?

PR #2526, opened 2019-10-09
Add more error checking to Android Camera1

cristianoccazinsp:
# Summary
Add synchronized checks on some code that might cause issues if called concurrently.
Add try catch on setDisplayOrientation

Tested on Pixel 2 (Android 10), Motorola G5 (Android 8.1).

Note: Awaiting from more field testing from a bunch of other devices since I wasn't able to reproduce the issue myself.

-

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist


cristianoccazinsp:
Adding this PR for some quick review. Do the changes look good?

Basically, it synchronizes a few methods that might be called from different threads, and adds a missing try/catch to the setDisplayOrientation call.

**PR Only for review**, I need at least a week to get enough crash logs from multiple devices to confirm the issue is resolved.  Will report back with findings.

Issue #2538, opened 2019-10-12
Video is not recording warning


igor-lemon:
If I start to record video and after fast stop recording video I have `Video is not recording.` warning. It's strange because video recording continues. 

How to reproduce
1) Set `maxDuration`
2) Start recording video
3) Fast stop recording video

igor-lemon:
Very interesting. So I found that video start recording after ~1-1.5 sec after call `recordAsync` method. How can I catch the start recording event? I see only 1 way, it's call `isRecording` method. But
1) It method only for iOS platform
2) Need to call this method

I think good way to add some start video recording callback

cristianoccazinsp:
I wonder if moving the stop recording call to the session queue would solve the issue (https://github.com/react-native-community/react-native-camera/blob/master/ios/RN/RNCamera.m#L876) 

That would mean that stop recording gets enqueued rather than stopping immediately which might end up causing other issues.

Something like:

```
dispatch_async(self.sessionQueue, ^{
    if ([self.movieFileOutput isRecording]) {
        [self.movieFileOutput stopRecording];
    } else {
        RCTLogWarn(@"Video is not recording.");
    }
})
```

Is this something you can test? Should be pretty straight forward to update from XCode.

igor-lemon:
@n1ru4l It didn't solve anything. The issue is still actual.

cristianoccazinsp:
Can you tell us what are you trying to do exactly? Video should definitely be stopping. I've added even further checks to stop video here (https://github.com/react-native-community/react-native-camera/pull/2542) in case you are stopping it before it even starts to record. Can you try that?

igor-lemon:
LOL. I don't want to upset but nope, the issue is still actual.
![image](https://user-images.githubusercontent.com/14764799/67161533-14a2bf00-f364-11e9-835e-deaf5ac6a317.png)

cristianoccazinsp:
But does the video stop with 1 call? You will still get the message if you
call it more than once while the video setup process is happening, but it
should definitely be stopping the video.

igor-lemon:
Yes. I started to record a video and stopped after 1 sec.
I didn't do any actions with the camera during recording time.
If I record video more than 1 sec the problem does not reproduce.

cristianoccazinsp:
If it stops, then there's really nothing to worry about (it would
previously not stop, so you would need more than 1 call). The video setup
is slow ( especially if the capture preset or quality needs to be updated),
[...]

Issue #2533, opened 2019-10-11
The camera view flickers after video start recording.


igor-lemon:
The camera view flickers after video start recording. Added `defaultVideoQuality` param, it didn't solve the issue.

cristianoccazinsp:
Is this on iOS? I've noticed this as well, although I haven't really given it too much importance. This is mostly related to the photo to video switch, and I guess why most native camera apps have two completely different previews for video and photo. Not entirely sure how this can be improved, adding a slight delay between configuration changes and record start maybe?

If this is Android, the reason is similar, but with the addition that some code runs on the UI thread so under cheap devices the "freeze" might be noticeable.

cristianoccazinsp:
If you set the prop `pictureSize` to match either the `quality` prop of `recordAsync` or `defaultVideoQuality`, you won't have a flicker when the video recording starts (except maybe the first time after a camera mount, I don't know why). However, you will end up also capturing lower quality/resolution photos.

cristianoccazinsp:
@sibelius maybe you have some more ideas about this. 

The flicker/first frame of the video being darker issue seems more obvious after one of the fixes I did that would set the photo preset back to normal after each video recording. Before that fix, the camera preset could remain at a value different than the default one (e.g., 4k), causing subsequent records to not flicker (because recording and photo would use the same preset). However, that was also leaving the camera in a bad state where switching to the front camera would then fail if it did not support that preset, and subsequent pictures would be taken with a preset that was not "Photo", possibly resulting in a different quality and even aspect ratio.

The flicker of the preview can't really be prevented if photo and video are captured with different presets, but it would be great for the flicker to not end up as part of the first frame of the recorded video. I believe the current behaviour matches as close as possible the native camera app (Photo preset for Photos, video preset for videos), but currently has this issue.

Maybe a tiny delay before the preset change to the record start call could help?

Related discussion on SO:
https://stackoverflow.com/questions/39229477/avcapturesession-captures-black-dark-frames-after-changing-preset


Related issues: 
https://github.com/react-native-community/react-native-camera/issues/1934
https://github.com/react-native-community/react-native-camera/pull/1886/files

nachumFreedman:
add this to the camera component tag made the flicker shorted. 

defaultVideoQuality={RNCamera.Constants.VideoQuality["720"]}

igor-lemon:
@nachumFreedman It's maybe strange, but with 720p screen flickering longer than with 1080p

[...]

PR #2539, opened 2019-10-12
Remove flickering due to audio input

cristianoccazinsp:
# Summary
Adding/removing audio input would cause the video recording to flicker. The current code tries to prevent it by only doing this if necessary, but it will always flicker the first time a record is done (because the audio input gets added there). With these changes, the audio input is always available (just like the video input), but it is turned on/off through the connection. This also has the advantage that audio could potentially be muted/unmuted on demand even while recording, but that's for another PR.

Recording will also batch all operations in a configure/commit block in order to make the change faster the internal block does configuration changes.

Lastly, stop recording is called in the same session queue in order to prevent an issue when stopping too early.

Should fix https://github.com/react-native-community/react-native-camera/issues/2538
Should improve https://github.com/react-native-community/react-native-camera/issues/2533 . However, there will still be flickering there's a preset change when recording (e.g., video quality is different than picture quality)

## Test Plan
Tested on iPhone 7

Real device

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -764,6 +764,13 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
     }
 
     NSInteger orientation = [options[@"orientation"] integerValue];
+    
+    // some operations will change our config
+    // so we batch config updates, even if inner calls
+    // might also call this, only the outermost commit will take effect
+    // making the camera changes much faster.
+    [self.session beginConfiguration];
+    
 
     if (_movieFileOutput == nil) {
         // At the time of writing AVCaptureMovieFileOutput and AVCaptureVideoDataOutput (> GMVDataOutput)
@@ -782,6 +789,7 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
     }
 
     if (self.movieFileOutput == nil || self.movieFileOutput.isRecording || _videoRecordedResolve != nil || _videoRecordedReject != nil) {
+        [self.session commitConfiguration];
       return;
     }
 
@@ -807,11 +815,6 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
         if (self.session.sessionPreset != newQuality) {
             [self updateSessionPreset:newQuality];
         }
-    }
-
-    // only update audio session when mute is not set or set to false, because otherwise there will be a flickering
-    if ([options valueForKey:@"mute"] == nil || ([options valueForKey:@"mute"] != nil && ![options[@"mute"] boolValue])) {
-        [self updateSessionAudioIsMuted:NO];
     }
 
     AVCaptureConnection *connection = [self.movieFileOutput connectionWithMediaType:AVMediaTypeVideo];
@@ -846,6 +849,16 @@ static NSDictionary *defaultFaceDetectorOptions = nil;
             }
         }
     }
+    
+    // sound recording connection, we can easily turn it on/off without manipulating inputs, this prevents flickering.
+    AVCaptureConnection *audioConnection = [self.movieFileOutput connectionWithMediaType:AVMediaTypeAudio];
+
+    if ([options valueForKey:@"mute"] == nil || ([options valueForKey:@"mute"] != nil && ![options[@"mute"] boolValue])) {
+        audioConnection.enabled = YES;
+    }
+    else{
+        audioConnection.enabled = NO;
+    }
 
     dispatch_async(self.sessionQueue, ^{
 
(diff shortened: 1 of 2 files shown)


cristianoccazinsp:
I've got one more update to push it here to also remove any flickering that might end up on the final video by just adding a small delay before starting the recording. But still testing. Combining these two video recording should be quite smooth even if the video and photo presets do not match.

ios/RN/RNCamera.m
```diff
@@ -863,21 +876,44 @@ - (void)record:(NSDictionary *)options resolve:(RCTPromiseResolveBlock)resolve r
                 [connection setVideoMirrored:YES];
             }
         }
+        
+        // finally, commit our config changes before starting to record
+        [self.session commitConfiguration];
+        
+        // after everything is set, start recording with a tiny delay
+        // to ensure the camera already has focus and exposure set.
+        double delayInSeconds = 0.5;
```
sibelius: is this a magic number?

cristianoccazinsp:
Kind of. Older devices like the iPhone 6 might need more time, but 0.5
seems to be both imperceptible to the user and high enough to account for
most devices. Should we make this variable or another parameter on the
options?

sibelius:
good magic number for now, we can customize if someone needs it

mars-lan:
Looks like this led to a request of microphone permission for apps that only need still image captures. Is it possible to avoid that?

cristianoccazinsp:
You're absolutely right! Although the same happens (and has happened before this) with Android, so a change for this should probably be applied to both. 

The recommendation from the discussion here https://github.com/react-native-community/react-native-camera/issues/1934 about using a defaultMuted prop is probably a good idea to ensure that the microphone is not used unless needed. However, "turning on" the microphone would then cause the video to flicker again the first time it is used.

MichielDeMey:
@mars-lan Indeed, this just caused our app (using react-native-qrcode-scanner) to crash on a release build due to missing microphone permissions. 🙈

cristianoccazinsp:
Ouch, I'm very sorry for that.

sibelius:
Open an issue and let’s fix this

cristianoccazinsp:
Issue here: https://github.com/react-native-community/react-native-camera/issues/2541

Looks like we don't need a `defaultMuted` property after all, there's already a `captureAudio` one for it which we should be using.

cristianoccazinsp:
@Michaelvilleneuve  can you confirm the app actually crashed? I've been testing, and even rejecting the audio permission, the app will not crash, it will just silently ignore / do not record any audio (just like the Android version). I did notice that the audio permission request does not go away even with `captureAudio` false.

MichielDeMey:
The app crashes due to missing `NSMicrophoneUsageDescription` key in the `info.plist` file which should not be required if you set `captureAudio` to false.

Right now it will offer to capture the microphone, regardless of the `captureAudio` setting.
(Which is set to `false` in the `react-native-qrcode-scanner` module)

igor-lemon:
That doesn't fix anything. The camera is still flickering...

igor-lemon:
The recorded screen video inside
[RPReplay_Final1571133539.MP4.zip](https://github.com/react-native-community/react-native-camera/files/3728577/RPReplay_Final1571133539.MP4.zip)

cristianoccazinsp:
Hi @igor-lemon . The preview will still flicker if the video and photo sizes/resolution are different, there's really no way around that if you intend to capture photo and video without changing the preview (take a look at the native camera app for example).

However, what the update does two things: First and most importantly, the flicker is NOT visible on the final recording (you would often see a dark first frame before). Secondly, if you set the photo size to match the video resolution, then the camera does not need to switch modes, and there should no be flickering (before, it would still flicker due to audio input/output changes).
Here's what I mean by a dark frame:
dark video



If you are recording at 720p (either setting quality on recordAsync or defaultVideoQuality), it means you will need to set the `pictureSize` prop to "1280x720". This will force both the preview for pictures and video to be on the same preset and remove flickering due to settings change. Note that I haven't change this, this prop has been always been there (and not really documented), but I figured out it could be used to remove the flickering even on the preview if the audio issue was solved.

Now the default values for both video and picture are "High" and "Photo". These two might be internally the same for some devices iOS devices, but there's no guarantee about it.

Note that this is iOS only, Android doesn't seem to do any change to the preview even if it is recording at a different resolution.

xHeinrich:
This seems to have broken torch mode on ios.

cristianoccazinsp:
Are you sure it was working before? I've never got flash to work with
videos at all on neither platform. Flash wasn't touched neither, I will
take a look!

cristianoccazinsp:
@xHeinrich you are right, torch is not working properly. If torch is enabled by default, it will blink and get turned off right after the camera mounts. This also happens with 3.6.0 and is related to flash being updated before the session is commited (which also turns it off for some reason). I'm testing a fix right now.

Issue #2541, opened 2019-10-15
Honor the `captureAudio` property on iOS


cristianoccazinsp:
# Bug Report

**To Do First**

- [X] Did you try latest release?
- [X] Did you try master?
- [X] Did you look for existing matching issues?

**Platforms**

Android
iOS

**Versions**

- react-native-camera: 3.7.2

**Description/Current Behaviour**

PR https://github.com/react-native-community/react-native-camera/pull/2539 introduced a change that would break apps without audio recording permission on iOS. 

In order to fix this, code should correctly honor the `captureAudio` (default true) prop and do not create an audio recording connection by default if set to false.

On Android, this prop doesn't seem to be used, but doesn't seem to be needed neither since the JS code will set mute to true by default if this is set or no permission has been given; thus, preventing any crash related to permissions.

cristianoccazinsp:
Can someone give me the rationale behind only requesting audio permission if DEBUG? https://github.com/react-native-community/react-native-camera/blob/master/ios/RN/RNCameraManager.m#L452

Since the captureAudio prop is not even passed down to native, I'm trying to avoid it altogether by just requesting audio if there's permission. Since the camera only renders when all permissions have been checked, this should work. However, the above line seems to only request audio permission if debug, which is something I don't quite get.

Passing down the captureAudio prop might be problematic since it might not be set when the camera is initializing.

SmartArray:
@cristianoccazinsp Thanks for your contribution. Finally it works perfect on my iOS 13 device

PR #2542, opened 2019-10-15
Honor captureAudio flag by not requesting audio input if set to false.

cristianoccazinsp:
# Summary

Fix for https://github.com/react-native-community/react-native-camera/issues/2541 to prevent iOS from prompting for audio recording permission if captureAudio is set to false.

Ideally, the captureAudio prop should not reach native at all, and the request permission call should be used just like Android. In practice, however, this makes Apple complain about microphone usage (even if it is not used at all, https://github.com/react-native-community/react-native-camera/issues/2085). For this reason, the permission request call on iOS is dummy when not in DEBUG mode. The flag on native should not be needed at all if we can do the permission check in a way Apple does not complain.

In addition, fix https://github.com/react-native-community/react-native-camera/issues/2538 even further. The 0.5 delay might still allow for the video not recording log to happen (and video not getting stopped). With the added check, an enqueued video recording should not start if requested to stop early.

## Test Plan

Tested on iPhone 7, iOS 13. The following cases were tested without any crashes:

- Fresh install, captureAudio true (default): App requests audio permission as soon as the camera mounts, this behaves just like Android. However, video is already visible (camera fully mounted) due to the dummy iOS audio permission check. Furthermore, on iOS production build, audio permission will be always incorrectly reported as given due to that dummy implementation ( https://github.com/react-native-community/react-native-camera/blob/master/ios/RN/RNCameraManager.m#L452 )

- After the above install. Audio permission removed from settings. Videos record fine, no audio, no crash. No permission checks done neither, but audio device request error is handled.

- Fresh install, captureAudio=false. Camera mounts and no audio permission prompt. 
- Update app, set captureAudio=true. Camera mounts, audio prompt happens. 

Real device

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/RN/RNCamera.h
+++ b/ios/RN/RNCamera.h
@@ -45,6 +45,7 @@
 @property(nonatomic, assign) BOOL canReadText;
 @property(nonatomic, assign) BOOL canDetectFaces;
 @property(nonatomic, assign) BOOL canDetectBarcodes;
+@property(nonatomic, assign) BOOL captureAudio;
 @property(nonatomic, assign) CGRect rectOfInterest;
 @property(assign, nonatomic) AVVideoCodecType videoCodecType;
 @property(assign, nonatomic)
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -43,6 +43,9 @@
 
 static NSDictionary *defaultFaceDetectorOptions = nil;
 
+BOOL _recordRequested = NO;
+
+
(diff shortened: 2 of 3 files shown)


sibelius:
@MichielDeMey  could you test this?

MichielDeMey:
@sibelius Tested and verified working.
It no longer crashes the due to missing microphone permissions. 👍

Issue #1634, opened 2018-06-13
flashMode.torch not turning on light RNCamera


mcoder14:
Using *RNCamera* 

### Steps to reproduce
<RNCamera
					ref={cam => {
						this.camera = cam;
					}}
					style={this.getStyle().preview}
					type={RNCamera.Constants.Type.back}
					flashMode={RNCamera.Constants.FlashMode.torch}
					onFocusChanged={() => { }}
					onZoomChanged={() => { }}
					defaultTouchToFocus
					mirrorImage={false}
					barCodeTypes={[
						RNCamera.Constants.BarCodeType.aztec,
						RNCamera.Constants.BarCodeType.code39,
						RNCamera.Constants.BarCodeType.code39mod43,
						RNCamera.Constants.BarCodeType.code93,
						RNCamera.Constants.BarCodeType.code128,
						RNCamera.Constants.BarCodeType.datamatrix,
						RNCamera.Constants.BarCodeType.ean13,
						RNCamera.Constants.BarCodeType.ean8,
						RNCamera.Constants.BarCodeType.interleaved2of5,
						RNCamera.Constants.BarCodeType.itf14,
						RNCamera.Constants.BarCodeType.pdf417,
						RNCamera.Constants.BarCodeType.qr,
						RNCamera.Constants.BarCodeType.upce,
					]}
					onBarCodeRead={barcodeReadFunction}
				>

### Expected behaviour
The torch mode should turn on the flashlight and it should stay on

### Actual behaviour
The flash flickers twice and then does not turn on

### Environment
-  "react-native": "0.55.3",
-  iOS 11.3

### react-native-camera
      "react-native-camera": "^1.1.2",

jlemm45:
Torch mode needs to be set after the camera has already loaded so essentially try using a state variable tied to a button toggle of some sort to control which flashMode is on. 

Example:
`flashMode={torchOn ? Camera.Constants.FlashMode.torch : Camera.Constants.FlashMode.off}`

n1ru4l:
We should improve this!

zhangxinagjunHJH:
Have you solved the problem already？

zhangxinagjunHJH:
I rebuild project from xcode, it worked!

joaaomanooel:
I also have this problem on Android and tried these solutions, but it did not work

GotiMehta:
[...]

PR #2543, opened 2019-10-16
Torch fixes for iOS and a few nil checks.

cristianoccazinsp:
# Summary

Fixes for iOS torch:
- Torch turned on by default should work as expected (turned on after mounting and camera type switch)
- Torch maintained when video recording starts (flash will flicker if capture preset changes, not visible on the recording though)
- Torch resumed when app comes back from background (it is disabled on session pause)
- Added a few nil checks to avoid code calls if not needed (e.g., changing flash when device hasn't been setup yet)

Related SO discussions about flash issues:

https://stackoverflow.com/questions/47181531/cant-start-iphone-x-flash-in-touch-mode
https://stackoverflow.com/questions/53657378/device-torch-turns-off-when-starting-avcapturesession

Update: Android's Camera1 is now also fixed, and the camera's config values should be properly set when recording is started.

Related issues:
https://github.com/react-native-community/react-native-camera/issues/1634 (torch does not work if turned on by default)
https://github.com/react-native-community/react-native-camera/pull/2539#issuecomment-542496916

## Test Plan
Tested on iPhone 7.
Tested on Moto G5 (Android 8), Pixel 2 (Android 10)

Real device

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -254,6 +254,10 @@ BOOL _recordRequested = NO;
 {
     AVCaptureDevice *device = [self.videoCaptureDeviceInput device];
     NSError *error = nil;
+    
+    if(device == nil){
+        return;
+    }
 
     if (self.flashMode == RNCameraFlashModeTorch) {
         if (![device hasTorch])
@@ -309,11 +313,16 @@ BOOL _recordRequested = NO;
 - (void)defocusPointOfInterest
 {
     AVCaptureDevice *device = [self.videoCaptureDeviceInput device];
+    
 
     if (self.isFocusedOnPoint) {
 
         self.isFocusedOnPoint = NO;
 
+        if(device == nil){
+            return;
+        }
+        
         device.subjectAreaChangeMonitoringEnabled = NO;
         [[NSNotificationCenter defaultCenter] removeObserver:self name:AVCaptureDeviceSubjectAreaDidChangeNotification object:device];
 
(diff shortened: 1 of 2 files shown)


cristianoccazinsp:
@sibelius  what's wrong with the checks?

sibelius:
maybe we should migrate to github actions

cristianoccazinsp:
Added a fix for Android's Camera1 as well. Torch should consistently be working when recording is started rather than working only "sometimes".

SO discussion here: https://stackoverflow.com/questions/11623157/enabling-camera-flash-while-recording-video

ios/RN/RNCamera.m
```diff
@@ -254,6 +254,10 @@ - (void)updateFlashMode
 {
     AVCaptureDevice *device = [self.videoCaptureDeviceInput device];
     NSError *error = nil;
+    
+    if(device == nil){
```
sibelius: why so many decide checks?

cristianoccazinsp: They don't really change much if they are there or not, because the following lines will always end up exiting due to how objective C handles nil (there won't be a null pointer exception), but since most props are set by RN (all setters called) when the component mounts but the camera is not ready, all those setters will end up failing / doing nothing. Those nil checks will just return early and avoid running nil through all the following calls.

Issue #2555, opened 2019-10-21
Android: do not use UI thread for heavy operations


cristianoccazinsp:
# Feature-Request

**Describe the Feature**
After some digging, looks like most of the ANRs and slow camera startup comes from the fact that most of the heavy operations (open and preview start) are run on the main thread. Ideally, they should be dispatched on a serial queue just like the iOS implementation. That would help prevent both race conditions while also greatly improving performance and responsiveness. I'm not entirely sure what was the reason to make everything run on the UI thread, but after some testing the camera startup is twice as fast (possibly due to my heavy component test) and resuming from background is also much faster with nearly no black screens on a Pixel 2. The difference is probably smaller on lower end devices.

**Possible Implementations**
However, I did only a few dirty tests. I can't figure out how to move some stuff to other threads while keeping the same internal API to return true/false on most failed operations.

A few options: 
- Use a task queue inside Camera1 and Camera2 implementations. This has the advantage of having a lot of control about what runs where, but would require a lot of code.

- Use a task queue on the view implementation and call the Camera1/2 instance methods that are heavy through the queue. Has the issue that there's not a lot of control over the implementation, and also some callbacks are still fired on an unknown thread (e.g., orientation and surface change events).

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate
> Want this feature to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

cristianoccazinsp:
Initial work started here: https://github.com/react-native-community/react-native-camera/compare/master...cristianoccazinsp:android-ui-thread?expand=1

Will provide a PR once I confirm this fixes most ANRs that I've been seeing on Samsung.

However, here's a quick look at the difference on a Pixel 2. You can see both startup and resume from background are noticeably faster!

Camera startup before, running on UI thread:
https://zinspector.s3.amazonaws.com/usi/2/1571764113181ed9d616484ddd9c51615dec0574f6.mp4

Camera startup after, running on bg thread and with removed redundant calls:
https://zinspector.s3.amazonaws.com/usi/2/15717640689f901d3fb5154d6e891d718caf6c644c.mp4


Camera resume from background before:
https://zinspector.s3.amazonaws.com/usi/2/1571764220f781e999b00e4ff4899bb82d29b242c3.mp4

Camera resume from background after:
[...]

PR #2560, opened 2019-10-26
Android ui thread

cristianoccazinsp:
# Summary

Setup a HandlerThread (background thread) to remove heavy operations from the UI thread. This also removes a few redundant calls to camera/preview start, and adds more error checks. Overall, this results in a faster camera startup time, and faster resume from background.

Update is aimed mostly to Camera1 implementation, but Camera2 will also benefit for the start method being called in a non UI thread. 

Issue and video samples here: https://github.com/react-native-community/react-native-camera/issues/2555

Possible improvements: Identify other "slow" methods that might also need to be fired in the background thread. So far, I've only seen start/stop and preview related calls to be slow, but this might be different for other devices.

## Test Plan

Local tested on a Pixel 2 (Android 10) and Motorola G5 (Android 8). Field tested for nearly a week on production (Samsung S9, S8, Motorola Z4 and Z3, Pixel 3 XL, Pixel 4, and a few other models) without issues.

However, could use some extra testing on older devices since our app only supports Android 5 and above.

Real device

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -665,7 +665,6 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
                 mCamera.takePicture(null, null, null, new Camera.PictureCallback() {
                     @Override
                     public void onPictureTaken(byte[] data, Camera camera) {
-                        isPictureCaptureInProgress.set(false);
 
                         // this shouldn't be needed and messes up autoFocusPointOfInterest
                         // camera.cancelAutoFocus();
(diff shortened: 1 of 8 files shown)


cristianoccazinsp:
Can't tell what failed with the checks :S

cristianoccazinsp:
There's definitely something wrong with the tests...

sibelius:
flaky

cristianoccazinsp:
@sibelius do you know what's failing exactly with the tests? Doesn't seem to be any java related issue. It's also interesting that it went from "good" to "failed" with a commit that literally just removed two unused imports.

cristianoccazinsp:
Been running on Production for almost 2 weeks, not a single camera crash reported yet, has to be a record!

@sibelius  any updates with tests failing? Did I miss something? The project seems to compile just fine.

mars-lan:
Looks like this PR may have introduced a bug where the preview doesn't get loaded properly again in subsequent captures. It showed a forever spinning wheel instead.

cristianoccazinsp:
@mars-lan  can you provide some more details about that? How are you initializing the camera? I haven't seen any issues so far, and our app is made to take a lot of pictures one behind the other.

Where is the spinning wheel coming from as well? I don't think RNCamera renders any spinning wheel except when awaiting for permissions.

mars-lan:
My bad. Turns out that I was using `pauseAfterCapture` and must call `cameraRef.resumePreview()` before using the camera again.

Issue #2548, opened 2019-10-17
iOS - Camera preview stops while on a phone call


cristianoccazinsp:
# Feature-Request

**Describe the Feature**
Not sure if this is possible or not, since the native camera prevents video/audio recording while on a phone call. However, photos can still be taken on the iOS native app, but the RNCamera preview stops when a phone call starts.

This might be an iOS limitation (just like the forced camera shutter sound). Have anyone thought about this or found any work around? It's odd because the preview starts for a brief moment and then stops. I will be doing some debugging, but not an easy issue to test!

cristianoccazinsp:
Turns out the preview won't stop if audio recording is disabled. So if captureAudio={false} this won't be an issue. If audio is enabled, however, it gets interesting.

This function here: https://github.com/react-native-community/react-native-camera/blob/master/ios/RN/RNCamera.m#L1291 will pause the session/video recording, or resume it. 
Here's the PR that added this: https://github.com/react-native-community/react-native-camera/commit/8cc3ca2898f863806c47860ea1fef5aed5ee4462

Here's a similar SO discussion about how to handle session interruptions: https://stackoverflow.com/questions/35839604/camera-freeze-when-open-app-while-call-is-running

Not entirely sure how to handle this, since testing with telegram calls also behave quite erratic (telegram seems to also fail to handle audio interruptions and the call will randomly cut if opening the camera with audio!)

cristianoccazinsp:
Removing the audio input when the interruption happens allows to get the preview re-enabled, but trying to record with audio again, will end up with the same issue, and generate a loop of record start / interrupt / record stop. It is really bad that there's no way to tell if adding the audio input would cause the session to pause or not, so it doesn't seem possible to detect that problem early before recording with audio.

Looks like only the "user" can fix this by turning off audio recording if he wants the preview and/or video to still work while on a phone call. For this to happen, RNCamera should fire some event when the camera session is interrupted due to audio issues.  I think this would be the only way to also make any change not alter the current behaviour when a call comes in right now.

We know that the preview and recording always stops if the camera goes to background, so we could have an event `onAudioInterrupted` that the user can handle to set the `captureAudio` flag to false and hence recover the camera preview. Or, if nothing is done, the preview remains paused just like it happens right now. The user would then need to manually re-enable captureAudio since it is impossible to detect if the device is available again.

cristianoccazinsp:
[...]

PR #2565, opened 2019-11-02
Ios - handle audio interruptions

cristianoccazinsp:
# Summary

These changes aim at allowing the camera to run and work while audio is unavailable (due to phone calls or other events). Currently, if captureAudio is true, and a phone call comes in, the video preview becomes unavailable. 

The update does some code cleanup, and adds two new events that can be used to listen to audio interruptions so the JS side can act accordingly, and update the UI if needed.

In order to reliably detect if audio is available, and if captureAudio is true, the camera now explicitly attempts to active the audio session, and releases it once no longer used. This does not change the current behaviour if music playback is happening in background (which is stopped), but has the side effect of actually allowing music to resume once the camera is disposed due to the audio session being released.

Summary:

- Use a more generic event to handle session interruptions. This removes the need to listen to foreground/background events, **and stopping the session this way was actually redundant/wrong** (see https://forums.developer.apple.com/thread/61406). Also makes session stopping detection more reliable (calls, suspension due to a call or notification, etc., which would previously not set the recording interrupted flag on every case)

From the above docs: "No, incorrect. You _never_ need to stop your capture session. The capture session automatically stops itself when your app goes to the background and resumes itself when you come back to the foreground."

- add onAudioInterrupted and onAudioConnected events so the UI can handle scenarios where audio is wanted but not available. This should also help in keeping the preview active even if audio is interrupted and we have captureAudio={true}. 

- check, activate, and release audio sessions (if captureAudio) so that background music/playback can be resumed after the camera stops, and so we can detect early if audio is available before attempting to connect the input. This will effectively allow us to open the camera if a phone call is currently going on. Currently, the preview will end up black and it won't resume automatically after the phone call ends if it was already running before the camera opened.

- use proper observer for session error instead of of the strong self block. No benefits, but makes code more readable and allows access to instance variables

- getDeviceOrientationWithBlock might fire more than once under some circumstances, ending up taking a picture or video twice. Add a lock and additional check to prevent this.

- Check for session running before trying to record and capture. Will save from a crash if this is not controlled on the UI side. Move all camera startup to the session queue, this should also make it faster since it won't be blocking the UI and there won't be redundant calls to camera startup.

- update orientation on the session queue only to avoid race conditions. No need to get it on component constructor. This fixes a rare case that the orientation might end up wrong if moving the device quickly

Main issue here: https://github.com/react-native-community/react-native-camera/issues/2548

## Test Plan
Locally tested on iPhone 7 (iOS 12 and 13). Tested for almost 2 weeks on iOS test flight (about 36 users and 18 devices). 
Tests consisted on doing phone calls, facetime calls, and telegram calls. Calls would be done to interrupt an existing camera session, and before a session was actually started.

Can definitely use additional testing from other use cases.

Real device

- 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -167,18 +167,18 @@ BOOL _sessionInterrupted = NO;
          selector:@selector(orientationChanged:)
              name:UIApplicationDidChangeStatusBarOrientationNotification
            object:nil];
-                
+
         [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sessionWasInterrupted:) name:AVCaptureSessionWasInterruptedNotification object:self.session];
-        
+
         [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sessionDidStartRunning:) name:AVCaptureSessionDidStartRunningNotification object:self.session];
-        
+
         [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sessionRuntimeError:) name:AVCaptureSessionRuntimeErrorNotification object:self.session];
-        
+
         [[NSNotificationCenter defaultCenter] addObserver:self
             selector:@selector(audioDidInterrupted:)
             name:AVAudioSessionInterruptionNotification
             object:[AVAudioSession sharedInstance]];
-        
+
 
         // this is not needed since RN will update our type value
         // after mount to set the camera's default, and that will already
@@ -190,11 +190,11 @@ BOOL _sessionInterrupted = NO;
         [[NSNotificationCenter defaultCenter] removeObserver:self name:UIApplicationDidChangeStatusBarOrientationNotification object:nil];
 
         [[NSNotificationCenter defaultCenter] removeObserver:self name:AVCaptureSessionWasInterruptedNotification object:self.session];
-        
+
         [[NSNotificationCenter defaultCenter] removeObserver:self name:AVCaptureSessionDidStartRunningNotification object:self.session];
-        
+
         [[NSNotificationCenter defaultCenter] removeObserver:self name:AVCaptureSessionRuntimeErrorNotification object:self.session];
-        
+
         [[NSNotificationCenter defaultCenter] removeObserver:self name:AVAudioSessionInterruptionNotification object:[AVAudioSession sharedInstance]];
 
         [self stopSession];
@@ -255,7 +255,7 @@ BOOL _sessionInterrupted = NO;
 {
     AVCaptureDevice *device = [self.videoCaptureDeviceInput device];
     NSError *error = nil;
-    
+
     if(device == nil){
         return;
     }
@@ -314,7 +314,7 @@ BOOL _sessionInterrupted = NO;
 - (void)defocusPointOfInterest
 {
     AVCaptureDevice *device = [self.videoCaptureDeviceInput device];
-    
+
 
     if (self.isFocusedOnPoint) {
 
@@ -323,7 +323,7 @@ BOOL _sessionInterrupted = NO;
         if(device == nil){
             return;
         }
-        
+
         device.subjectAreaChangeMonitoringEnabled = NO;
         [[NSNotificationCenter defaultCenter] removeObserver:self name:AVCaptureDeviceSubjectAreaDidChangeNotification object:device];
 
@@ -345,7 +345,7 @@ BOOL _sessionInterrupted = NO;
 
     if(self.isExposedOnPoint){
         self.isExposedOnPoint = NO;
-        
+
         if(device == nil){
             return;
         }
@@ -365,7 +365,7 @@ BOOL _sessionInterrupted = NO;
 
     if(self.isExposedOnPoint){
         self.isExposedOnPoint = NO;
-        
+
         if(device == nil){
             return;
         }
@@ -383,7 +383,7 @@ BOOL _sessionInterrupted = NO;
 {
     AVCaptureDevice *device = [self.videoCaptureDeviceInput device];
     NSError *error = nil;
-    
+
     if(device == nil){
         return;
     }
@@ -462,7 +462,7 @@ BOOL _sessionInterrupted = NO;
 {
     AVCaptureDevice *device = [self.videoCaptureDeviceInput device];
     NSError *error = nil;
-    
+
     if(device == nil){
         return;
     }
@@ -517,7 +517,7 @@ BOOL _sessionInterrupted = NO;
 - (void)updateZoom {
     AVCaptureDevice *device = [self.videoCaptureDeviceInput device];
     NSError *error = nil;
-    
+
     if(device == nil){
         return;
     }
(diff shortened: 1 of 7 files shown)


sibelius:
I think we need to upgrade example to make CI work again

cristianoccazinsp:
The example app can definitely use some love. Would also be awesome to have it demonstrate tap to focus, pinch to zoom, camera switching when using multi-camera devices, etc.

However, the current example app seem to be building just fine. Can't tell what's wrong from the CI errors.

ios/RN/RNCamera.m
```diff
@@ -167,19 +166,17 @@ - (void)willMoveToSuperview:(nullable UIView *)newSuperview;
              name:UIApplicationDidChangeStatusBarOrientationNotification
            object:nil];
 
-        [[NSNotificationCenter defaultCenter] addObserver:self
-                 selector:@selector(bridgeDidBackground:)
-                     name:UIApplicationDidEnterBackgroundNotification
-                   object:nil];
+        [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sessionWasInterrupted:) name:AVCaptureSessionWasInterruptedNotification object:self.session];
```
sibelius: will this happen only when audio is enabled?

cristianoccazinsp: This fires when anything causes the session to be interrupted. Not only audio, but also the app going to background. It basically replaces the background/foreground events that were previously manually stopping the session, and which was wrong (the session is automatically stopped/resumed when the app goes to background/foreground).

If our session was never set to use audio, audio interruptions should not interrupt the session; hence, this event will not fire in that case.

Lastly, there were cases that our session could be interrupted, even if the app did not go to background. A good example of this would be during a phone call. When a phone call is taking place, opening the app drawer will also pause the app (even if it is not sent to background), interrupting any recording. When this happens, the previous code would not properly set the video interrupted flag.

xHeinrich:
@cristianoccazinsp i've had a go at using this and it doesnt appear to fire the onAudioInterrupted event when you pull down on a push notification which causes the audio session to be interrupted.

cristianoccazinsp:
@xHeinrich thanks for bringing that one up. Does the interruption happen only when there's a notification? Or always when you pull down the top menu?

Are you sure the camera session is indeed interrupted due to an audio interruption? The event only fires if the session's audio was removed due to an interruption. So if you were not capturing with audio, or the session was not interrupted, the event won't fire.

xHeinrich:
13.2, it's definitely interrupted. Tested on ipod touch 7th and 6th gen. If I pull down on the notification and dismiss the audio cuts and doesn't recover. It might be possible its only notifications from the current app?

cristianoccazinsp:
That's odd, it should recover at least. However, I'm unable to get the app
to pause when pulling a notification. Is it a notification from the same
app? I need to test that case.

xHeinrich:
Yes, notification is from the same app recording. In my case im using firebase notifications

cristianoccazinsp:
@xHeinrich  I can't still seem to reproduce your exact issue, can you send me a video or something or what you're doing exactly?
Please see this: https://zinspectordev2.s3.amazonaws.com/usi/2/1575217187ac0c1f0261fd41389c1dd81e999b2ac9.mp4

I'm showing a notification from the same app, and it doesn't matter if I swipe it up, press it, or ignore it, the video recording (and sound) never stops.

xHeinrich:
@cristianocca 

Video recording the ipod:
https://dk73odkxmhy1p.cloudfront.net/1fbed379-c133-4910-b503-ff8a7af29795/raw.mp4

Ipod recorded video:
https://dk73odkxmhy1p.cloudfront.net/6b3563f4-5977-4660-be11-4c0a1d7c932f/raw.mp4

cristianoccazinsp:
@xHeinrich that is quite interesting. I see that the video finishes the recording just fine, but the audio gets cut, is that right? From all my testing, when anything interrupts the audio session, the whole video preview, and any in progress recording is always stopped. So from your videos, I can only conclude that there are no interruption events getting fired, not even the session is aware of the audio interruption.

I will try to give it more testing on my test device, but I'm sure I'm doing the exact same thing as you on my video, and audio records fine until the end.

Are you doing anything on app state changes? Opening the notification drawer like that will put the app in "inactive" state (not background, not foreground, and the session normally continues just fine in this state).

Are the notifications muted or with sound? I think the notifications I've tried did not have any sound (actually, I don't think I'm getting any notifications with sound at all!)

xHeinrich:
@cristianoccazinsp correct the video records fine just the audio cuts out. I actually just tested this aswell just by pulling the draw down and the same thing happens with the audio cutting out. It looks like the entire frontend ui cuts out(the timer on the bottom right stops when pulling the draw). Im not doing anything wierd on app state change but I wonder if this might be something in the expo bare workflow cutting the ui thread out when doing it?

cristianocca:
I'm not using exp, but may it be possible that expo is pausing the app on inactive and not just on background state? Have you tried any other device? Even if the UI pauses, it shouldn't be breaking the audio session, definitely weird that video can continue but not audio. Normally both would pause/stop.

Do you see the same behaviour with incoming calls? What about speech to text?

El 1 dic. 2019 21:44, Nathan Heinrich <notifications@github.com> escribió:

@cristianoccazinsp<https://eur03.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fcristianoccazinsp&data=02%7C01%7C%7Cb3bfb2b66c1d426e35a808d776c0d9c2%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637108442962303221&sdata=crIdpTXa020s6y%2FGQEnjUTiXpxmxWTbW1SfOhyZdCks%3D&reserved=0> correct the video records fine just the audio cuts out. I actually just tested this aswell just by pulling the draw down and the same thing happens with the audio cutting out. It looks like the entire frontend ui cuts out(the timer on the bottom right stops when pulling the draw). Im not doing anything wierd on app state change but I wonder if this might be something in the expo bare workflow cutting the ui thread out when doing it?

xHeinrich:
@cristianoccazinsp Happens on every device we've tested thus far(6s, 8plus and 10 max). Incoming calls and back-grounding the app both stop the recording but notifications, audible notifications and sometimes sounds from the current app stop the audio. Not sure how to test speech to text, is that just Siri?

cristianoccazinsp:
Speech to text could be tested with a text input or something on the same
screen.

In app sounds shouldn't be breaking audio neither, is that something you
can reproduce every time? Maybe it's better to test than notifications.

xHeinrich:
@cristianoccazinsp I just rebuilt the app without expo unimodules and it seems to no longer lock up. Might be something to do with this? https://github.com/expo/expo/blob/master/packages/expo-av/ios/EXAV/EXAudioSessionManager.m#L38

cristianoccazinsp:
@xHeinrich it may be related. Do you see that the app pauses or is interrupted anyhow without it now? I don't even think the audio interrupt events are fired at all. It might be something related to this instead https://github.com/expo/expo/blob/master/packages/expo-av/ios/EXAV/EXAudioSessionManager.m#L116 . Modules getting paused when going background?

You could try adding some NSLog statements or breakpoints on xcode and see what's going on both expo and rncamera.

Issue #2566, opened 2019-11-05
Speech-to-Text may break the camera on iOS 12 and 13


cristianoccazinsp:
# Bug Report

**To Do First**

- [X] Did you try latest release?
- [X] Did you try master?
- [X] Did you look for existing matching issues?

**Platforms**

iOS

**Versions**

- Android:
- iOS: 12.4, 13.2
- react-native-camera: 3.9.0

**Description/Current Behaviour**
When using speech-to-text, audio is interrupted (similar to a phone call), and the camera will be forced to disable audio (if enabled) in order for the session to continue running. 

However, what seems an Apple bug, makes it so the session resumed event is never fired and the property `session.isRunning` (which is from Apple's AVCaptureSession class) is incorrectly reported as False, even if it's actually running.

This makes anything that checks `session.isRunning` to behave erroneously. This means that these checks (added by myself to prevent other crashes) (https://github.com/react-native-community/react-native-camera/blob/master/ios/RN/RNCamera.m#L718 and https://github.com/react-native-community/react-native-camera/blob/master/ios/RN/RNCamera.m#L842) will report an incorrect value, even if the session is indeed running, and the camera will not capture until another interruption happens.

**Expected Behaviour**
The session's `isRunning` value should reflect the real session status. 

**Steps to Reproduce**
Start the camera with audio capture, and start a speech-to-text session (through any text input). Try to capture after recording, and capture will fail. A breakpoint will show that `session.isRunning` is NO, even if the session is actually running.

**Possible Fixes**

- Find out why the session running variable is not reflecting it's actual state. File a bug with Apple? 

- Remove the session running checks and catch any error that might arise from capturing without a valid session.

- Track the session running status with our own variable.

> Love react-native-camera? Please consider supporting our collective: 👉 https://opencollective.com/react-native-camera/donate
> Want this issue to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

PR #2567, opened 2019-11-05
bugfix: Handle audio interruption in session queue

cristianoccazinsp:
# Summary
Tiny update so that the session interrupted event fires its audio changes in the session queue to prevent concurrent changes to the AVCaptureSession instance. Such concurrent changes could lead to an inconsistent `isRunning` property.

Should fix:  https://github.com/react-native-community/react-native-camera/issues/2566
Will report back with other devices in a day or so.

## Test Plan

Test on iPhone 7 (13.2) using speech-to-text on a screen with the camera preview and audio enabled.

A real device

Cause an audio interruption without leaving the actual app (e.g., text to speech and music)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -80,7 +80,6 @@ BOOL _sessionInterrupted = NO;
         _recordRequested = NO;
         _sessionInterrupted = NO;
 
-
         // we will do other initialization after
         // the view is loaded.
         // This is to prevent code if the view is unused as react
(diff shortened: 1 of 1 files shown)


ios/RN/RNCamera.m
```diff
@@ -1437,7 +1436,10 @@ - (void)sessionWasInterrupted:(NSNotification *)notification
         // if we have audio, stop it so preview resumes
         // it will eventually be re-loaded the next time recording
         // is requested, although it will flicker.
-        [self removeAudioCaptureSessionInput];
+        dispatch_async(self.sessionQueue, ^{
```
sibelius: why this fix the issue?

cristianoccazinsp: We remove the audio input when the session is interrupted so the session can resume, even while audio is not available.  Removing the audio input automatically starts/resumes the session. However, when this was done in the same session interrupted event, some kind of race condition was messing up the internal session `isRunning` flag. Apple states that this value is not thread safe, so it makes sense. Moving the call to the session queue removed the concurrency issues and the flag is now properly updated.

sibelius: tks for the explanations

PR #2569, opened 2019-11-06
Fix camera orientation of some hardware and emulator which gives wron…

Chimaa123:
Fix issue that some android hardware and emulator gives 90 degrees rotated image.
Android | ✅ |

[ x ] I have tested this on a device LG K10 model and a android studio emulator


android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
```diff
@@ -239,20 +253,34 @@ private Bitmap flipHorizontally(Bitmap source) {
         return Bitmap.createBitmap(source, 0, 0, source.getWidth(), source.getHeight(), matrix, true);
     }
 
+    private int getCameraOrientation() {
+        int orientation = 0;
+        if (Build.VERSION.SDK_INT >= 21) {
+            CameraManager manager = (CameraManager) mThemedReactContext.getSystemService(Context.CAMERA_SERVICE);
+            try {
+                String cameraId = manager.getCameraIdList()[0];
```
cristianoccazinsp: Are you sure this won't cause issues? Always getting the first available camera? What if the picture was taking with the front camera?

android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
```diff
@@ -108,12 +115,19 @@ protected WritableMap doInBackground(Void... voids) {
                 int orientation = exifInterface.getAttributeInt(ExifInterface.TAG_ORIENTATION,
                         ExifInterface.ORIENTATION_UNDEFINED);
 
+                int cameraOrientation = getCameraOrientation();
```
cristianoccazinsp: You are getting the camera orientation after the picture has been taken in an async task. The phone might have been rotated by now to another position. Have you tried what happens in such case? You will probably be using a rotation that's different than what the picture was taken with.  You may need to do this on the picture captured events of each camera implementation.

cristianoccazinsp:
Do you know what devices this happens with? I've never seen this reported so far, but since I'm using `skipProcessing: true` and `fixOrientation: false` I wouldn't know if I have actually faced this problem at all. Most likely because the image resizing library already handles EXIF data and rotates it accordingly. 

I really thought EXIF would be enough to fix any orientation issues, have you found any phone that does not use EXIF and also takes pictures rotated? I'm curious to test this, but need the specific device so I can inquire the users using it to see if they had issues.

android/.classpath
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cristianoccazinsp: You probably don't want to include eclipse files in your commit.

android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
cristianoccazinsp: You can probably already get the camera orientation from each Camera implementation from these lines. The implementation will already handle getting the orientation from the right camera.

https://github.com/react-native-community/react-native-camera/blob/master/android/src/main/java/com/google/android/cameraview/Camera1.java#L768
https://github.com/react-native-community/react-native-camera/blob/master/android/src/main/java/com/google/android/cameraview/Camera2.java#L693

So rather than getting the camera orientation after the picture has been taken and the image is getting processed, you may be able to get it right before the capture event from those lines, and pass it down to the async task. What do you think? That would ensure you always get the orientation from the right camera, at the right time.

android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
Chimaa123: That is nice catch. I will update the code.

android/.classpath
Chimaa123: Yeah, this issue occures very rare with specific devices in our customers. 
Such as 57513  OPPO  CPH1853  android  8.1.0, samsung  SM-T295  android  9, OPPO  CPH1911  android  9, samsung  SM-A750GN  android  9, OnePlus  ONEPLUS A5010  android  9, LG K10 android 5.1.1 and also emulators whichs android below android 5.1.1

android/src/main/java/com/google/android/cameraview/Camera1.java
```diff
@@ -697,7 +697,7 @@ public void onPictureTaken(byte[] data, Camera camera) {
                         isPictureCaptureInProgress.set(false);
 
                         mOrientation = Constants.ORIENTATION_AUTO;
-                        mCallback.onPictureTaken(data, displayOrientationToOrientationEnum(mDeviceOrientation));
+                        mCallback.onPictureTaken(data, displayOrientationToOrientationEnum(mDeviceOrientation), 0);
```
cristianoccazinsp: Camera1 implementation doesn't have the orientation value? Why the hard-coded 0? I believe Camera1 has `getCameraOrientation` to get it's currently selected camera orientation.

android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java
Chimaa123: I pushed new commit. Please, check this out

android/src/main/java/com/google/android/cameraview/Camera1.java
Chimaa123: Thank you. That was nice catch. I didn't recognize there was getCameraOrientation function. I have just checked there is no mCameraOrientation member variable, and I thought which is why the cameramanager doesn't suppout <21 android versions. Anyway,  I pushed new commit. Pls check it.

cristianoccazinsp:
Changes look good! Although I don't have time right now to actually run/test it. 

Have you confirmed it fixes the issues you were seeing with those devices?  Most importantly, any breaking change for devices that reports the right exif data?

Chimaa123:
> Changes look good! Although I don't have time right now to actually run/test it.
> 
> Have you confirmed it fixes the issues you were seeing with those devices? Most importantly, any breaking change for devices that reports the right exif data?

This issue is occured in really rare device such as  OPPO CPH1853 android 8.1.0
samsung SM-T295 android 9
OPPO CPH1911 android 9
samsung SM-A750GN android 9
OnePlus ONEPLUS A5010 android 9
LG K10 android 5.1.1 and android emulators. 

 I 've tested it on LG K10, and android emulators. LG K10 is the only one device which shows this issue. I will monitor other devices from customers data.

fabriziobertoglio1987:
I believe that you need to rebase and resolve those conflicts.

Kuhne1:
Tested on MI 9 SE 
MIUI V11.0.2
React-Native 61.5

The image is saved upside down when using the mirror option. Is fine on all other android devices I have tested on.

Issue #2571, opened 2019-11-08
Metadata is lost on iOS


cristianoccazinsp:
# Bug Report or Feature request?

**To Do First**

- [X] Did you try latest release?
- [X] Did you try master?
- [X] Did you look for existing matching issues?

**Platforms**

**Versions**

- iOS: 13.2
- react-native-camera: 3.9.0

**Description/Current Behaviour**

It looks like compared to Android (with skipProcessing), iOS does not store any (or all) Metadata info on the final image. It would be ideal for all the camera metadata to be preserved, just like Android. It looks like the iOS photo only stores the EXIF rotation, but not any other metadata (test with http://exif.regex.info/exif.cgi)

Perhaps some of the data is lost here: https://github.com/react-native-community/react-native-camera/blob/master/ios/RN/RNCamera.m#L747

Here's a comparison between a photo taken with RNCamera, and another one taken with the native app.

![image](https://user-images.githubusercontent.com/48869228/68440298-52e40f00-01a9-11ea-895d-36868280e192.png)

It is also quite interesting, how the RNCamera version of the photo is 3 times bigger.
![image](https://user-images.githubusercontent.com/48869228/68440382-a0607c00-01a9-11ea-8c09-35a73e6c0eff.png)
![image](https://user-images.githubusercontent.com/48869228/68440421-c2f29500-01a9-11ea-97a1-53ddc0370f4a.png)

It looks like the resizing might be causing some side effects, not only the metadata loss. I will do some testing when I get free time, but I wonder if this was discussed already.

**Expected Behaviour**

Image metadata is stored with the photo.

**Steps to Reproduce**
Capture a photo with iOS and inspect its metadata.

**Additionals**

> Love react-native-camera? Please consider supporting our collective: 👉 https://opencollective.com/react-native-camera/donate
> Want this issue to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

PR #2573, opened 2019-11-09
iOS: Preserve exif/metadata on photo capture.

cristianoccazinsp:
# Summary
This PR attempts to preserve (and correct it if the image is manipulated) any exif/metadata captured by the original capture event. Right now, any captured picture is stripped from all it's metadata.
Related issue: https://github.com/react-native-community/react-native-camera/issues/2571

Also added a few comments to the capture process to make it easier to read.

## Test Plan

Tested on iOS 13.2 with an iPhone 7. Tested with both default settings, `forceUpOrientation`, and `width` settings to ensure that exif values related to those are also properly updated.

Note: Android already preserves any exif data only if `skipProcessing` is true, or if `writeExif` (undocumented option that defaults to false) is true. Metadata is otherwise lost.

Question: Should this option be removed, and exif always written? Or, should this prop be documented and used on both iOS and Android with its current false default? I really see absolutely no reason to not store the photo's EXIF data. Not to mention it is important when it is used for rotation.

Picture with capture metadata:
image

Real device

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist


cristianoccazinsp:
@sibelius I did not notice any performance change.

In any case, the most noticeable change is this line `NSData *takenImageData = UIImageJPEGRepresentation(takenImage, quality);` which converts the image data to the compressed / final data, with `CGImageDestinationAddImage(destination, takenImage.CGImage, (__bridge CFDictionaryRef) metadata)` which does pretty much the same thing, but with added metadata dictionary.

Still not sure if it would be a good idea to make the Android's `writeExif ` flag a public/documented option, and implement the same thing on iOS, just in case anyone wants to disable EXIF. However, disabling EXIF is dangerous if the fix orientation flags are not used.

sibelius:
We can have a flag to disable metadata
Default could be to save metadata
It could be useful for GDPR stuff

cristianoccazinsp:
Android already has the undocumented option here: https://github.com/react-native-community/react-native-camera/blob/master/android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java#L131

Not sure if the goal of whoever added that option was the same or not, but it's definitely false by default. Unless `skipProcessing` is true, in which case everything is always written regardless. A bit confusing.

cristianoccazinsp:
We also have the thing about camera's not recording GPS location automatically. Such info must be added manually from somewhere else. Probably not a good idea to query any location from RNCamera at all since that would require the app to use location permissions.  In my current project our goal is to get all metadata and also GPS/time data added as EXIF, so I might be adding another PR soon to be able to pass additional metadata to the stored file/exif.

ios/RN/RNCamera.m
```diff
@@ -6,6 +6,7 @@
 #import <React/RCTLog.h>
 #import <React/RCTUtils.h>
 #import <React/UIView+React.h>
+#import <MobileCoreServices/MobileCoreServices.h>
```
cristianoccazinsp: @sibelius  do you know if this import might add an extra framework requirement? Looks like the MobileCoreServices framework is always available for iOS applications (and it doesn't need to be added manually or anything), but just to be sure this won't cause any conflicts.

cristianoccazinsp:
Add a small update so that `onPictureTaken`, which was only available for Android, is now also available for iOS. I found this event to be very useful to update the UI to allow for new pictures while processing is still happening (especially on slow devices).

ios/RN/RNCamera.m
sibelius: this is available to everyone I guess

sibelius:
`onPictureTaken` is a cool addition

cristianoccazinsp:
Hey @sibelius , thanks for merging this.

My next step and PR, closely related to this, will be something so that additional metadata can be added to the EXIF dict. Android already implements this through the `writeExif` prop, which when set to an object, will be merged into the final picture's EXIF (unless skipProcessing is used).

Do you think it would be a good idea to implement `writeExif` for iOS as well? Should it also behave like Android and avoid any EXIF writing at all if not given?

sibelius:
yep

we need consistency among android and ios

cristianoccazinsp:
How are you passing the exif data? I've been using it for a while now and it works fine on both iOS and Android. 

Here's a snippet of the code I'm using. Note the additional metadata that I'm also passing; it should be irrelevant.

```
let extraExif = {
    "UserComment": JSON.stringify({
      datetime: uploadData.datetime,
      
      docId: uploadData.docId
    }),
    "Software": `${conf.versionText}`,
    "Artist": this.props.user.email,
    "GPSLatitude": uploadData.latitude,
    "GPSLongitude": uploadData.longitude,
    "GPSAltitude": uploadData.altitude
}

      
let options = {
  quality: PICTURE_COMPRESSION,
  fixOrientation: true,
  forceUpOrientation: true,
  writeExif: extraExif
};

data = await this.camera.takePictureAsync(options);
```

Any chance you're not passing the right options? Can you double check the picture with http://exif.regex.info/exif.cgi ?

Related code is here if you want to debug, but it should be working: https://github.com/react-native-community/react-native-camera/blob/master/ios/RN/RNCamera.m#L861

phyzical:
Thanks for the reply even though i removed, i found your second pr merging the IOS so was going to add as a legit issue haha,

this is how im passing which looks to be correct.

```
 constwriteExif: {
                                GPSLatitude: position.coords.latitude,
                                GPSLongitude: position.coords.longitude,
                                GPSAltitude: position.coords.altitude,
                            },
```
```
const options = {base64: true, writeExif};
const data = await this.camera.takePictureAsync(options);
```
I uploaded straight from the ios device to that provided exif inspector and i could not see the location metadata there either hmm. 

Thanks for the direction code side i will have a closer look and see what happens.

cristianoccazinsp:
I have a feeling it is the base64 prop.  You may want to start checking that, but I don't see anything on the code that might cause the base64 string to not contain the exif data.

You can also debug with with the `exif` option so the exif info is returned to the JS.

phyzical:
Thanks for the additional suggestions,

* trying base64 false didnt change the final output metadata wise
*  Am i using the exif prop wrong? as the following didnt see to return any exif data in the const "data"
```
const options = {base64: false, writeExif, exif: true};const data = await this.camera.takePictureAsync(options);
console.log(data)
```

i will confirm .m side just incase

cristianoccazinsp:
Looks right to me. The exit data is not in the final image neither? Are you
sure you're using the latest version?

phyzical:
Android it seems to save fine.

I believe it is the latest, i atleast confirmed your code as present
im running 
```
"react-native-camera": {
      "version": "3.15.1",
      "resolved": "https://registry.npmjs.org/react-native-camera/-/react-native-camera-3.15.1.tgz",
}
```
and podfile.lock
```
- react-native-camera (3.15.1):
    - React
    - react-native-camera/RCT (= 3.15.1)
    - react-native-camera/RN (= 3.15.1)
  - react-native-camera/RCT (3.15.1):
    - React
  - react-native-camera/RN (3.15.1):
    - React
```

phyzical:
Hmm so ive tried some debugging in the .m suggested  just note i am pretty noob when it comes to the swift side, but i was able to NSLog and breakpoint the process on  onPictureTaken and initWithBridge. But it never hits anything in takePicture so i couldnt see what was being done with the exif data ios core side any idea why that might be?

cristianoccazinsp:
You should put a breakpoint here: https://github.com/react-native-community/react-native-camera/blob/master/ios/RN/RNCamera.m#L743

What device are you using?

phyzical:
Okay thanks ill try again on monday, (dont have access to it on the weekend)

i was using an iphone 8 simulator i think. 
i will try iphone 11 instead incase it helps makes a difference.

phyzical:
i found why my breakpoint wasnt being hit by tracing from the onTakePicture opps.. didnt see that mentioned in the docs maybe i missed it?

```
#if TARGET_IPHONE_SIMULATOR
            customSimulatorFunc
#else
            [view takePicture:options resolve:resolve reject:reject];
#endif
```

ill try again with a real device. Should we look at adding metadata as part of the "customSimulatorFunc" ?

cristianoccazinsp:
Good find, never thought about the simulator being the root cause. Honestly, anything related to the camera is not good enough on a simulator so I wouldn't really bother with it. But if you have free time, why not :)

phyzical:
okay a real device for ios is returning the exif as expected :).
if i get a chance i will look at adding the IOS Simulator functionality shouldn't be too hard to implement.

But for now my problem has moved across to saving the image to device not storing the meta data for only IOS but that is CameraRoll module's issue.

Thanks for all the speedy replys @cristianoccazinsp 
Talk again soon hopefully!

castor55:
@phyzical I have a similar issue.
"react": "16.13.1",
"react-native": "0.63.3",
"react-native-camera": "^3.40.0"

```
let exifData = {};
if (position) {
  exifData = {
    GPSLatitude: 65.9667,
    GPSLongitude: -18.5333,
    GPSAltitude: 15.04444408,
   };
}

const options = {
  quality: 0.5,
  base64: true,
  writeExif: exifData,
};

const data = await this.camera.takePictureAsync(options);
Exif.getExif(data.uri)
  .then(msg => console.log('OK: ' + JSON.stringify(msg)))
  .catch(msg => console.log('ERROR: ' + msg));
```

Adding GPS info in Android works but not in iOS.
I mean, I could get the GPS info through this website: http://exif.regex.info/exif.cgi
I checked the EXIF data in the console log.

Android:
```
"exif":
{"GPSLongitude":"18/1,31/1,598800000/10000000",
"GPSTimeStamp":null,
"GPSLatitudeRef":"N",
"GPSDateStamp":null,
"GPSAltitudeRef":"0",
"GPSLatitude":"65/1,58/1,1200000/10000000",
"GPSAltitude":"150444/10000",
"GPSProcessingMethod":null}
```
iOS:
```
"exif":
{"Orientation":6,
"{GPS}":
  {"LongitudeRef":"W",
  "LatitudeRef":"N",
  "AltitudeRef":0,
  "Longitude":18.5333,
  "Altitude":15.044444084121091,
  "Latitude":65.96669666666666}
}
```

Why do I get the different data? However, iOS GPS EXIF data is not detected.
How can I fix this issue?

cristianoccazinsp:
@castor55 which device are you testing on? Looks like the iOS version does have the data, but in a different format? Have you tried without base64 to see if that's the issue?

I really suggest you set some breakpoints around here https://github.com/react-native-camera/react-native-camera/blob/master/ios/RN/RNCamera.m#L874 and debug the issue.

castor55:
@cristianocca I'm testing on iPhone 6s, iOS 14.6. Let me debug the issue.

cristianoccazinsp:
> @cristianoccazinsp I'm testing on iPhone 6s, iOS 14.6. Let me debug the issue.

My oldest device is an iphone 7, sadly, but there should really be no differences between the two. For me, exif metadata for GPS location is working fine and confirmed when using http://exif.regex.info/exif.cgi

Please try without base64 first to see what's happening. Also, are you sure it's not a bug with the `Exif.getExif(data.uri)` code you're using?

castor55:
@cristianoccazinsp 
1. Tested on iPhone 7, see the same issue.
2. Tried ```base64: false```, see the same issue.
3. Added ```exif: true``` to see the added EXIF data and checked the captured image data. Same issue...
```
{
  exif: {
    DPIHeight: 72,
    '{GPS}': {
      Latitude: 65.9666976928711,
      Altitude: 15.04444408416748,
      Longitude: 18.533300399780273,
      AltitudeRef: 0,
      LatitudeRef: 'N',
      LongitudeRef: 'W' },
    Orientation: 6,
    ... }
  ...
}
```

castor55:
@cristianoccazinsp Could you check this screenshot?
Screen Shot 2021-08-06 at 10 13 08 AM

cristianoccazinsp:
I think the structure is actually fine, but iOS nests the gps info different than android. However, the info should be valid if you view it on windows or with an exif viewer like the URL you linked. Are you sure it’s not just a problem with the library you are using to read exif info?

castor55:
> I think the structure is actually fine, but iOS nests the gps info different than android. However, the info should be valid if you view it on windows or with an exif viewer like the URL you linked. Are you sure it’s not just a problem with the library you are using to read exif info?

@cristianoccazinsp Somehow, I don't see the info. As I said, I added `exif: true` to the option parameter this time and check the `exif` object in photo data. I mean I didn't use the library.

castor55:
@cristianoccazinsp Please let me know if you have any updates. Thanks!

cristianoccazinsp:
@castor55 we have been using the library for a while and this is definitely working for us. Tested on both iPhone 7 and 12 Pro (iOS 14.6). We do maintain a separate fork as our app relies on the camera in a way we can't afford using untested versions from the main branch. Feel free to use it and compare to see if it gives you different results, but the code should be pretty much the same: https://github.com/cristianoccazinsp/react-native-camera/tree/internal

Also, here's part of how we use the code to capture the exif info:

```
let extraExif = {
  Software: `appName ${conf.versionText}`,
  Artist: this.props.user.email,
};

if (
  uploadData.latitude != null &&
  uploadData.longitude != null &&
  uploadData.altitude != null
) {
  // native camera library handles setting the right format for us
  Object.assign(extraExif, {
    GPSLatitude: uploadData.latitude,
    GPSLongitude: uploadData.longitude,
    GPSAltitude: uploadData.altitude,
  });
}


let options = {
  quality: 1,
  writeExif: extraExif,
  path: conf.fs.THUMBNAILS_PATH_ABS(`res_${new Date().getTime()}.jpg`),
};

data = await this.camera.takePictureAsync(options);
```

Lastly, please see attached the photo that contains exif info and use it to compare if you wish:


Screen Shot 2021-08-09 at 21 13 47

castor55:
@cristianoccazinsp I really appreciate your help. Sorry for taking your time but, it was my bad...
After taking the photo, I upload it to the server. I was testing with the uploaded one. There is an issue on the server-side and, local images have the GPS info. This is not an issue with this library.

cristianoccazinsp:
Glad you got it figured out after all!

PR #2574, opened 2019-11-10
Doc update for react-navigation

TheodoreSevvos:
# Summary
This PR is about the  documentation of how to implement  `react-native-camera` with `react-navigation`.  I found the documentation a bit hard to follow with the focus update, in order to keep track of component's state. I think `React Navigation`, version 4.x, offers a very good solution since it supports a higher order component to check the focus state when a component is rendered, [link](https://reactnavigation.org/docs/en/with-navigation-focus.html). It also makes the example more clear and straightforwards, since the reader won't have to mess with the state of the component. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅    |

## Checklist

### DIFF
--- a/docs/react-navigation.md
+++ b/docs/react-navigation.md
@@ -4,29 +4,25 @@ title: React Navigation
 sidebar_label: React Navigation
 ---
 
-React-navigation does not unmount components when switching between tabs. So when you leave and return back to the screen with the camera component it will just be black view. So a good solution is to use `componentDidMount` and added two listeners `willFocus` and `willBlur` to help you mount and unmount the views.
+React-navigation does not unmount components when switching between tabs. So when you leave and return back to the screen with the camera component it will just be black view. So a good solution is to use `withNavigationFocus`, which is a higher order component, and wrap it around your component. Then, you can have access to `isFocused` from `props`.
 
 ```jsx
-componentDidMount() {
-   const { navigation } = this.props;
-   navigation.addListener('willFocus', () =>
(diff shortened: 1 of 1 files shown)


cristianoccazinsp:
>  So when you leave and return back to the screen with the camera component it will just be black view.

Is that 100% true in any case? I'm sure I'm using the camera and leaving it open while navigating to other secondary screens, and the camera keeps running just fine when I come back, without any blur or focused event listeners.  I'm not using face/barcode recognition though, perhaps that causes issues.

TheodoreSevvos:
Hi there!

Just to clarify that the quote, was part of the previous documentation.
Regarding the issue, it depends on how someone uses the navigator component. 
In the case of `createBottomTabNavigator`, the issue exists since the parent navigator will not unmount the components. 
Barcode/face recognition has nothing to do with that since I tested the `RNCamera` component without using them at all.

cristianoccazinsp:
Have you tried the same with a stack navigator? I understand that the previous screens are not unmounted, but for me with a stack navigator I can add 3 or even 5 screens on top of the camera screen and it will still work fine after popping all those screens.

TheodoreSevvos:
You are correct. I just tried with a stack navigator and I didn't have the issue. 
So, I don't know if this specific case has to be mentioned in the doc.

sibelius:
this is better than before

but we need to also provide a hooks based example

PR #2577, opened 2019-11-11
Implement writeExif for iOS, Android improvements

cristianoccazinsp:
# Summary

- Improve Android code so skipProcessing is not needed, the code is more in line with iOS, and is "fast" by default. **This means that skipProcessing is no longer needed** (nor used), and adding additional options will "slow down" the capture as expected, rather than having always a lot of processing and a slow capture by default. This shouldn't be a breaking change.

- The quality (JPEG compression parameter) is also now used at the capture level. This has the advantage that the quality can be controlled even if `skipProcessing` is used, that is, if no additional processing is required. The final JPEG image is still compressed with `quality` when the processing is used.

- document the writeExif option, and implement it for iOS as well and in line with the Android version.

Notes about `skipProcessing`: 
For people previously using this setting, there should be no behaviour changes when no other image processing settings are used (width, orientation, mirroring, etc.). For people not using this setting, there should be a significant capture speed change when no image processing settings are used (due to unnecessary bitmap read/write no longer used), and no behaviour changes if any of the image processing flags are used.

## Test Plan

Tested on real devices for both Android (Pixel 2 10.0, Moto G5 8.1) and iOS (iphone 7, 13.2)

Could definitely use some more testing on Android to make sure the removal of skipProcessing didn't break anything, but ideally, it should have made capture much faster to anyone using default settings since the bitmap is no longer read and compressed if not needed to.

Real device

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist


sibelius:
why removing `skipProcessing` is not a breaking change?

cristianoccazinsp:
For the people using it, like myself, as long as you do not require other
image processing settings (e.g., up rotation, width), the final output will
be the same: the unmodified image.

For people not using it, the final output will also be the same, unless no
processing settings are used, in which case the capture should be much
faster since it won't go through the bitmap load/compress step and use the
image as is.

The only case it might have a different behavior is if you were using both
skip processing and any of the processing settings (which would make no
sense since they would be ignored if skip processing is used).

It's probably a good idea to write a note somewhere? Do you think there's a
case it would break?

sibelius:
adding a note could help

cristianoccazinsp:
> adding a note could help

Is that something you add when merging/releasing, or something that should be added on the PR?

sibelius:
On PR

cristianoccazinsp:
Added some notes about it.

sibelius:
@cristianoccazinsp can you rebase?

cristianoccazinsp:
Should be done.

cristianoccazinsp:
Added another commit. Turns out there might have been a small memleak from the metadata dictionary due to how Objective-C CF objects are managed.

sibelius:
CI is passing again after this memleak change

cristianoccazinsp:
CI has been acting funny. It passes 50% of the time randomly. Add another
commit and it will fail, add one more and it will most likely work again. I
don't have enough knowledge about it sadly.

PR #2581, opened 2019-11-13
Nitcheu mbouendeu

Njinou:
* The camera on android was not opening as soon as the user granted the authorization, because the refreshAuthorization was not being updated on realtime.
After the user grants the authorization, it still displayed the black screen with the spinning activityIndicator. But now with the change, as soon as the user grants the permission, it opens up the camera right away.
* Refactor the code to use the refreshAuthorization in ComponentDidMount, realize the function was created but never used and in ComponentDidMount was just a copy paste of that code's function.

### Steps to reproduce 
Install on the device and authorize the permission then camera will open up right away in Android

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |


src/RNCamera.js
```diff
@@ -740,15 +740,14 @@ export default class Camera extends React.Component<PropsType, StateType> {
...
-    } = await this.arePermissionsGranted();
-    if (this._isMounted === false) {
-      return;
+  async componentDidUpdate(prevProps) {
+      if (this.state.isAuthorized !== prevProps.isAuthorized) {
+        await this.refreshAuthorizationStatus();
+      }
     }
 
+  async componentDidMount() {
```
sibelius: can we move to hooks?

Njinou: Sure,

cristianoccazinsp:
Interesting, is this from the latest version? I haven't really noticed any delay after authorization is granted. Would this trigger a double render?

fabriziobertoglio1987:
to fix the Circle CI issue, I suggest you to rebase from master

```
git checkout <my-local-branch>
git fetch upstream
git rebase upstream/master
git push origin <my-local-branch>
```
you may have to use `git push origin <my-local-branch> --force`.

Issue #2562, opened 2019-10-28
Update Example App for RN > 60


cristianoccazinsp:
# Feature-Request

**Describe the Feature**

Example app could use some love and get updated to the latest RN version. Also needs to use auto linking (which probably means unlinking the current dependencies).

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate
> Want this feature to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

PR #2588, opened 2019-11-15
New example app using RN 0.61

cristianoccazinsp:
# Summary

New example app using RN 0.61, native-base, and react-navigation.

Includes a camera screen with support for the right dimensions, aspect ratio, flash, tap to focus, pinch to zoom, white balance, and multi camera support.

I've built this from the ground up, it might not look precisely pretty, but it uses most of the features I normally use and which I test most of the updates.

Can possibly be extended with a few more features/buttons for: barcode reading, face detection, etc.

It would probably be a good idea to make separate screens for other more complicated features.

Note: until the PR that adds `onPictureTaken` is merged, iOS will not correctly remove the "capturing picture" message. 

Note 2: I've set RNCamera version to "*" since it's much easier to install it from the npm source.

Note 3: Hopefully the bunch of files from native-base are not an issue, those are all for UI purposes and quick development.

https://github.com/react-native-community/react-native-camera/issues/2562

## Test Plan

Tested on both IOS and Android (real devices)

Real device

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist


cristianoccazinsp:
Captures here:

Home:
Screenshot_20191115-022252


Regular phone:
Screenshot_20191115-022244

Phone with 1 front and 3 back cameras (normal, ultra wide, telephoto):
Screenshot_20191115-022205

sibelius:
native base is too complicated (so many customization and theming), we can keep this right now and then improve later on

Issue #2587, opened 2019-11-14
[Android] RNCamera crashes/freezes occasionally


nartc:
# Bug Report

**To Do First**

- [x] Did you try latest release?
- [x] Did you try master?
- [x] Did you look for existing matching issues?

**Platforms**

Android

**Versions**

- Android:
- react-native-camera: 3.9.0
- react-native: 0.61.4
- react: 16.11.0
- react-navigation: 4.0.10

**Description/Current Behaviour**
I have an app that uses `react-navigation-tabs` (picture shown below). One of the tab displays a `RNCamera` and used as a barcode scanner. The problem I'm running into is the tab with `RNCamera` will occasionally freezes/crashes the app completely. 

I ran `adb logcat` to see what's going on but I'm not very efficient with Android native to tell so I'm going to post here.

When the `RNCamera` is working, here's what shown in the `adb logcat`
![image](https://user-images.githubusercontent.com/25516557/68874225-7554bb80-06c6-11ea-809a-d164eab395b0.png)
and here's the screen of a simulator:
![image](https://user-images.githubusercontent.com/25516557/68874253-83a2d780-06c6-11ea-8cd5-ac5782d0011a.png)

When the `RNCamera` isn't working:
![image](https://user-images.githubusercontent.com/25516557/68874286-91f0f380-06c6-11ea-98ef-15f23d35d880.png)
and here's the screen:
![image](https://user-images.githubusercontent.com/25516557/68874807-6a4e5b00-06c7-11ea-95a8-7314ed79349c.png)

I noticed the difference between two `logcat` is `Camera3-Device` related logs are nowhere to be found in the logcat when "RNCamera is not working".

More detail:
1. The screen (1) that contains the `RNCamera` is a StackNavigator.
2. Upon mount, (1) will check for CAMERA permission and request CAMERA permission if appropriate.

![image](https://user-images.githubusercontent.com/25516557/68874570-0b88e180-06c7-11ea-94e5-150390fc6589.png)
`<Scanner/>` is a regular component (not a React Navigation Component) that houses the `RNCamera`. I tried leaving just the `RNCamera` there without any parent or child, the issue is still happening. 

![image](https://user-images.githubusercontent.com/25516557/68874667-36733580-06c7-11ea-8b4a-a9b71aa2545e.png)
This is the Navigation Screen that houses the `Scanner` (which in turns render `RNCamera`). 

**Expected Behaviour**
Should not crash/freeze

**Steps to Reproduce**
[...]

PR #2591, opened 2019-11-16
Possible ANR due to events not fired on the main thread

cristianoccazinsp:
# Summary
Possible fix for https://github.com/react-native-community/react-native-camera/issues/2587

RN events were fired as is, and could end up being fired on a non main thread. The reason why the ANRs are happening only sometimes is still unknown, but this seems to fix the problem. Note that before 3.9.0, all code was running on the main thread, so this was never an issue.

PS: the changes on this file might add a merge conflict to https://github.com/react-native-community/react-native-camera/pull/2577 

Updated: Main thread wasn't needed at all, just the RN's native modules thread (which is most likely a main thread somewhere). 

## Test Plan
Tested on Android 8.1 and Android 10.

Test was performed using https://github.com/nartc/rncamera-navigation-android-test and my own code that uses this library.

Awaiting feedback from @nartc to confirm the issue is indeed resolved.

Real device

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/org/reactnative/camera/RNCameraViewHelper.java
+++ b/android/src/main/java/org/reactnative/camera/RNCameraViewHelper.java
@@ -6,8 +6,6 @@ import android.graphics.Color;
 import android.graphics.Paint;
 import android.media.CamcorderProfile;
 import android.os.Build;
-import android.os.Looper;
-import android.os.Handler;
 import androidx.exifinterface.media.ExifInterface;
 import android.view.ViewGroup;
 import com.facebook.react.bridge.Arguments;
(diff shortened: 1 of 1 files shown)


nartc:
@cristianoccazinsp Looks good so far. Haven't had ANRs with with your PR. I think we can call it a fix.

Jump back and forth navigation tabs many many times.
Bring app to background and open back up
Bring up multitask screen then go back to app
All 3 scenarios worked.

nartc:
This PR sure fixed the problem I was having. So far, I haven't experienced any ANR since yesterday. Can we get this merge soon?

Issue #2285, opened 2019-05-27
iOS - yellow warning "Class RNFaceDetectorModuleMLKit was not exported. Did you forget to use RCT_EXPORT_MODULE()?"


truongluong1314520:
# Bug Report

**Platforms**
IOS

**Versions**

- Android:
- iOS: 12.2
- react-native-camera: 2.9.0
- react-native: 0.57.0
- react: 16.5.0
- react-navigation: 2.14.2

**Description/Current Behaviour**

When I run the app, there is a warning in the first place: "Class RNFaceDetectorModuleMLKit was not exported. Did you forget to use RCT_EXPORT_MODULE()?"

The app only uses the camera for taking image, so I only follow your instruction of [iOS tab](https://github.com/react-native-community/react-native-camera#ios) no [face/text/barcode detection](https://github.com/react-native-community/react-native-camera#ios)

**Expected Behaviour**

There shouldn't be a warning

**Steps to Reproduce**

Install manually follow the [iOS tab](https://github.com/react-native-community/react-native-camera#ios) 

**Additionals**

![Simulator Screen Shot - iPhone 6 - 2019-05-27 at 21 58 38](https://user-images.githubusercontent.com/32918401/58427700-9bf84900-80ca-11e9-9760-87649bc87433.png)

**Environment**

React Native Environment Info:
    System:
      OS: macOS 10.14.5
      CPU: x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
      Memory: 28.10 MB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.15.0 - /usr/local/bin/node
      Yarn: 1.13.0 - /usr/local/bin/yarn
      npm: 6.9.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        Build Tools: 23.0.1, 25.0.2, 26.0.1, 26.0.2, 26.0.3, 27.0.3, 28.0.2, 28.0.3
        API Levels: 16, 17, 18, 19, 21, 22, 23, 24, 25, 26, 27, 28
    IDEs:
      Android Studio: 3.1 AI-173.4907809
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
[...]

PR #2599, opened 2019-11-19
implement proper RNFaceDetectorModuleMLKit if FirebaseMLVision is una…

mo22:
…vailable

# Summary

- prevents warning message described here: https://github.com/react-native-community/react-native-camera/issues/2285

- makes RNFaceDetectorModuleMLKit conform to bridge protocol if FirebaseMLVision is not installed

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

- [ ] I have tested this on a device and a simulator

### DIFF
--- a/ios/RN/RNFaceDetectorModuleMLKit.m
+++ b/ios/RN/RNFaceDetectorModuleMLKit.m
@@ -119,9 +119,25 @@ RCT_EXPORT_METHOD(detectFaces:(nonnull NSDictionary *)options
 @end
 #else
 @implementation RNFaceDetectorModuleMLKit
+
+@synthesize bridge = _bridge;
+
+- (void)setBridge:(RCTBridge *)bridge
+{
+    _bridge = bridge;
+}
+
++ (BOOL)requiresMainQueueSetup
+{
+    return NO;
+}
+
 - (NSDictionary *)constantsToExport
 {
     return [FaceDetectorManagerMlkit constants];
 }
+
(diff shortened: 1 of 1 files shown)


ios/RN/RNFaceDetectorModuleMLKit.m
```diff
@@ -119,9 +119,25 @@ - (int)exifOrientationFor:(UIImageOrientation)orientation
...
+
++ (BOOL)requiresMainQueueSetup
+{
+    return NO;
+}
+
 - (NSDictionary *)constantsToExport
 {
-    return [FaceDetectorManagerMlkit constants];
+    return @{};
```
sibelius: is this correct?

mo22: Well - the whole file depends on FirebaseMLVision, so that's why the `#if __has_include(<FirebaseMLVision/FirebaseMLVision.h>)` is in there. I'm not sure why the `#else` part with a dummy module is in there and I can't find any references to `RNFaceDetectorModuleMLKit` in the javascript side.

This right now does not change behaviour but fixes warnings.

As a clean solution I would rather put the #if completely around the module, so that `NativeModules.RNFaceDetectorModuleMLKit` does not exist if FirebaseMLVision is not linked.

sibelius: I think this is good enough for now

PR #2603, opened 2019-11-22
WIP: Honor autofocus props when useCamera2api is true

Martlohmus12:
# Summary

When useCamera2Api props is used during recording, autofocus=true/false is not being honored.

According to my understanding is that first CameraViewImpl is set to camera1 and all the props(aspect ratio, autofocus) are set to this instance.

Now if useCamera2Api props is used, then these set props are saved and when CameraViewImpl is switched to Camera2 or Camera2Api23 these props are applied again. 

It seems that props are not being applied again which causes the autofocus not being honored.

## Test Plan

Autofocus=true && useCamera2Api=true
Screenshot_20191122-160110

Autofocus=true && useCamera2Api=false
Screenshot_20191122-155740

Simple RN application which uses react-native-camera to record video. 

Set autoFocus={RNCamera.Constants.AutoFocus.on} and during preview/recording you can see that autofocus works.
Now set useCamera2Api={true} and autofocus is not working anymore.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |      ❌     |
| Android |    ✅    |

## Checklist

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera2.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera2.java
@@ -436,7 +436,7 @@ class Camera2 extends CameraViewImpl implements MediaRecorder.OnInfoListener, Me
             mStillImageReader.close();
         }
         if (size == null) {
-          if (mAspectRatio == null) {
+          if (mAspectRatio == null || mPictureSize == null) {
             return;
           }
           mPictureSizes.sizes(mAspectRatio).last();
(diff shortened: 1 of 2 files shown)


sibelius:
@cristianoccazinsp  what your thoughts on this?

cristianoccazinsp:
I have honestly not been using Camera2, but the changes seem right. You can see that the constructor from https://github.com/react-native-community/react-native-camera/blob/4b0a27dd46e0db0167486ca2ce6b193cf991a8a5/android/src/main/java/com/google/android/cameraview/CameraView.java#L109 is instantiated with `fallbackToOldApi=true` (https://github.com/react-native-community/react-native-camera/blob/4b0a27dd46e0db0167486ca2ce6b193cf991a8a5/android/src/main/java/org/reactnative/camera/RNCameraView.java#L71), so in theory, Camera1 is always used first, then changed to Camera2. This is probably because the prop setters are only called after creating the view and there's no way to get the props before that.

Issue #2528, opened 2019-10-09
Error while updating property 'cameraId' 


dcalixto:
After update to react-native-camera: "^3.6.0" and pass RN version 59.9 to androidX using jetifier and adding 
android.useAndroidX=true
android.enableJetifier=true 

to gradle.properties and implementation "androidx.appcompat:appcompat:1.0.0"
to android/build.gradle

and  maven { url "https://jitpack.io" }
        maven { url "https://maven.google.com" } 

to android/app/build.gradle

the cam start tl show:

> Error while updating property 'cameraId' of a view managed by: RNCamera

Someone facing the same issue?

sibelius:
cc @cristianoccazinsp

cristianoccazinsp:
Is there a stack trace? Can you check Android Studio / Logcat?

Also, shouldn't androidX be for RN >= 0.60?  If you are using RN < 0.60, you need the "reverse" jetifier to go from androidX to appcompat. Correct me if I'm wrong.

Also as written in the migration helper from RN (https://react-native-community.github.io/upgrade-helper/?from=0.59.9&to=0.60.6), you shouldn't be adding appcompat at all if using RN >= 0.60

gaykov:
I see the same crashes from time to time in Crashlytics. I wasn't able to reproduce them locally, but I noticed that they happen only on Android 4 and only on Samsung devices (maybe it will help to get some clue).

I'm on React Native `0.61.2` and `react-native-camera` is `3.8.0`.

I'm also able to attach a stack trace:

```
Fatal Exception: com.facebook.react.bridge.JSApplicationIllegalArgumentException: Error while updating property 'cameraId' of a view managed by: RNCamera
       at com.facebook.react.uimanager.ViewManagersPropertyCache$PropSetter.updateViewProp(ViewManagersPropertyCache.java:98)
       at com.facebook.react.uimanager.ViewManagerPropertyUpdater$FallbackViewManagerSetter.setProperty(ViewManagerPropertyUpdater.java:134)
       at com.facebook.react.uimanager.ViewManagerPropertyUpdater.updateProps(ViewManagerPropertyUpdater.java:54)
       at com.facebook.react.uimanager.ViewManager.updateProperties(ViewManager.java:48)
       at com.facebook.react.uimanager.NativeViewHierarchyManager.createView(NativeViewHierarchyManager.java:278)
[...]

Issue #2634, opened 2019-12-09
dont work on android 4.2.2 


moh76:
hi 
 <RNCamera
          ref={ref => {
            this.camera = ref;
          }}
          style={{
            flex: 1,
            width: '100%',
          }}
          onGoogleVisionBarcodesDetected={this.barcodeRecognized}
   > </RNCamera> 
        

this code worked on android v7 but on v4.2.2 show below error

"error while updating property 'cameraId' of a view managed by: RNCamera"

how fix it ?

cristianoccazinsp:
Can you try without barcodes first? Can you get an error stack trace from android studio / logcat?

moh76:
i tried #2635 and it worked 

thank you 

> Can you try without barcodes first? Can you get an error stack trace from android studio / logcat?
> 
> Related issue: #2528
> 
> Not sure why it was closed, did we ever get a PR to fix it? In short, this will crash with Android below API level 19 (Android 4.4).
> 
> Update: Can you try with this PR? #2635

PR #2635, opened 2019-12-10
Android API < 19 crash fix

cristianoccazinsp:
# Summary

`Objects.equals` was added on Android API 19 (4.4), and RN supports API 16, so we need to make a helper instead of using Objects so the camera doesn't crash on Android < 4.4.

Fixes https://github.com/react-native-community/react-native-camera/issues/2528 and https://github.com/react-native-community/react-native-camera/issues/2634 

Thanks @gaykov  for the replacement code.

## Test Plan

Tested on Android 8.1 and 10 (devices), and simulator with 4.2.2

-

-

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -34,11 +34,12 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.reactnative.camera.utils.ObjectUtils;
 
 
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/google/android/cameraview/Camera1.java
```diff
@@ -34,12 +34,13 @@
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.reactnative.camera.utils.ObjectUtils;
```
sibelius: you could also use this https://developer.android.com/reference/android/support/v4/util/ObjectsCompat

Issue #2656, opened 2019-12-27
Android crash - java.lang.NullPointerException


cristianoccazinsp:
# Bug Report

**To Do First**

- [X] Did you try latest release?
- [X] Did you try master?
- [X] Did you look for existing matching issues?

**Platforms**

Android

**Versions** 

- Android: Android 9 - Xioami Redmi Note 5 Pro
- react-native-camera: 3.15.0
- react-native: 0.61.5

**Description/Current Behaviour**
Just got this crash from the error reporting too:

```
java.lang.NullPointerException Attempt to invoke interface method 'java.lang.Object java.util.SortedSet.last()' on a null object reference 
    Camera1.java:948 com.google.android.cameraview.Camera1.adjustCameraParameters
    Camera1.java:162 com.google.android.cameraview.Camera1$2.run
    Handler.java:873 android.os.Handler.handleCallback
    Handler.java:99 android.os.Handler.dispatchMessage
    Looper.java:201 android.os.Looper.loop
    HandlerThread.java:65 android.os.HandlerThread.run
```

Which traces back to https://github.com/react-native-community/react-native-camera/blob/master/android/src/main/java/com/google/android/cameraview/Camera1.java#L948

Not entirely sure what's the issue, but looks like a missing null check. Why the picture sizes would be null is unknown, but I suspect an unsupported aspect ratio or something. It would be good for the code to not crash and rather fallback to some value.

**Expected Behaviour**

No Crash

**Steps to Reproduce**
Mount camera and capture. Possibly with an unsupported aspect ratio (e.g., 1:1)

**Additionals**

> Love react-native-camera? Please consider supporting our collective: 👉 https://opencollective.com/react-native-camera/donate
> Want this issue to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

PR #2662, opened 2019-12-30
Android crash fixes.

cristianoccazinsp:
# Summary
Make sure no unsupported aspect ratio is used, and do not crash when there are no cameras available.

Fixes https://github.com/react-native-community/react-native-camera/issues/2656 and prevents a crash if the device has no cameras.

## Test Plan
Tested on Motorola G5 and Google Pixel 2.

**Note**: I haven't been able to actually reproduce the crash from #2656 since it happened on a xioami device. Still waiting on the owner of the device to test it, but I'm confident the fix should be good. Will update when I get the test results back.

Real device

Test with various / unsupported aspect ratios.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist


cristianoccazinsp:
Just got confirmation that the crash no longer happens on the xiaomi device.

PR #2681, opened 2020-01-21
fix(RNCameraManager) expose videoStabilizationMode from native

mauriciopf:
### Motivation

Working on an enterprise project using RNCamera while setting **videoStabilizationMode** prop specifically for iOS, it doesn't seem to have any effect on the shakiness from the video, even though the implementation for setting the stabilization mode is correct  [connection setPreferredVideoStabilizationMode:self.videoStabilizationMode]; .

**self.videoStabilizationMode is always 0**

### Solution

Expose the videoStabilizationMode property from native code to JS.

### Steps to reproduce

Start recording video with the **recordAsync** method and the response from that video should respect the option that you assigned by the prop

- `RNCamera.Constants.VideoStabilization['off']`
- `RNCamera.Constants.VideoStabilization['standard']`
- `RNCamera.Constants.VideoStabilization['cinematic']`
- `RNCamera.Constants.VideoStabilization['auto']`

### Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### Checklist


sibelius:
can you also add to the docs?

mauriciopf:
> can you also add to the docs?
@sibelius that `prop` is already included in the [docs](https://github.com/react-native-community/react-native-camera/blame/master/docs/RNCamera.md#L331) I also updated the declaration file for typescript.

PR #2711, opened 2020-02-21
Expose an ability to change the frames per second.

cinjon:
# Summary

This PR implements a change to frames per second for iOS. There is a new `fps` option in recording that lets the dev do this. It tries to stay true to the current selected quality, so it will forgo changing the fps if the quality does not allow for it.

## Test Plan

It's been working in production for my app, however I have not had a chance to write tests for this and produces videos that have higher FPS. Example ffmpeg output:

```
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from 'Documents/2-21-2020_11-25-34-AM.mov':
  Metadata:
    major_brand     : qt
    minor_version   : 0
    compatible_brands: qt
    creation_time   : 2020-02-21T16:25:35.000000Z
  Duration: 00:00:08.98, start: 0.000000, bitrate: 8822 kb/s
    Stream #0:0(und): Video: hevc (Main) (hvc1 / 0x31637668), yuvj420p(pc, smpte170m/bt709/bt709), 1280x720, 8727 kb/s, 60.01 fps, 60 tbr, 600 tbn, 600 tbc (default)
    Metadata:
      creation_time   : 2020-02-21T16:25:35.000000Z
      handler_name    : Core Media Video
      encoder         : HEVC
    Stream #0:1(und): Audio: aac (LC) (mp4a / 0x6134706D), 44100 Hz, mono, fltp, 82 kb/s (default)
    Metadata:
      creation_time   : 2020-02-21T16:25:35.000000Z
      handler_name    : Core Media Audio

```

Add in `fps = 60` to the recordOptions passed to `recordAsync` on a compatible device+camera. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

- [X ] I have tested this on a device and a simulator
- [X ] I updated the typed files (TS and Flow)

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -1050,10 +1050,7 @@ BOOL _sessionInterrupted = NO;
     }
     [connection setVideoOrientation:orientation];
 
-
-
     BOOL recordAudio = [options valueForKey:@"mute"] == nil || ([options valueForKey:@"mute"] != nil && ![options[@"mute"] boolValue]);
-
 
     // sound recording connection, we can easily turn it on/off without manipulating inputs, this prevents flickering.
     // note that mute will also be set to true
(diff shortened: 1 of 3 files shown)


cinjon:
(Major credit to @zviadm and @pdsouza)

cristianoccazinsp:
How hard would it be to implement it for Android as well?

cinjon:
Re Android, I am not sure - I have not developed for Android before.

cinjon:
Re FPS after the codec, why should that be?

cristianoccazinsp:
> Re FPS after the codec, why should that be?

Have you tried it? Codecs might impact FPS, although I'm just guessing, we saw some bug reports regarding codecs and FPS before. The device might not be able to keep up high FPS without HVEC codec.

cinjon:
I have a device for which this is true. The result when using H264 was to just fallback to 30fps. This didn't seem to be an issue.

fabriziobertoglio1987:
I'm interested on the android side, as my understanding [android.hardware.Camera][1] api does not allow you to set the frame per second for recording (but for preview [setPreviewFrameRate][2]), but the newer api [`android.harware.Camera2`][2] has this functionality.

I am interested in writing a pull request

[1]: https://developer.android.com/reference/android/hardware/Camera.Parameters
[2]: https://developer.android.com/reference/android/hardware/Camera.Parameters#setPreviewFrameRate(int)

cristianoccazinsp:
That's a bummer! Camera2 is still really bad. So it cant be done with
Camera1?

sibelius:
what about CameraX?

cristianoccazinsp:
Wouldn't we require a complete re-write of either Camera1 or Camera2 to implement it with CameraX? The good side of that would be that we could get rid of both Camera1 and 2 and only use CameraX (which seems to already wrap both?). But the interface is still different so there's a re-write involved.

sibelius:
https://developer.android.com/training/camerax

android 5, api level 21

my minSDK right now is level 16, not sure if it is worth the bump

or can we make it optional?

cristianoccazinsp:
You are right, there's also that. I guess we need to stick with Camera1/Camera2 for a while until react native decides to bump the min SDK to 21. I would definitely not recommend implementing yet another CameraX "mode" to end up with 3 different implementations, it's already a pain to maintain like it is right now :(

cristianoccazinsp:
@fabriziobertoglio1987 Wouldn't the frame rate be set with in the MediaRecorder object?

https://developer.android.com/reference/android/media/MediaRecorder#setVideoFrameRate(int)

I really have no time to test that right now, but perhaps that works with Camera1. I also don't think I have a camera with 60FPS recording around with Android like to test it :(

fabriziobertoglio1987:
You are right cristianoccazinsp. Thanks a lot. I just add that I will try to implement this feature. 
I have an android phone and I can test this feature at lower then 60fps (I tested with this [instructions][3]) and I will give it a try. My phone crashes when recording at 60fps.

[3]: https://www.maketecheasier.com/shoot-cinematic-video-android/

cristianoccazinsp:
I will be more than happy to review the changes. Passing the prop shouldn't be too hard (just some copy paste of existing code). Hardest part is to test both Camera1 and Camera2 implementations :)

sibelius:
can we merge the iOS version now, and wait for the android parity?

cristianoccazinsp:
> can we merge the iOS version now, and wait for the android parity?

I agree with that!

bkljajic:
This PR seems to break focusing on iOS. I was able to narrow it down to this one after testing versions in sequence.
It produces a somewhat pulsating motion while focusing on the subject.
Any idea what the reason might be?
I've described it in this [issue](https://github.com/react-native-camera/react-native-camera/issues/3231)

Issue #2669, opened 2020-01-11
React Native for Windows vNext support for react-native-camera


stmoy:
# React Native for Windows vNext support

**Describe the Feature**

[React Native for Windows](https://github.com/microsoft/react-native-windows) enables React Native apps to target Windows via UWP. React Native for Windows is under active development by Microsoft. Most recently, the platform has been updated to leverage high-performance C++ (known as "RNW vNext"). Now that the RNW vNext supports native modules as a platform, we have identified several key native modules to add Windows support for (including react-native-camera) to help validate that the platform has the right set of capabilities and to bootstrap the community of modules that target Windows.

**Possible Implementations**

Microsoft devs are signed up to help build out this implementation, but we wanted to track this work here as well to help provide some transparency that we're doing this work and hoping to contribute back to the repo. 

**Related Issues**

https://github.com/microsoft/react-native-windows/issues/2550

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate
> Want this feature to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

stmoy:
Also: hello from the React Native Windows team here at Microsoft 👋

PR #2723, opened 2020-03-03
Add Cpp/WinRT camera module support for React Native Windows

ddalp:
# Summary

#2669 Adding C++/WinRT Camera module support for RN windows.
The current CSharp version of camera module for react native windows will be retired, due to:
1. RNW is now built on C++/WinRT with a new set of Native Module interfaces. The current .Net version of RNW will not be supported anymore.
2. There are issues with the current CSharp version of Camera library, it has not been updated for a while and it is currently broken with the latest RNCamera.

This change introduce a new camera library based on C++/WinRT and is compatible with the latest  RNW and RNC. It supports most of functionalities what current CSharp library does, with one exception that bar code reader has not been implemented yet. We will leave the CSharp library in the depot for now(with some simple change so it works with current RNC again). The plan is to add the code reader support (pending integrating with a C++ version of ZXing) to C++/WinRT library then delete CSharp library completely.

Also made a simple change to RNCamera.js to call a async checkMediaCapturePermission method on Windows.

Only need to test with Windows, the change to RNCamera.js is scoped with platform check.
Tested on windows with latest RNW cli generated App, major test scenarios are:
1. Snapping a picture/ recording a video to disk or to memory.
2. Tested on a tablet for orientation sensor and display orientation scenarios.
3. Tested device going in and out of connected standby and verified the video preview feature still works.
4. Tested switching back and front cameras, turning on/off flash etc...
N/A not a bug.
## Compatibility
N/A
| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

- [x ] I added the documentation in `README.md`
- [ ?] I mentioned this change in `CHANGELOG.md`
Will figure out how to update changelog.md
- [x ] I updated the typed files (TS and Flow)
not needed
- [x ] I added a sample use of the API in the example project (`example/App.js`)
not needed

### DIFF
--- a/windows/RNCamera/RCTCameraModule.cs
+++ b/windows/RNCamera/RCTCameraModule.cs
@@ -24,8 +24,6 @@ namespace RNCamera
         public const int CameraAspectFill = 0;
         public const int CameraAspectFit = 1;
         public const int CameraAspectStretch = 2;
-        public const int CameraCaptureModeStill = 0;
-        public const int CameraCaptureModeVideo = 1;
         public const int CameraCaptureTargetMemory = 0;
         public const int CameraCaptureTargetDisk = 1;
         public const int CameraCaptureTargetCameraRoll = 2;
@@ -64,7 +62,7 @@ namespace RNCamera
         {
             get
             {
-                return "CameraModule";
+                return "RNCameraModule";
             }
         }
 
@@ -77,8 +75,7 @@ namespace RNCamera
                     { "Aspect", GetAspectConstants() },
                     { "BarCodeType", GetBarcodeConstants() },
                     { "Type", GetTypeConstants() },
-                    { "CaptureQuality", GetCaptureQualityConstants() },
-                    { "CaptureMode", GetCaptureModeConstants() },
+                    { "VideoQuality", GetCaptureQualityConstants() },
                     { "CaptureTarget", GetCaptureTargetConstants() },
                     { "Orientation", GetOrientationConstants() },
                     { "FlashMode", GetFlashModeConstants() },
@@ -90,9 +87,26 @@ namespace RNCamera
         public CameraForViewManager CameraManager { get; } = new CameraForViewManager();
 
         [ReactMethod]
-        public async void capture(JObject options, IPromise promise)
+        public async void checkMediaCapturePermission(IPromise promise)
         {
-            var viewTag = options.Value<int>("view");
+            MediaCapture newCapture = new MediaCapture();
+            try
+            {
+                await newCapture.InitializeAsync();
+            }
+            catch
+            {
+                promise.Resolve(false);
+                return;
+            }
+
+            promise.Resolve(true);
+        }
+
+        [ReactMethod]
+        public async void takePicture(JObject options, int viewTag, IPromise promise)
+        {
+            //var viewTag = options.Value<int>("view");
             var cameraForView = CameraManager.GetCameraForView(viewTag);
             if (cameraForView == null)
             {
@@ -100,24 +114,30 @@ namespace RNCamera
                 return;
             }
 
-            var mode = options.Value<int>("mode");
-            if (mode == CameraCaptureModeVideo)
-            {
-                if (_recordingTask != null)
-                {
-                    promise.Reject("Cannot run more than one capture operation.");
-                    return;
-                }
+            await CapturePhotoAsync(cameraForView, options, promise).ConfigureAwait(false);
+        }
 
-                _recordingCancellation = new CancellationTokenSource();
-                _recordingTask = RecordAsync(cameraForView, options, promise, _recordingCancellation.Token);
+        [ReactMethod]
+        public void record(JObject options, int viewTag, IPromise promise)
+        {
+            //var viewTag = options.Value<int>("view");
+            var cameraForView = CameraManager.GetCameraForView(viewTag);
+            if (cameraForView == null)
+            {
+                promise.Reject("No camera found.");
                 return;
             }
-            else
+
+            if (_recordingTask != null)
             {
-                await CapturePhotoAsync(cameraForView, options, promise).ConfigureAwait(false);
+                promise.Reject("Cannot run more than one capture operation.");
+                return;
             }
+
+            _recordingCancellation = new CancellationTokenSource();
+            _recordingTask = RecordAsync(cameraForView, options, promise, _recordingCancellation.Token);
         }
+
 
         [ReactMethod]
         public async void stopCapture(IPromise promise)
@@ -415,15 +435,6 @@ namespace RNCamera
                 { "high", CameraCaptureQualityHigh },
                 { "720p", CameraCaptureQuality720p },
                 { "1080p", CameraCaptureQuality1080p },
-            };
-        }
-
-        private static IReadOnlyDictionary<string, object> GetCaptureModeConstants()
-        {
-            return new Dictionary<string, object>
-            {
-                { "still", CameraCaptureModeStill },
-                { "video", CameraCaptureModeVideo },
             };
         }
 
--- a/windows/ReactNativeCameraCPP/CameraRotationHelper.cpp
+++ b/windows/ReactNativeCameraCPP/CameraRotationHelper.cpp
@@ -78,7 +78,7 @@ namespace winrt::ReactNativeCameraCPP::implementation
         return result;
     }
 
-    int CameraRotationHelper::GetConvertedCameraPreviewOrientation()
+    int CameraRotationHelper::GetCameraPreviewClockwiseDegrees()
     {
         auto rotation = GetCameraPreviewOrientation();
         return ConvertSimpleOrientationToClockwiseDegrees(rotation);
--- a/windows/ReactNativeCameraCPP/CameraRotationHelper.cpp
+++ b/windows/ReactNativeCameraCPP/CameraRotationHelper.cpp
@@ -1,0 +1,242 @@
+﻿#include "pch.h"
+#include "CameraRotationHelper.h"
+#include "CameraRotationHelper.g.cpp"
+
+namespace winrt {
+    using namespace Windows::Devices::Enumeration;
+    using namespace Windows::Devices::Sensors;
+    using namespace Windows::Storage::FileProperties;
+    using namespace Windows::Graphics::Display;
+} //namespace winrt
+
+namespace winrt::ReactNativeCameraCPP::implementation
+{
+    CameraRotationHelper::CameraRotationHelper(EnclosureLocation location)
+    {
+        m_cameraEnclosureLocation = location;
+
+        if (!IsEnclosureLocationExternal(m_cameraEnclosureLocation) && m_orientationSensor != nullptr)
+        {
+            m_sensorOrientationChanged_revoker = m_orientationSensor.OrientationChanged(winrt::auto_revoke, [ref = get_weak()](auto const& sender, auto const& args) 
+            {
+                if (auto self = ref.get()) {
+                    self->SimpleOrientationSensor_OrientationChanged(sender, args);
+                }
+            });
+        }
+        m_displayOrientationChanged_revoker = m_displayInformation.OrientationChanged(winrt::auto_revoke, [ref = get_weak()](auto const& sender, auto const& args)
+        {
+            if (auto self = ref.get()) {
+                self->DisplayInformation_OrientationChanged(sender, args);
+            }
+        });
+    }
+
+    PhotoOrientation CameraRotationHelper::GetConvertedCameraCaptureOrientation()
+    {
+        auto orientation = GetCameraCaptureOrientation();
+        return ConvertSimpleOrientationToPhotoOrientation(orientation);
+    }
+
+    SimpleOrientation CameraRotationHelper::GetCameraCaptureOrientation()
+    {
+        if (IsEnclosureLocationExternal(m_cameraEnclosureLocation))
+        {
+            // Cameras that are not attached to the device do not rotate along with it, so apply no rotation
+            return SimpleOrientation::NotRotated;
+        }
+
+        // Get the device orientation offset by the camera hardware offset
+        auto deviceOrientation = m_orientationSensor ? m_orientationSensor.GetCurrentOrientation() : SimpleOrientation::NotRotated;
+        auto result = SubtractOrientations(deviceOrientation, GetCameraOrientationRelativeToNativeOrientation());
+
+        // If the preview is being mirrored for a front-facing camera, then the rotation should be inverted
+        if (ShouldMirrorPreview())
+        {
+            result = MirrorOrientation(result);
+        }
+        return result;
+    }
+
+    SimpleOrientation CameraRotationHelper::GetCameraPreviewOrientation()
+    {
+        if (IsEnclosureLocationExternal(m_cameraEnclosureLocation))
+        {
+            // Cameras that are not attached to the device do not rotate along with it, so apply no rotation
+            return SimpleOrientation::NotRotated;
+        }
+
+        // Get the app display rotation offset by the camera hardware offset
+        auto result = ConvertDisplayOrientationToSimpleOrientation(m_displayInformation.CurrentOrientation());
+        result = SubtractOrientations(result, GetCameraOrientationRelativeToNativeOrientation());
+
+        // If the preview is being mirrored for a front-facing camera, then the rotation should be inverted
+        if (ShouldMirrorPreview())
+        {
+            result = MirrorOrientation(result);
+        }
+        return result;
+    }
+
+    int CameraRotationHelper::GetConvertedCameraPreviewOrientation()
+    {
+        auto rotation = GetCameraPreviewOrientation();
+        return ConvertSimpleOrientationToClockwiseDegrees(rotation);
+    }
+
+    PhotoOrientation CameraRotationHelper::ConvertSimpleOrientationToPhotoOrientation(SimpleOrientation orientation)
+    {
+        switch (orientation)
+        {
+            case SimpleOrientation::Rotated90DegreesCounterclockwise:
+                return PhotoOrientation::Rotate90;
+            case SimpleOrientation::Rotated180DegreesCounterclockwise:
+                return PhotoOrientation::Rotate180;
+            case SimpleOrientation::Rotated270DegreesCounterclockwise:
+                return PhotoOrientation::Rotate270;
+            case SimpleOrientation::NotRotated:
+            default:
+                return PhotoOrientation::Normal;
+        }
+    }
+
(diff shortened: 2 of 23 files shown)


windows/ReactNativeCameraCPP/CameraRotationHelper.cpp
```diff
@@ -0,0 +1,242 @@
...
+        }
+        return orientation;
+    }
+
+    SimpleOrientation CameraRotationHelper::AddOrientations(SimpleOrientation a, SimpleOrientation b)
+    {
+        auto aRot = ConvertSimpleOrientationToClockwiseDegrees(a);
+        auto bRot = ConvertSimpleOrientationToClockwiseDegrees(b);
+        auto result = (aRot + bRot) % 360;
+        return ConvertClockwiseDegreesToSimpleOrientation(result);
```
asklar: can either of these be negative, if so do we need to +360 before doing %360?

windows/ReactNativeCameraCPP/CameraRotationHelper.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+        // If the preview is being mirrored for a front-facing camera, then the rotation should be inverted
+        if (ShouldMirrorPreview())
+        {
+            result = MirrorOrientation(result);
+        }
+        return result;
+    }
+
+    int CameraRotationHelper::GetConvertedCameraPreviewOrientation()
```
asklar: >GetConvertedCameraPreviewOrientation [](start = 30, length = 36)

This returns an int but it isn't clear from the name of the method that the int represents an angle in degrees

windows/ReactNativeCameraCPP/CameraRotationHelper.h
```diff
@@ -0,0 +1,61 @@
...
+
+        winrt::Windows::Devices::Sensors::SimpleOrientation SubtractOrientations(
+            winrt::Windows::Devices::Sensors::SimpleOrientation a,
+            winrt::Windows::Devices::Sensors::SimpleOrientation b);
+        winrt::Windows::Devices::Sensors::SimpleOrientation MirrorOrientation(winrt::Windows::Devices::Sensors::SimpleOrientation orientation);
+        winrt::Windows::Devices::Sensors::SimpleOrientation AddOrientations(winrt::Windows::Devices::Sensors::SimpleOrientation a, winrt::Windows::Devices::Sensors::SimpleOrientation b);
+
+        winrt::Windows::Storage::FileProperties::PhotoOrientation ConvertSimpleOrientationToPhotoOrientation(
+            winrt::Windows::Devices::Sensors::SimpleOrientation orientation);
+        int ConvertSimpleOrientationToClockwiseDegrees(winrt::Windows::Devices::Sensors::SimpleOrientation orientation);
```
asklar: >ConvertSimpleOrientationToClockwiseDegrees [](start = 12, length = 42)

this should probably be a static as it doesn't access any of the members in the class

windows/ReactNativeCameraCPP/ReactCameraConstants.h
```diff
@@ -0,0 +1,140 @@
...
+
+    static std::map<std::wstring, int>   GetTorchModeConstants() noexcept {
+        return std::map<std::wstring, int>
+        {
+            { L"off", CameraTorchModeOff },
+            { L"on", CameraTorchModeOn },
+            { L"auto", CameraTorchModeAuto },
+        };
+    }
+
```
asklar: this is pretty wasteful and increases load/startup time
can we design this differently?

windows/ReactNativeCameraCPP/ReactCameraView.cpp
```diff
@@ -0,0 +1,537 @@
...
+        auto props = co_await storageFile.Properties().GetImagePropertiesAsync();
+        auto searchTitle = options.find(L"title");
+        if (searchTitle != options.end()) {
+            const auto& titleValue = options.at(L"title");
+            if (titleValue.Type() == JSValueType::String)
+            {
+                auto titleString = titleValue.String();
+                props.Title(winrt::to_hstring(titleString));
+            }
+            co_await props.SavePropertiesAsync();
```
asklar: else throw?

windows/ReactNativeCameraCPP/ReactCameraModule.h
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+#pragma endregion
+
+    public:
+        RNCameraModule() {
+        }
+
+        ~RNCameraModule() {
+        }
```
asklar: can write as RNCameraModule() = default; 
same with dtor

windows/ReactNativeCameraCPP/ReactCameraView.cpp
```diff
@@ -0,0 +1,537 @@
...
+        }
+    }
+    
+    // RNW has a bug where the numeric value is set as int in debug but double in release
+    // ToDo: remove this function once bug https://github.com/microsoft/react-native-windows/issues/4225 is fixed.
+    bool ReactCameraView::GetNumericValue(std::map<std::wstring, JSValue> const& options, const std::wstring key, int &value)
```
asklar: >GetNumericValue [](start = 26, length = 15)

TryGetInt ? or TryGetValueAsInt

windows/ReactNativeCameraCPP/ReactCameraConstants.h
```diff
@@ -0,0 +1,140 @@
+#pragma once
+
+#include <functional>
+
+namespace winrt::ReactNativeCameraCPP
+{
+    static const int CameraAspectFill = 0;
+    static const int CameraAspectFit = 1;
+    static const int CameraAspectStretch = 2;
```
asklar: do these need to actually be ints? can't these be values in an enum class? same with others?

windows/ReactNativeCameraCPP/ReactCameraViewManager.cpp
```diff
@@ -0,0 +1,156 @@
...
+} //namespace winrt
+
+
+namespace winrt::ReactNativeCameraCPP::implementation {
+    // static vector to contain all currently active camera view instances
+    // This is a temporary workaround to map the view tag to view instance, since
+    // Method call from ReactCameraModule constains view tag as parameter and we need
+    // to forward the call to CameraView instance.
+    std::vector<winrt::com_ptr<ReactNativeCameraCPP::ReactCameraView>> ReactCameraViewManager::m_cameraViewInstances;
+
```
asklar: is it even possible to have more than one? if i understand correctly you can only use the camera in one view in winrt

windows/ReactNativeCameraCPP/ReactCameraViewManager.cpp
```diff
@@ -0,0 +1,156 @@
...
+        winrt::Microsoft::ReactNative::ReactPromise<bool>& result)
+    {
+        auto mediaCapture = winrt::Windows::Media::Capture::MediaCapture();
+        bool hasPermission = true;
+        try {
+            co_await mediaCapture.InitializeAsync();
+        }
+        catch (winrt::hresult_error const&) {
+            hasPermission = false;
+        }
```
asklar: should prob be a debug spew here

windows/RNCamera/RCTCameraModule.cs
```diff
@@ -90,35 +87,58 @@ public override string Name
...
-            var viewTag = options.Value<int>("view");
+            MediaCapture newCapture = new MediaCapture();
+            try
+            {
+                await newCapture.InitializeAsync();
+            }
+            catch
+            {
+                promise.Resolve(false);
+                return;
```
asklar: debug print?

windows/RNCamera/RCTCameraModule.cs
```diff
@@ -90,35 +87,58 @@ public override string Name
...
-                _recordingCancellation = new CancellationTokenSource();
-                _recordingTask = RecordAsync(cameraForView, options, promise, _recordingCancellation.Token);
+        [ReactMethod]
+        public void record(JObject options, int viewTag, IPromise promise)
+        {
+            //var viewTag = options.Value<int>("view");
+            var cameraForView = CameraManager.GetCameraForView(viewTag);
+            if (cameraForView == null)
+            {
+                promise.Reject("No camera found.");
```
asklar: >. [](start = 47, length = 1)

you used ! on the C++ side but . here :) standardize on one

windows/ReactNativeCameraCPP/CameraRotationHelper.cpp
ddalp: No, the value will be >= 0, either 0,90,180 or 270. The rotation helper code are exact copy from MSDN, except I converted it from CSharp to C++/WinRT.

windows/ReactNativeCameraCPP/CameraRotationHelper.cpp
ddalp: Good catch, I shall rename the function (GetCameraPreviewClockwiseDegrees), it is about only few places I made changes to the sample code.

windows/ReactNativeCameraCPP/CameraRotationHelper.h
ddalp: will do

windows/ReactNativeCameraCPP/ReactCameraConstants.h
ddalp: This is to adhere to the current native module interface design, to return json values for contant definition. We can probably delay creating the std::maps when queried, but still needs time to create them at runtime.

windows/ReactNativeCameraCPP/ReactCameraView.cpp
ddalp: Sounds good!

windows/ReactNativeCameraCPP/ReactCameraConstants.h
ddalp: We could, but I will have to cast it twice. Once from module to RN and once from RN back to module since RN only deal with number. That's why I created a reactconstants.h and defined everything as numeric there.

windows/ReactNativeCameraCPP/ReactCameraViewManager.cpp
ddalp: You can have multiple views in your RN app, for example displaying both front and back cameras at same time. Hence in that situation, we have will two ReactCameraView instances, each is backed by a seperate CaptureElement.

Since we only have one camera module instance, and there are method calls on the module instance to deal with camera view object (using a viewTag parameter). We need a static vector for us to find the view object from the view tage, eventually(not yet supported) Native module interfaces should be able to get view instance from ReactContext object, and we can remove this vector and FindCamera logic in this native module.

windows/ReactNativeCameraCPP/ReactCameraViewManager.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+        nativeProps.Insert(L"aspect", ViewManagerPropertyType::Number);
+        nativeProps.Insert(L"type", ViewManagerPropertyType::Number);
+        nativeProps.Insert(L"autoFocus", ViewManagerPropertyType::Boolean);
+        nativeProps.Insert(L"whiteBalance", ViewManagerPropertyType::Number);
+        nativeProps.Insert(L"torchMode", ViewManagerPropertyType::Number);
+        nativeProps.Insert(L"flashMode", ViewManagerPropertyType::Number);
+        //        nativeProps.Insert(L"barcodeScannerEnabled", ViewManagerPropertyType::Boolean);
+        //        nativeProps.Insert(L"barCodeTypes", ViewManagerPropertyType::Array);
+        nativeProps.Insert(L"keepAwake", ViewManagerPropertyType::Boolean);
```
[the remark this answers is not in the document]
ddalp: It will be needed once we support bar code, that is for future improvements. We need to figure out a ZXing C++ library to do that.

windows/ReactNativeCameraCPP/ReactCameraViewManager.cpp
ddalp: Actually it is taken care by RN Camera, as it will show a "NotSupportedView" for RNCamera if this method returns no permission.

windows/RNCamera/RCTCameraModule.cs
ddalp: Same, a yellow box will be shown in debugger mode for RN app with the error message.

windows/RNCamera/RCTCameraModule.cs
ddalp: Fixed c++ side.

windows/ReactNativeCameraCPP/ReactCameraView.cpp
asklar: looks like a typo snuck in by 😛

---
In reply to: [389814407](https://github.com/react-native-community/react-native-camera/pull/2723#discussion_r389814407) [](ancestors = 389814407)

windows/ReactNativeCameraCPP/ReactCameraConstants.h
asklar: My bad - I read these as being members of a class but they're in a namespace.
I think what you had before was fine. In fact you could just say constexpr int CameraAspectFill = 0; if you wanted to be more explicit that this shouldn't take up any storage.

---
In reply to: [387861878](https://github.com/react-native-community/react-native-camera/pull/2723#discussion_r387861878) [](ancestors = 387861878)

windows/ReactNativeCameraCPP/ReactCameraConstants.h
```diff
@@ -0,0 +1,140 @@
...
+    static const int CameraWhiteBalanceFluorescent = 5;
+    static const int CameraCaptureQualityHigh = (int)winrt::Windows::Media::MediaProperties::VideoEncodingQuality::HD1080p;
+    static const int CameraCaptureQualityLow = (int)winrt::Windows::Media::MediaProperties::VideoEncodingQuality::HD720p;
+    static const int CameraCaptureQuality1080p = (int)winrt::Windows::Media::MediaProperties::VideoEncodingQuality::HD1080p;
+    static const int CameraCaptureQuality720p = (int)winrt::Windows::Media::MediaProperties::VideoEncodingQuality::HD720p;
+    static const int MediaTypeImage = 1;
+    static const int MediaTypeVideo = 2;
+
+#pragma warning( push )
+#pragma warning( disable : 4505 )
```
asklar: I think the reason you're having to put this here is because now that you marked these static you're getting the "global/namespace  static" semantics which are different from the "class static" semantics:

class static: you don't need to have an instance to call this method
global/namespace  static: this function won't be used outside of this file (or other files that include it)

so there is probably some cpp file that includes this header, but isn't using this function - so you get a warning saying hey did you mean to do that? because nobody in that CPP file is using this function and this function is marked as a global/namespace static function, meaning nobody else will be able to call them and the code will be optimized out because of it

I think these static methods in the namespace could just be constexpr, but if that doesn't work I think what you had before the static was ok (sorry again!!)

windows/ReactNativeCameraCPP/ReactCameraViewManager.cpp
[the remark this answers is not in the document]
asklar: thanks, great to know :) - could you link to a github issue for that?

---
In reply to: [389819655](https://github.com/react-native-community/react-native-camera/pull/2723#discussion_r389819655) [](ancestors = 389819655)

windows/ReactNativeCameraCPP/ReactCameraConstants.h
ddalp: Scoped constants to under a class to avoid issues with global name space

woodrufs:
Hi @ddalp & @asklar,
Sorry to bother you, but I hope I have just two quick questions for you both.

1) How come react-native-windows is not mentioned in [packge.json](https://github.com/react-native-community/react-native-camera/blob/master/package.json) anywhere? Shouldn't it be listed as a peer dependency?
2) Is there an ETA when barcode scanning will be supported in the new Cpp/WinRT implementation that can be used in v.61 of RNW?

Thank you in advanced.

asklar:
@woodrufs hmmm not really sure. CC @chrisglein for awareness in case he knows what the plan is

ddalp:
> @woodrufs hmmm not really sure. CC @chrisglein for awareness in case he knows what the plan is
Adding barcode functionality needs ZXing C++ support, there is one at https://github.com/glassechidna/zxing-cpp, not sure how easy/feasible to integrate with RNW C++ Camera though. Worst case, we can convert the old Windows Camera module which integrates with .net ZXing to latest RNW native module.

woodrufs:
@ddalp,
I really appreciate the reply, thank you. We use react-native-camera and its bar code scanning functionality within our Production app that is currently on v.59 that we are trying to get to v.61. If the previous Windows Camera module that integrates .net ZXing was converted to the latest RNW native module, MSFT's partner Plex Systems would be eternally grateful. :)

chrisglein:
> @ddalp Di Da FTE,
> I really appreciate the reply, thank you. We use react-native-camera and its bar code scanning functionality within our Production app that is currently on v.59 that we are trying to get to v.61. If the previous Windows Camera module that integrates .net ZXing was converted to the latest RNW native module, MSFT's partner Plex Systems would be eternally grateful. :)

We're trying to track Windows community module gaps and help out where we can. There are a lot of modules, so we have to balance our investment. Some rough info on our tracking is linked [here](https://microsoft.github.io/react-native-windows/docs/parity-status#supported-community-modules). I'll raise this with our prioritization (@stmoy FYI). But, for transparency, we're only able to fund so many ourselves and will be relying on broad community effort to scale up and get Windows support for most modules.

For this, probably open an issue on this repo to make sure the need is captured outside of the PR?

woodrufs:
@chrisglein,
I get where you are coming from and that you guys have a lot on your plate. I created a feature request [here](https://github.com/react-native-community/react-native-camera/issues/2830) for adding bar code scanning to Cpp/WinRT implementation. I wanted to mention this to you guys since Plex was mentioned as one of Microsoft's partners at Ignite 2019, and this is a critical piece of our application. 

MSFT Partners


Plex App

Our users are employees at manufacturing plants who rely on this to perform many of their functions, and without it, they are not very efficient. Users are often using this feature to scan bar codes multiple times a minute. Not only do they scan often, but sometimes the bar codes are extremely long. For instance, bar codes coming from Toyota's supply chain are over 250 characters... 

At this point, I am just trying to think of anything to say to get you guys to decide to do some sort of bar code support.. Please let me know if there is any way I can sell you guys on it! 🤪

Sorry if you're starting to get tired of my questions the last couple weeks. I hope I can become an asset to the rest of the community soon.. Thanks again for all the work you guys do.

Issue #2747, opened 2020-03-14
Shutter sound in Camera1 is broken for some devices


cristianoccazinsp:
# Bug Report

**To Do First**

- [X] Did you try latest release?
- [X] Did you try master?
- [X] Did you look for existing matching issues?

**Platforms**

Android 10

**Versions**

- Android: 10
- iOS:
- react-native-camera: 3.19.1

**Description/Current Behaviour**

Related to https://github.com/react-native-community/react-native-camera/issues/2521#issuecomment-598940977 discussion.

On some devices (most likely Android 10), Camera1's default behaviour is to have a shutter sound. However, `playSoundOnCapture` plays its own sound. As a result, when the prop is set to false, Camera1 will still play a sound, and even worse, when the prop is set to true, Camera1 will play two sounds (one from the manual call, and another one automatically). Lastly, the sound also plays at the wrong time (on capture start, rather than actual capture).

**Posisble fix**
The possible fixes are the following:

1) Move `playSoundOnCapture` to each Camera implementation. Camera2 would continue to play the sound as it currently does, while Camera1 would just call `enableShutterSound`. Note that in some countries, cameras are FORCED to play sound, and `enableShutterSound` will not always work in such countries (e.g., Tokyo, Japan).

2) Make Camera1 always call `enableShutterSound(false)`, and leave the shutter sound to RNCameraView as it currently does (although move it to the on picture taken event). This has the problem, however, that if the sound cannot be turned off, and someone uses `enableShutterSound=true`, we would end up with 2 sounds. For this reason, option 1 makes more sense.

**Additionals**

> Love react-native-camera? Please consider supporting our collective: 👉 https://opencollective.com/react-native-camera/donate
> Want this issue to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

cristianoccazinsp:
@noambonnie  this is related to your shutter sound issue. I might have a PR soon for it, the fix is relatively easy and you would be able to use Camera1 :)

cristianoccazinsp:
Closed with https://github.com/react-native-community/react-native-camera/pull/2748

Issue #2727, opened 2020-03-04
exposure value/prop not documented


cristianoccazinsp:
# Bug Report

Looks like the `exposure` prop is not documented. I've been trying to see if the camera supports brightness (exposure) adjustments, but when searching for it, the prop is poorly documented.

What are the possible values? How does it interact with auto focus point of interest?

> Love react-native-camera? Please consider supporting our collective: 👉 https://opencollective.com/react-native-camera/donate
> Want this issue to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

cristianoccazinsp:
Here's some info from iOS' source code. I guess Android's behaviour would be the same.


https://github.com/react-native-community/react-native-camera/blob/master/ios/RN/RNCamera.m#L645

```
/// Set the AVCaptureDevice's ISO values based on RNCamera's 'exposure' value,
/// which is a float between 0 and 1 if defined by the user or -1 to indicate that no
/// selection is active. 'exposure' gets mapped to a valid ISO value between the
/// device's min/max-range of ISO-values.
///
/// The exposure gets reset every time the user manually sets the autofocus-point in
/// 'updateAutoFocusPointOfInterest' automatically. Currently no explicit event is fired.
/// This leads to two 'exposure'-states: one here and one in the component, which is
/// fine. 'exposure' here gets only synced if 'exposure' on the js-side changes. You
/// can manually keep the state in sync by setting 'exposure' in your React-state
/// everytime the js-updateAutoFocusPointOfInterest-function gets called.
```

cristianoccazinsp:
Closed with https://github.com/react-native-community/react-native-camera/pull/2748

PR #2748, opened 2020-03-14
Fix android shutter sound

cristianoccazinsp:
# Summary

Fixes for Camera1 and Camera2 shutter sound. (https://github.com/react-native-community/react-native-camera/issues/2747). This includes making Camera1 and Camera2 behave consistently (when possible) and fixing MediaActionSound being instantiated each time and failing to play sound from time to time.

Shutter sound should now be consistent with the `playSoundOnCapture` prop. Additionally, it should happen on successful capture as opposed to camera capture start (consistent with iOS and any camera app)

Lastly, added missing props to the docs (https://github.com/react-native-community/react-native-camera/issues/2727), fixed issue related to double camera start (https://github.com/react-native-community/react-native-camera/issues/2521), and improved Camera2 to Camera1 fallback logic so Camera2 is not instantiated unnecessarily.

@noambonnie if you could give this PR a try. You shall now be able to use Camera1 and also `playSoundOnCapture=false` as long as your country allows for shutter sound to be disabled.

## Test Plan
Tested on Pixel 2 (Android 10) and Motorola G5 (Android 8).  Tested both Camera1 and Camera2 implementations. 

Real device

Test `playSoundOnCapture` true and false

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera2.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera2.java
@@ -305,6 +305,7 @@ class Camera2 extends CameraViewImpl implements MediaRecorder.OnInfoListener, Me
     boolean start() {
         if (!chooseCameraIdByFacing()) {
             mAspectRatio = mInitialRatio;
+            mCallback.onMountError();
             return false;
         }
         collectCameraInfo();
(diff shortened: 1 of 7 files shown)


android/src/main/java/com/google/android/cameraview/CameraView.java
```diff
@@ -311,7 +313,9 @@ public void setUsingCamera2Api(boolean useCamera2) {
             }
             mImpl = new Camera1(mCallbacks, mImpl.mPreview, mBgHandler);
         }
-        start();
+        if(wasOpened){
```
cristianoccazinsp: The following change is causing issues on devices that fallback to Camera1 even when Camera2 is used.... Implementing further fixes for those.

cristianoccazinsp: Changes done.

android/src/main/java/com/google/android/cameraview/Camera2.java
```diff
@@ -727,19 +770,16 @@ private boolean chooseCameraIdByFacing() {
                 int internalFacing = INTERNAL_FACINGS.get(mFacing);
                 final String[] ids = mCameraManager.getCameraIdList();
                 if (ids.length == 0) { // No camera
-                    throw new RuntimeException("No camera available.");
+                    Log.e(TAG, "No cameras available.");
```
MateusAndrade: What do you think about returning something to the JS? Some people don't know how to see the `Logcat` provided  by android. Removing the `throw`(that probably show which error triggered that) can lead some confusion.

cristianoccazinsp:
I believe the mount error callback will fire when that error happens. The
change was to prevent full app crashes when used on devices with no cameras
(making it behave like Camera1).

I agree about sending these errors to the JS side, but neither camera1 nor
2 have a way to pass down this error message through the mount error
callback. There would be a bunch of changes involved to do so (on both
camera implementations).  We could make another PR for that change, but
keep in mind both camera implementations needs to be changed.

MateusAndrade:
>  We could make another PR for that change, but keep in mind both camera implementations needs to be changed. 

Totally agree with that. Also, we will need to change the same behaviour on `ios`. Btw, this looks good for me. What do you think @sibelius , @fabriziobertoglio1987 ?

cristianoccazinsp:
It would be great to have @noambonnie test it on the Pixel 3 which seemed to be a problematic device.

noambonnie:
I tested this on the following devices/versions:
Galaxy S10/Android 10
Galaxy S8/Android 9
Pixel 3a/Android 10

- I can confirm the sound flag works as expected. 
- I can confirm this resolves the issue discussed in https://github.com/react-native-community/react-native-camera/issues/2521

I would love to see this merged. 

Thank you for your help!

* Unfortunately due to COVID19 I don't currently have access to other devices. I will try to go to the office sometime this week to test, unless someone else has access to older or different devices/versions.

fabriziobertoglio1987:
@MateusAndrade thanks sorry... I am not feeling 100% and I was trying to complete my work on another pull request... I see this is already merged so I will work on the other pull requests. Thanks a lot

MateusAndrade:
> @MateusAndrade thanks sorry... I am not feeling 100% and I was trying to complete my work on pull [2751](https://github.com/react-native-community/react-native-camera/pull/2751)... I see this is already merged so I will review the other pull requests. Thanks a lot

Sorry for that, i was thinking that we already got your review 😢

PR #2759, opened 2020-03-22
call onStatusChange after refreshAuthorizationStatus

seblau:
# Summary

Those changes make sure that `onStatusChanged` is called after `refreshAuthorizationStatus`. The component is listening on `onStatusChanged` in order to show or hide authorization infos. After navigating back from the settings screen (changing permissions), `refreshAuthorizationStatus` gets called, but any changes to that stay unnoticed.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/src/RNCamera.js
+++ b/src/RNCamera.js
@@ -741,11 +741,14 @@ export default class Camera extends React.Component<PropsType, StateType> {
       return;
     }
 
-    this.setState({
-      isAuthorized: hasCameraPermissions,
-      isAuthorizationChecked: true,
-      recordAudioPermissionStatus,
-    });
+    this.setState(
+      {
+        isAuthorized: hasCameraPermissions,
+        isAuthorizationChecked: true,
+        recordAudioPermissionStatus,
+      },
+      this._onStatusChange,
+    );
   }
 
   async componentDidMount() {


src/RNCamera.js
```diff
@@ -741,11 +741,14 @@ export default class Camera extends React.Component<PropsType, StateType> {
...
-      isAuthorizationChecked: true,
-      recordAudioPermissionStatus,
-    });
+    this.setState(
+      {
+        isAuthorized: hasCameraPermissions,
+        isAuthorizationChecked: true,
+        recordAudioPermissionStatus,
+      },
+      this._onStatusChange,
```
fabriziobertoglio1987: To give you my complete feedback:
I tested this functionality on android and I can not reproduce this issue on the master branch. 

I tested as follows the basic and advanced example:

1) run the app on the master branch
2) take a picture etc.
3) change the authorization settings
Settings->Security->Permissions->Apps->RNCameraExample
set Use Camera to Allow
4) Switch to the app and grant camera authorization
5) The camera shows the preview and I can take pictures/videos without any need for refresh

I had the chance with the advanced example to check `node_modules/react-native-camera/src/RNCamera.js` and your code is not included.

Possible reason this bug does not reproduce in my test are:

1) I am switching between the RNCamera example Activity and the Settings Activity, which could cause the app to refresh... while you could have experienced this using different steps and never left the Camera App Activity to change the authorization settings.

2) Another reason would be that the issue is only related to IOS.

3) The issue is related to code implemented outside of the react-native-camera library.

Thanks a lot
Fabrizio

cristianoccazinsp:
Does this mean that `onStatusChange` might fire more than once after mount?  Should there be checks to only fire it if the status did in fact change (e.g., compare prev/next status) ?

src/RNCamera.js
seblau: Hi Fabrizio,

actually it's not really a bug that this PR should resolve. The only thing is, that after `refreshAuthorizationStatus` gets called, the `onStatusChange` callback will never get called. The way my business logic is built, I am listening on the `onStatusChange` callback in order to trigger certain events if the camera status changes.

I was expecting this callback to fire on every status change, which it apparently does not do after a refreshAuthorizationStatus. Thats why I created this PR. I think it should be safe to assume that we just fire onStatusChange after refreshAuthorizationStatus?

fabriziobertoglio1987: @seblau Thanks for the reply. I am working on a PR myself and I'll come back testing this by calling `refreshAuthorizationStatus` instead of using the settings app from android to have the App request again authorization to [use the camera](https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fwww.swrve.com%2Fimages%2Fuploads%2Finner%2FScreen_Shot_2016-04-29_at_14.45.58.png&f=1&nofb=1).

fabriziobertoglio1987:
maybe cristianoccazinsp refers to the usage of [shouldComponentUpdate][https://reactjs.org/docs/react-component.html#shouldcomponentupdate] or something similar.

src/RNCamera.js
konstantinmv: Any update on this @fabriziobertoglio1987 ?

PR #2769, opened 2020-03-27
Accept path while taking picture in android

AsminBudha:
# Summary
With this PR, user are able to add a path options while taking picture same as recording picture. It will save image in given path i.e if it is file path then picture will be save with give in that location else if it is directory then name will be randomly generated.
 It impact the photo taking process.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/docs/RNCamera.md
+++ b/docs/RNCamera.md
@@ -51,7 +51,7 @@ class ExampleApp extends PureComponent {
     );
   }
 
-  takePicture = async() => {
+  takePicture = async () => {
     if (this.camera) {
       const options = { quality: 0.5, base64: true };
       const data = await this.camera.takePictureAsync(options);
@@ -217,8 +217,8 @@ On iOS, focusing will not change the exposure automatically unless autoExposure 
 Hint:
 for portrait orientation, apply 90° clockwise rotation + translation: [Example](https://gist.github.com/Craigtut/6632a9ac7cfff55e74fb561862bc4edb)
 
-
 ### iOS `onSubjectAreaChanged`
+
 iOS only.
 
 if autoFocusPointOfInterest is set, this event will be fired when a substancial change is detected with the following object: `{ nativeEvent: { prevPoint: { x: number; y: number; } } }`
@@ -287,6 +287,7 @@ The idea is that you select the appropriate white balance setting for the type o
 Use the `whiteBalance` property to specify which white balance setting the camera should use.
 
 ### `exposure`
+
 Value: float from `0` to `1.0`, or `-1` (default) for auto.
 
 Sets the camera's exposure value.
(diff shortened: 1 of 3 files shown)


docs/RNCamera.md
```diff
@@ -598,6 +599,8 @@ writeExif = {
 
 - `orientation` (string or number). Specifies the orientation that us used for taking the picture. Possible values: `"portrait"`, `"portraitUpsideDown"`, `"landscapeLeft"` or `"landscapeRight"`.
 
+- `path` (file path on disk). Specifies the path on disk to save picture to.
```
MateusAndrade: What about warning that setting this, can slow down the proccess of taking a picture?

AsminBudha: Without path option, it saves in Android/data/{app_package_name}/files so I think it won't much affect the speed if picture is save in another location.

docs/RNCamera.md
```diff
@@ -599,7 +599,7 @@ writeExif = {
 
 - `orientation` (string or number). Specifies the orientation that us used for taking the picture. Possible values: `"portrait"`, `"portraitUpsideDown"`, `"landscapeLeft"` or `"landscapeRight"`.
 
-- `path` (file path on disk). Specifies the path on disk to save picture to.
+- `path` (file path on disk). Specifies the path on disk to save picture to. Picture taking process might be slower as picture is saved other than cache directory.
```
sibelius: Is it?

Do we have a benchmark for this?

AsminBudha: I did check on by seeing difference in System.time between starting of taking photo process to completion of photo taking process. And, cache directory was about 100ms faster.

docs/RNCamera.md
AsminBudha: @MateusAndrade @sibelius 
Sorry for confusion but my above test was an fault might be due to setting debugger on. Now, I think it won't affect the speed as in both cases we have to write file into directory.

cristianoccazinsp:
Hey guys, I think this PR introduced a bug.

The following code was updated: https://github.com/AsminBudha/react-native-camera/blob/09823726f38e755220a14fb2e376a3a272ebbe05/android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java#L184 

However, the alternative case (in the else clause) wasn't updated: https://github.com/AsminBudha/react-native-camera/blob/09823726f38e755220a14fb2e376a3a272ebbe05/android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java#L238

I believe the same update needs to be applied to https://github.com/AsminBudha/react-native-camera/blob/09823726f38e755220a14fb2e376a3a272ebbe05/android/src/main/java/org/reactnative/camera/tasks/ResolveTakenPictureAsyncTask.java#L324

sibelius:
can @AsminBudha or @cristianoccazinsp send a PR to fix this?

cristianoccazinsp:
I would like @AsminBudha  to take a look first, since although there's a confirmed bug/inconsistency there, it still doesn't explain this other bug report: https://github.com/react-native-community/react-native-camera/issues/2778

I will try to test this now that I'm preparing a PR to fix https://github.com/react-native-community/react-native-camera/issues/2780

sibelius:
tks

cristianoccazinsp:
I'm working on a PR to fix this bug and also implement this for iOS.

PR #2785, opened 2020-04-10
Take pictures in PNG format on iOS

rfischer:
# Summary
Right now, it is only possible to take a picture in JPEG format with lossy compression. Therefore, I added a new property `imageType` to `TakenPictureOption`, which allows choosing between JPEG and PNG output format on iOS.

## Test Plan
Take pictures of the same motive with different `imageType` settings and compare the resulting filename, file size and, document type.

|      Testcase      | Result |
| ------- | :---------: |
| don't set `imageType`  |  *.jpg file with JPEG document type and applied lossy compression as in quality option specified    |
| set `imageType` to `'jpg'`  |  same as above   |
| set `imageType` to `'whatever'`  |  same as above   |
| set `imageType` to `'png'`  |   *.png file with PNG document type and lossless compression; larger file size      |

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ❌     |

## Checklist

- [X] I have tested this on a device 


ios/RN/RNCamera.m
```diff
@@ -1284,7 +1295,7 @@ - (void)startSession
 
         AVCaptureStillImageOutput *stillImageOutput = [[AVCaptureStillImageOutput alloc] init];
         if ([self.session canAddOutput:stillImageOutput]) {
-            stillImageOutput.outputSettings = @{AVVideoCodecKey : AVVideoCodecJPEG};
+            stillImageOutput.outputSettings = @{AVVideoCodecKey : AVVideoCodecJPEG, AVVideoQualityKey: @(1.0)};
```
cristianoccazinsp, line 1298: What's this one exactly? Is this a change to video recording quality?

cristianoccazinsp:
Have you tried what happens when you try to add extra exif/metadata and use PNG? Does PNG support exif metadata? Would it crash? Would it get ignored?

```
let exif = {
          "GPSLatitude": latitude,
          "GPSLongitude": longitude,
          "GPSAltitude": altitude
}
takePictureAsync({...options, writeExif: writeExif});

```


Lastly, what about Android's implementation? I guess Android always takes photos as JPEG, and then compresses it again if there was any manipulation. I guess we should update Camera1 to receive the RAW callback instead of JPEG (https://developer.android.com/reference/android/hardware/Camera#takePicture(android.hardware.Camera.ShutterCallback,%20android.hardware.Camera.PictureCallback,%20android.hardware.Camera.PictureCallback,%20android.hardware.Camera.PictureCallback))  . Would it be feasible?

ios/RN/RNCamera.m
rfischer: Setting this key ensures the highest quality for the original taken image instead of relying on the iOS default value (which I could not find). The quality wanted for the resulting image is applied with these lines of code
```obj-c
float quality = [options[@"quality"] floatValue];
[metadata setObject:@(quality) forKey:(__bridge NSString *)kCGImageDestinationLossyCompressionQuality];
...
CGImageDestinationAddImage(destination, takenImage.CGImage, writeExif ? ((__bridge CFDictionaryRef) metadata) : nil);
```

By looking at this code, I think that I found a bug: The wanted quality is only applied if `writeExif` is not specified or is set to `true`. If a developer uses `writeExif: false` with `quality: 1.0` the resulting image has not the same quality/file size as if `writeExif: true` was used. Can anyone confirm this?

rfischer:
@cristianoccazinsp thank you for your feedback. I tested it and PNG does support exif/metadata.

Screenshot 2020-04-15 at 10 44 44 Screenshot 2020-04-15 at 10 44 40 Screenshot 2020-04-15 at 10 44 38 
Screenshot 2020-04-15 at 10 44 33 Screenshot 2020-04-15 at 10 44 31 Screenshot 2020-04-15 at 10 44 28

I have no experience in android development, so I can't tell you the best approach there, but your suggestion sounds plausible.

ios/RN/RNCamera.m
cristianoccazinsp: You may be right about the exif bug.

About the above line change. That is interesting; by looking at `quality=1` photos size on an iphone 7 (3-4mb), I would say the quality was already 1. Is that something that can be tested easily? Now I wonder if the quality value should be applied to the above setting instead of the metadata.

cristianoccazinsp: I've just tested, and using ` AVVideoQualityKey: @(1.0)` did not change anything to the file size on my test device (which is good, not a breaking change I guess, but it's better to be verbose about that setting).

About the exif bug, do you think you could fix it by just adding a "fake" metadata object with just the compression flag in the case exif is not used? Seems like a simple change as opposed to changing the value of `stillImageOutput.outputSettings` dynamically.

rfischer: Thank you for testing :) . I made a new commit https://github.com/react-native-community/react-native-camera/pull/2785/commits/55feaabe1f1fee5e689c6a9e83330771945b86f0 wich ensures that the compression flag is always specified.

rfischer:
Hey guys. What is the status of this pull request? Do you need more info from me?

tewarig (approved):

MateusAndrade:
Hey @cristianoccazinsp , @rfischer  do you guys think we can go on with this PR?

cristianoccazinsp:
Looks ok to me.

PR #2792, opened 2020-04-14
test: add on-device tests

wolverineks:
# Summary

## Test Plan

```bash
cd examples/tests
yarn && yarn android & yarn ios
```
result: passed on android emulator/device, ios simulator
not yet tested on ios device

devices attached or simulators/emulators install

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS - device     |    ❌     |
| iOS - simulator    |    ✅     |
| Android - device |    ✅     |
| Android - emulator |    ✅     |

## Checklist

- [X] I have tested this on android device
- [X] I have tested this on android emulator
- [ ] I have tested this on a ios device
- [X] I have tested this on a ios simulator


wolverineks:
doesnt have any ci/cd integration
only has 2 simple tests
havent tested it on ios device

wolverineks:
ios simulators dont have cameras, so `<RNCamera />` doesnt render completely, so most tests fail with no camera available

wolverineks:
which tests are most valuable?
which tests do contributors run most frequently?
which tests take the longest to setup/run?

is there a list of all the manual tests?

sibelius:
can you provide some gifs of tests running?

wolverineks:
> can you provide some gifs of tests running?

rncamera-tests

wolverineks:
ive written a handful of other tests, but most dont work on ios simulator
so ive included only a few that work everywhere

sibelius:
I think we should include all of them, and figure it out how to enable device testing on CI

maybe a device farm

wolverineks:
Screen Shot 2020-04-14 at 14 47 47

wolverineks:
also, i could only get the front camera to work, for both photos and videos
the rear camera would fail a bunch of tests that the front camera passes
the camera would flash a bunch, the tests got really slow and then fail, for some unknown reason

sibelius:
can you add this failing tests?

I think we can run locally all of them for each pull requests to make sure it won't break

does it test both camera1 and camera2 apis?

wolverineks:
> can you add this failing tests?
> 
> I think we can run locally all of them for each pull requests to make sure it won't break
> 
> does it test both camera1 and camera2 apis?

if you're talking about `useCamera2Api`, no, there are no tests that use that
is there `useCamera2Api` for front and back cameras?

ill try and duplicate all the tests for front and back camera, as well as `useCamera2Api`

looking at all the different props, methods, and args, there is a lot opportunity for de-duplication, so any suggestions would be helpful

wolverineks:
are there plans to deprecate the legacy camera api on android?

wolverineks:
is `useCamera2API` for photos as well as videos?

sibelius:
I think both

@cristianocca has some thoughts on this

it looks like Camera1 is stable than Camera2 in some devices

we also have CameraX now

cristianocca:
Camera2 is quite unstable for now. Although it doesn't crash often, it is very slow to capture, and hasn't received a bunch of improvements Camera1 has (crash fixes, race condition fixes, ANRs fixes, and some others). For me, it being this slow makes it unusable on my project so I haven't had time to look at all of its issues.

Perhaps CameraX is now a better alternative than Camera2, and if it behaves more like Camera1 then it might be a better option to work on that rather than trying to fix all Camera2's issues.

sibelius:
do you wanna add a camera X implementation?

we could have a prop like `api={API.CAMERA1}`, so we can easily switch from camera1/camera2 and cameraX

cristianocca:
If cameraX provides to be good enough, I believe it should replace Camera2? Doesn't CameraX build on top of Camera2 with a simplified API? Such change would make Camera2 more reliable since there's more abstraction, but I don't know for sure.

sibelius:
we can't break users, so we would need to add cameraX first, and then deprecated camera1 and camera2

cristianocca:
If things are done right and well tested, CameraX should be able to replace Camera2 in place with no breaking changes (they are the same base API after all). I agree with Camera1, that one has a long way to get deprecated and probably will stay alive until React Native bumps the min SDK to 21. I don't really think Camera1 can be removed since even older phones with an up-to-date Android can only use Camera1 :(

PR #2808, opened 2020-04-27
**fix 'int android.media.CamcorderProfile.quality' on a null object r…

zhonghanwen:
…eference.

# Summary

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅    |

## Checklist

- [] I added the documentation in `README.md`
- [] I mentioned this change in `CHANGELOG.md`
- [] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/android/src/main/java/org/reactnative/camera/RNCameraView.java
+++ b/android/src/main/java/org/reactnative/camera/RNCameraView.java
@@ -274,13 +274,21 @@ public class RNCameraView extends CameraView implements LifecycleEventListener, 
       @Override
       public void run() {
         try {
+          int currentCameraId = -1;
+          if (getCameraId() != null && getCameraId().length() > 0){
+            currentCameraId = Integer.parseInt(getCameraId());
+          }else {
+            promise.reject("E_RECORDING_FAILED", "Starting video recording failed. CameraId might be error.");
+            return;
+          }
+
           String path = options.hasKey("path") ? options.getString("path") : RNFileUtils.getOutputFilePath(cacheDirectory, ".mp4");
           int maxDuration = options.hasKey("maxDuration") ? options.getInt("maxDuration") : -1;
           int maxFileSize = options.hasKey("maxFileSize") ? options.getInt("maxFileSize") : -1;
 
-          CamcorderProfile profile = CamcorderProfile.get(CamcorderProfile.QUALITY_HIGH);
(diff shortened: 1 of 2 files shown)


cristianoccazinsp:
What does this fix already? What was the bug? Did recording with a different camera have issues?

cristianoccazinsp:
Also, doesn't Camera2 return "non integer" camera ids? Does this work with Camera2?

android/src/main/java/org/reactnative/camera/RNCameraView.java
```diff
@@ -274,13 +274,21 @@ public void record(final ReadableMap options, final Promise promise, final File
       @Override
       public void run() {
         try {
+          int currentCameraId = -1;
+          if (getCameraId() != null && getCameraId().length() > 0){
+            currentCameraId = Integer.parseInt(getCameraId());
+          }else {
+            promise.reject("E_RECORDING_FAILED", "Starting video recording failed. CameraId might be error.");
```
cristianoccazinsp, line 281: Should it be "CameraId may be invalid."?

android/src/main/java/org/reactnative/camera/RNCameraView.java
```diff
@@ -274,13 +274,21 @@ public void record(final ReadableMap options, final Promise promise, final File
...
           String path = options.hasKey("path") ? options.getString("path") : RNFileUtils.getOutputFilePath(cacheDirectory, ".mp4");
           int maxDuration = options.hasKey("maxDuration") ? options.getInt("maxDuration") : -1;
           int maxFileSize = options.hasKey("maxFileSize") ? options.getInt("maxFileSize") : -1;
 
-          CamcorderProfile profile = CamcorderProfile.get(CamcorderProfile.QUALITY_HIGH);
+          CamcorderProfile profile = CamcorderProfile.get(currentCameraId, CamcorderProfile.QUALITY_HIGH);
           if (options.hasKey("quality")) {
-            profile = RNCameraViewHelper.getCamcorderProfile(options.getInt("quality"));
+            profile = RNCameraViewHelper.getCamcorderProfile(currentCameraId, options.getInt("quality"));
           }
```
cristianoccazinsp, line 292: Should you still check if `profile` is null?

PR #2815, opened 2020-04-30
Feature/add native zoom

SimonErm:
# Summary
Since none of the js pinch-to-zoom implementation
offer a satisfying solution, whether PanResponder or react-native-gesture-handler is used. I have implemented a native solution which can be activated with the property "useNativeZoom". 
But the Android touch event handling causes that child views catch the touch events and therefore the implementation does not work in combination with the touch-to-focus implementation from the advanced example.
If this PR is accepted, I would make another one, which extends the Camera-View with onTap and onDoubleTab events to provide a solution/workaround

For testing a real device is required
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅   |
| Android |    ✅    |

## Checklist

- [x] I added ~~the~~ documentation ~~in `README.md`~~

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -95,6 +95,15 @@ BOOL _sessionInterrupted = NO;
     }
     return self;
 }
+-(float) getMaxZoomFactor:(AVCaptureDevice*)device {
+    float maxZoom;
+    if(self.maxZoom > 1){
+        maxZoom = MIN(self.maxZoom, device.activeFormat.videoMaxZoomFactor);
+    }else{
+        maxZoom = device.activeFormat.videoMaxZoomFactor;
+    }
+    return maxZoom;
+}
 
 -(void) handlePinchToZoomRecognizer:(UIPinchGestureRecognizer*)pinchRecognizer {
     const CGFloat pinchVelocityDividerFactor = 5.0f;
@@ -603,13 +612,7 @@ BOOL _sessionInterrupted = NO;
         return;
     }
 
-    float maxZoom;
-    if(self.maxZoom > 1){
-        maxZoom = MIN(self.maxZoom, device.activeFormat.videoMaxZoomFactor);
-    }
-    else{
-        maxZoom = device.activeFormat.videoMaxZoomFactor;
-    }
+    float maxZoom = [self getMaxZoomFactor:device];
 
     device.videoZoomFactor = (maxZoom - 1) * self.zoom + 1;
 
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -114,10 +114,11 @@ BOOL _sessionInterrupted = NO;
             return;
         }
         NSError *error = nil;
+        float maxZoom = [self getMaxZoomFactor:device]
         if ([device lockForConfiguration:&error]) {
             CGFloat desiredZoomFactor = device.videoZoomFactor + atan2f(pinchRecognizer.velocity, pinchVelocityDividerFactor);
             // Check if desiredZoomFactor fits required range from 1.0 to activeFormat.videoMaxZoomFactor
-            device.videoZoomFactor = MAX(1.0, MIN(desiredZoomFactor, device.activeFormat.videoMaxZoomFactor));
+            device.videoZoomFactor = MAX(1.0, MIN(desiredZoomFactor, maxZoom));
             [device unlockForConfiguration];
         } else {
             NSLog(@"error: %@", error);
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -13,6 +13,7 @@
 @property (nonatomic, weak) RCTBridge *bridge;
 @property (nonatomic,strong) RNSensorOrientationChecker * sensorOrientationChecker;
 
+@property (nonatomic,strong) UIPinchGestureRecognizer *pinchGestureRecognizer;
 @property (nonatomic, strong) RCTPromiseResolveBlock videoRecordedResolve;
 @property (nonatomic, strong) RCTPromiseRejectBlock videoRecordedReject;
 @property (nonatomic, strong) id textDetector;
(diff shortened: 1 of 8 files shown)


SimonErm:
I am glad to hear :)

cristianoccazinsp:
This is awesome, looks promising!

One question, what happens if the `zoom` property is also used? Which one takes precedence/control over the camera's zoom?

Any recommendation to "migrate" from the JS implementation to the native implementation? Can't wait to test it, on Android, the JS implementation is definitely slow on cheap devices :)

SimonErm:
You can still use the zoom property to set the default zoom or for resetting the zoom for example. 
So if you set a new value for the zoom property which triggers a rerender, this value is taken as zoom. The pinch zoom uses the current zoom to compute the new one, so there won't be a problem.

To migrate just remove your javascript Zoomview and set the useNativeZoom property.
The only thing to keep in mind, is the Android-Event-Handling problem mentioned in the summary.

ios/RN/RNCamera.m
```diff
@@ -94,6 +95,36 @@ - (id)initWithBridge:(RCTBridge *)bridge
...
+-(void) handlePinchToZoomRecognizer:(UIPinchGestureRecognizer*)pinchRecognizer {
+    const CGFloat pinchVelocityDividerFactor = 5.0f;
+
+    if (pinchRecognizer.state == UIGestureRecognizerStateChanged) {
+        AVCaptureDevice *device = [self.videoCaptureDeviceInput device];
+        if(device == nil){
+            return;
+        }
+        NSError *error = nil;
+        float maxZoom = [self getMaxZoomFactor:device]
```
ThomasGoatITMedia, line 117: Hi! Great work but a semicolon is missing, it won't build at this moment.

sibelius: try the latest version

cristianoccazinsp:
@simistern hello. I'm trying this, but it looks like it will not work if the RNComponent renders a TouchableOpacity inside it (e.g.. to provide tap to focus). Any way to work around this limitation?

SimonErm:
@cristianoccazinsp yes that's an known issue which is already mentioned in the summary.
https://github.com/react-native-community/react-native-camera/pull/2827 will add a workaround.

I think a callback for this would be a good idea.

cristianoccazinsp:
Thanks, I misunderstood the warning :D.

Looking forward to testing the work around.

peterasorensen:
Similar to @cristianocca concerns, but if you use any kind of encapsulating component inside of your RNCamera the useNativeZoom property does not work. For instance, using a SafeAreaView or a View to handle taps blocks the pinch to zoom. 

Code example where useNativeZoom isn't working:
```js
        <View style={{flex: 1}}>
          <RNCamera
            style={{flex: 1}}
            type={this.state.cameraDirection}
            useNativeZoom={true}
            flashMode={RNCamera.Constants.FlashMode.off}
          >
            <SafeAreaView style={{flex: 1}}>
              // more content inside
            </SafeAreaView>
          </RNCamera>
        </View>
```

I tried messing with zIndex to push the whole SafeAreaView to the back and then pinching works but obviously not a great solution because the rest of my components disappear too ...

joebernard:
@peterasorensen Try adding `pointerEvents="box-none"` to a view inside `SafeAreaView`.

```javascript
      <View style={{flex: 1}}>
        <RNCamera
          style={{flex: 1}}
          type={this.state.cameraDirection}
          useNativeZoom={true}
          flashMode={RNCamera.Constants.FlashMode.off}>
          <SafeAreaView style={{flex: 1}}>
            <View pointerEvents="box-none">
              // more content inside
            </View>
          </SafeAreaView>
        </RNCamera>
      </View>
```

You may be able to add it directly to the `SafeAreaView` though it wasn't working when I tried it a few weeks ago.

peterasorensen:
@joebernard I got that to work directly in the SafeAreaView by changing it to `pointerEvents="none"` to make sure all touch events go to the ancestor RNCamera. This works for the SafeAreaView but what if I have a component that also handles taps across the entire screen? For example an encapsulating `TouchableWithoutFeedback` instead:

```js
      <View style={{flex: 1}}>
        <RNCamera
          style={{flex: 1}}
          type={this.state.cameraDirection}
          useNativeZoom={true}
          flashMode={RNCamera.Constants.FlashMode.off}>
          <TouchableWithoutFeedback onPress={this.handleTap}>
            // more content inside
          </TouchableWithoutFeedback>
        </RNCamera>
      </View>
```

Adding a pointerEvents='none' to the TouchableWithoutFeedback will prevent the this.handleTap from working. So I am forced to choose between my own functionality or the useNativeZoom functionality. Is there some way to merge these functionalities without overriding the other?

SimonErm:
@peterasorensen take a look at https://github.com/react-native-community/react-native-camera/pull/2827. I think that should solve your problem.

borgen12:
I have an issue when using this prop where it will zoom a little bit, freeze for a second, then the camera will flash really fast between green/red/normal colors. Code looks like this:
`<RNCamera
                        ref={(cam) => {
                            this.camera = cam;
                        }}
                        style={styles.preview}
                        captureAudio={false}
useNativeZoom={true}
                    >
                        <View style={styles.previewHeader}>
                            <Text style={styles.headerButton} onPress={() => this.props.closeCamera()}>Back</Text>
                            <View>
                                <Text>{roomName}</Text>
                                <Text>{itemDisplayed.displayName}</Text>
                            </View>
                        </View>
                    </RNCamera>`

borgen12:
I have an issue when using this prop where it will zoom a little bit, freeze for a second, then the camera will flash really fast between green/red/normal colors. Code looks like this:
```<RNCamera
                     <RNCamera
                        ref={(cam) => {
                            this.camera = cam;
                        }}
                        style={styles.preview}
                        captureAudio={false}
                        useNativeZoom={true}>
                        <View style={styles.previewHeader}>
                            <Text style={styles.headerButton} onPress={() => this.props.closeCamera()}>Back</Text>
                            <View style={{ alignItems: 'center' }}>
                                <Text style={{ color: styleConstants.colors.TEXT_SECONDARY, fontSize: styleConstants.fontSize.LARGE }} >{roomName}</Text>
                                <Text style={{ color: '#fff' }}>{itemDisplayed.displayName}</Text>
                            </View>
                            <View style={{ width: 30 }}></View>
                        </View>
                    </RNCamera>```

sibelius:
@borgen12 I think it is better to open another issue with your use case

PR #2827, opened 2020-05-07
Feature/add on tap events

SimonErm:
# Summary
This PR adresses the touch issue mentioned in #2815 by adding an onTouch Event which contains the touch origin and a boolean that marks doubletaps.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅   |
| Android |    ✅    |

## Checklist

- [x] I have tested this on a device
- [x] I added ~~the~~ documentation ~~in `README.md`~~

### DIFF
--- a/docs/RNCamera.md
+++ b/docs/RNCamera.md
@@ -431,15 +431,20 @@ Event will contain the following fields:
 
 Function to be called when native code stops recording video, but before all video processing takes place. This event will only fire after a successful video recording, and it will not fire if video recording fails (use the error returned from `recordAsync` instead).
 
-### `onTouch`
+### `onTap`
 
 Function to be called when a touch within the camera view is recognized.
+The function is also called on the first touch of double tap.
 Event will contain the following fields:
+- `x`
+- `y`
 
-- `touchOrigin`
-  - `x`
-  - `y`
-- `isDoubleTap` tells if touch is a double tap (First touch of double tap is also recognized with `isDoubleTap` set to `false`)
+### `onDoubleTap`
+
+Function to be called when a double touch within the camera view is recognized.
+Event will contain the following fields:
+- `x`
+- `y`
 
 ### Bar Code Related props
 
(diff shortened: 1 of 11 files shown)


docs/RNCamera.md
```diff
@@ -431,6 +431,16 @@ Event will contain the following fields:
...
 
+### `onTouch`
+
+Function to be called when a touch within the camera view is recognized.
+Event will contain the following fields:
+
+- `touchOrigin`
+  - `x`
+  - `y`
+- `isDoubleTap` tells if touch is a double tap (First touch of double tap is also recognized with `isDoubleTap` set to `false`)
```
renatobenks, line 442: I was wondering if it wouldn't be better to have the double-tap feature behavior as an own prop instead?

such as`onDoubleTouch`, what do you think?

or even keeping just a single prop to touch but having a quick timeout between the touch in native and the calling of the react prop, to check before if it's a double touch or not, to avoid a miscalling call to be called just a once and not twice.

renatobenks: unless that you think that it will be resulting in a lot of hard work to do more, because I don't think that's a big problem for now, so we could move forward with that and create an issue to improve it in the future.

SimonErm: > I was wondering if it wouldn't be better to have an own prop to double-tap instead? something like aonDoubleTouch. what do you think?

I could add an onDoubleTouch property easily if it's wanted. I think you're right that seems to be the clean way.

> or even keeping just a single prop to touch but having a quick timeout between the touch in native and the calling of the react prop, to check before if it's a double touch or not, to avoid a miscalling call to be called just a once and not twice.

I just added the isDoubleTap feature for convinience, since many camera apps use double tap to reset zoom or something else and the gestureHandler provides an onDoubleTap Callback out of the box.

cristianoccazinsp:
Thoughts on adding `onLongPress` and `delayLongPress` props? I can't still seem to use the native zoom feature while still allowing for tap to focus.  Additionally, I'm using long press detection for capturing on screen long press, which also seems impossible right now.

PR #2833, opened 2020-05-12
fix(android): Prevent deleting exifData after saving it to the response

valioz:
# Summary

exifData becomes empty after saving it to the response (response.putMap("exif", exifData)), therefore, it is necessary to copy the object.

## Test Plan

Take picture with some processing and exif=true.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [ ✅ ] I have tested this on a device and a simulator
- [ ❌ ] I added the documentation in `README.md`
- [ ❌ ] I mentioned this change in `CHANGELOG.md`
- [ ❌ ] I updated the typed files (TS and Flow)
- [ ❌ ] I added a sample use of the API in the example project (`example/App.js`)


MateusAndrade:
I think there is an error with this change. Actually I think it's because of the `copy`.

https://app.circleci.com/pipelines/github/react-native-community/react-native-camera/107/workflows/8be06d5f-ce50-4cc0-bcd7-343000a6addb/jobs/1428/parallel-runs/0/steps/0-103

valioz:
Sorry, i didn't check backward compatibility, copy is introduced in RN 0.61. It should be fine now.

MateusAndrade:
> Sorry, i didn't check backward compatibility, copy is introduced in RN 0.61. It should be fine now.

No worries, thanks for helping! 😄

cristianoccazinsp:
Was this a bug introduced with RN 0.61 or something? I wasn't aware that putMap clears the source object.
Does this mean that the exif data on the actual image was empty if it was also sent to the response? 
Can you check if the same happens to iOS? I believe it uses the same logic.

valioz:
No, it is not from version 0.61, but I am not familiar with the details.
In the first version of this PR #2263 (tested on RN 0.59), a copy of the data was also made.
I don't have access to iOS and I can't check how it works there.

marmik18:
when is the expected release for this change?

Issue #2850, opened 2020-06-02
Enable inverted colors on iOS


manishkumaracv:
# Feature-Request

**Describe the Feature**
The ability to scan inverted-colored Data Matrix barcodes was added to the Android part sometime back. It was not added to iOS because iOS did not support `GoogleVisionBarcodeDetection` at that time. Now that iOS does support it, can the ability to scan inverted barcodes be added to it? Thanks.

**Possible Implementations**

**Related Issues**
https://github.com/react-native-community/react-native-camera/pull/1896

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate
> Want this feature to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

manishkumar03:
I have forked the repo and I am trying to enable barcode scanning for color-inverted data matrix barcodes. I am an iOS developer but I am working on a React Native app right now. I have made some changes to `RNCamera.m` but I am just not able to have them picked up by my app. I can see that the module `react-native-camera` is being linked from my github repo in `yarn.lock`. I have googled a lot but could not find any satisfactory answer. Can you please help me?

PR #2851, opened 2020-06-03
feat: Add ability to scan inverted-color barcodes on iOS using Google Vision

manishkumar03:
# Summary
This PR adds the capability to scan inverted-color barcodes [on iOS](https://github.com/react-native-community/react-native-camera/issues/2850) using Google Vision ML library. This PR is similar to what was done for [Android](https://github.com/react-native-community/react-native-camera/pull/1896). 

Adds two additional modes - `inverted` and `alternate`, in addition to `normal` (default).

`inverted`: each image will be inverted before being passed to the barcode detector. In this case only `inverted` barcodes can be detected.

`alternate`: each second frame will be inverted before being passed to the barcode detector. This mode will detect both `normal` and `inverted` barcodes.

When no mode is provided the barcode scanner behaves like normal and only `normal` barcodes can be detected.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-native-camera",
   "description": "A Camera component for React Native. Also reads barcodes.",
-  "version": "3.28.7",
+  "version": "3.28.8",
   "author": "Lochlan Wansbrough <lochie@live.com> (http://lwansbrough.com)",
   "collective": {
     "type": "opencollective",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-native-camera",
   "description": "A Camera component for React Native. Also reads barcodes.",
-  "version": "3.28.6",
+  "version": "3.28.7",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "react-native-camera",
   "description": "A Camera component for React Native. Also reads barcodes.",
-  "version": "3.26.0",
+  "version": "3.27.0",
```
sibelius, line 4: no need to bump here, the bump is automated

manishkumar03: Have reverted it back to its original value.

manishkumar03:
> does alternate works well in android?

Unfortunately I don't have any means of testing that. But once these changes are merged, I'll ask a teammate to test it on Android.

manishkumar03:
Hi, Any ETA when this PR will get merged? If it's going to take a few days, then I'll use my forked version in the project. Thanks.

PR #2856, opened 2020-06-05
fix: (googlevision barcode mode)

renatobenks:
# Summary

In order to test and validate the feature recently introduced by #2851 to solve the #2850, allowing the camera to detect barcodes with their colors inverted for Android and iOS, I caught a bug on testing it, added by the API changes on exposing the constant around the barcode mode options.

By the way, besides this problem found, everything else it seems to be working fine.

## Test Plan

- You should run correctly now, the mlkit example project

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅    |

## Checklist

- [x] ~~I added the documentation in `README.md`~~
- [x] ~~I updated the typed files (TS and Flow)~~
- [x] ~~I added a sample use of the API in the example project (`example/App.js`)~~

### DIFF
--- a/examples/mlkit/App.js
+++ b/examples/mlkit/App.js
@@ -87,14 +87,14 @@ export default class CameraScreen extends React.Component {
     });
   }
 
-  takePicture = async function () {
+  takePicture = async function() {
     if (this.camera) {
       const data = await this.camera.takePictureAsync();
       console.warn('takePicture ', data);
     }
   };
 
-  takeVideo = async function () {
+  takeVideo = async function() {
     if (this.camera) {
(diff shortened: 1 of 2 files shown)


renatobenks:
yeap @sibelius, there's no `CameraManager.GoogleVisionBarcodeMode` to be exposed,  but instead, the barcode mode it's actually on the `CameraManager.GoogleVisionBarcodeDetection.BarcodeMode`, but in any case, it would be possible to access the barcode mode enum in the outside to pass it as a prop for rn-camera

Issue #2859, opened 2020-06-05
Add the white balance for advanced example now supported to use on iOS


renatobenks:
# "Feature-Request"

**Describe the Feature**

Recently, it was introduced by #2774, a feature to support the uses of a custom white-balance for iOS on using the camera. So, in order that we've decided don't add, for now, an example about how to use this feature for our advanced project example, as commented over here https://github.com/react-native-community/react-native-camera/pull/2774#issuecomment-639620759. Still, it would be nice to have an example of how should be used that.

**Possible Implementations**

Just try to read and have a good understanding the API introduced to use that, and what exactly is this feature, as described here https://github.com/react-native-community/react-native-camera/pull/2774#issue-397098124

**Related Issues**

#2774 

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate
> Want this feature to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

PR #2865, opened 2020-06-09
Enhanced advanced example project

rfischer:
Added the custom white balance feature (#2774) to the advanced example project (https://github.com/react-native-community/react-native-camera/issues/2859)

New controls will be shown, if the white balance is set to "CW" on iOS:
- button to toggle between the custom white balance settings (temp., tint, gain values)
- slider to adjust the appropriate value

Tested on iOS Device (iPhone SE, iOS 13).

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/examples/advanced/advanced/src/screens/Camera.js
+++ b/examples/advanced/advanced/src/screens/Camera.js
@@ -9,6 +9,7 @@ import {
   AppState,
   TouchableOpacity
 } from 'react-native';
+import Slider from '@react-native-community/slider';
 import _ from 'underscore';
 import { Container, Button, Text, Icon, Footer, FooterTab, Spinner, H2, connectStyle, Toast } from 'native-base';
 import { RNCamera } from 'react-native-camera';
@@ -50,8 +51,17 @@ const WB_OPTIONS_MAP = {
   2: "CL",
   3: "SH",
   4: "IN",
-  5: "FL"
+  5: "FL",
+  6: "CW"
 }
+
+const CUSTOM_WB_OPTIONS_MAP = {
+  temperature:      {label: "Temp.", min: 1000, max: 10000, steps: 500},
+  tint:             {label: "Tint", min: -20, max: 20, steps: 0.5},
+  redGainOffset:    {label: "Red", min: -1.0, max: 1.0, steps: 0.05},
+  greenGainOffset:  {label: "Green", min: -1.0, max: 1.0, steps: 0.05},
+  blueGainOffset:   {label: "Blue", min: -1.0, max: 1.0, steps: 0.05},
+};
 
 const getCameraType = (type) => {
 
@@ -95,6 +105,7 @@ const styles = StyleSheet.create({
 
   buttonsView: {
     flex: 1,
+    backgroundColor: 'black',
     width: '100%',
     flexDirection: 'row',
     alignItems: 'center',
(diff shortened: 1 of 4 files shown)


examples/advanced/advanced/src/screens/Camera.js
```diff
@@ -601,6 +645,35 @@ class Camera extends Component{
...
+            maximumTrackTintColor="#000000"
+            onValueChange={this.changeCustomWBOptionValue}
+          />
+          <Text style={[styles.customWBViewText, {minWidth: '15%'}]}>{currentValue.toFixed(1)}</Text>
+        </View>
+      );
+      whiteBalance = this.state.customWhiteBalance;
+    } else {
+      whiteBalance = WB_OPTIONS[wb];
+    }
```
renatobenks, line 675: I know that's an example of code, but I think it's pretty much harder than it should be, so what do you think about the following suggestion?

```suggestion
    let whiteBalance = wb >= WB_OPTIONS.length ? this.state.customWhiteBalance : WB_OPTIONS[wb];
    const { currentCustomWBOption } = this.state;
    let customWhiteBalanceValue = this.state.customWhiteBalance[currentCustomWBOption];
    let customWhiteBalanceOption = CUSTOM_WB_OPTIONS_MAP[currentCustomWBOption]
```

examples/advanced/advanced/src/screens/Camera.js
```diff
@@ -701,6 +774,7 @@ class Camera extends Component{
                       />
                     </View>
                   : null}
+                  {customWBView}
```
renatobenks, line 777: in order to continue the code suggested above...

```suggestion
                  {wb >= WB_OPTIONS.length && (
                    <View style={styles.customWBView}>
                      <Button style={styles.customWBViewButton} onPress={this.changeCustomWBOption}>
                        <Text style={styles.customWBViewText}>
                          {customWhiteBalanceOption.label}
                        </Text>
                      </Button>
                      <Slider
                        style={styles.customWBViewSlider}
                        value={customWhiteBalanceValue}
                        step={customWhiteBalanceOption.steps}
                        minimumValue={customWhiteBalanceOption.min}
                        maximumValue={customWhiteBalanceOption.max}
                        minimumTrackTintColor="#FFFFFF"
                        maximumTrackTintColor="#000000"
                        onValueChange={this.changeCustomWBOptionValue}
                      />
                      <Text style={[styles.customWBViewText, {minWidth: '15%'}]}>
                        {customWhiteBalanceValue.toFixed(1)}
                      </Text>
                    </View>
                  )}
```

rfischer:
> great work @rfischer, thank you for doing that! 🎉
> 
> I just let you some kinds of suggestions to improve the code, but the example of the feature itself seems to be very useful and pretty completed already. So, feel free to move forward with the suggestions or not, even because this code is just regarding example code, right?
> 
> Either way, let me know what you think about it, and before committing the code suggested, take a look into the code to make sure that there's not any typo that I've introduced in such suggestions.

Thank you for the feedback and suggestions you have made. The changes look plausible, and the code gets more readable. And yes, this PR affects only the example code.

examples/advanced/advanced/src/screens/Camera.js
rfischer: Yes, you are right. It's better to keep the UI code in one place and the logic in another.
I will also add this line of code `let isCustomWhiteBalance = wb >= WB_OPTIONS.length;` to have the custom white balance check only once.

PR #2866, opened 2020-06-09
Fix app crash in barcode detection without FirebaseMLVision

rfischer:
Update the mode of the barcode detector only if it is a real detector.
Otherwise, the app will crash due to `-[BarcodeDetectorManagerMlkit setMode:queue:]:  unrecognized selector sent to instance`.

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -2105,7 +2105,9 @@ BOOL _sessionInterrupted = NO;
 
 - (void)updateGoogleVisionBarcodeMode:(id)requestedMode
 {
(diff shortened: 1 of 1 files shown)


ios/RN/RNCamera.m
```diff
@@ -2105,7 +2105,9 @@ - (void)updateGoogleVisionBarcodeType:(id)requestedTypes
 
 - (void)updateGoogleVisionBarcodeMode:(id)requestedMode
 {
-    [self.barcodeDetector setMode:requestedMode queue:self.sessionQueue];
+    if ([self.barcodeDetector isRealDetector]) {
```
manishkumar03, line 2108: 👍

Issue #2872, opened 2020-06-13
Why is three a half second delay between barcode scans?


manishkumar03:
# Question

**To Do First**
- [ ] Take a look in the [README](https://github.com/react-native-community/react-native-camera/blob/master/README.md)
- [ ] Take a look in the [docs](https://github.com/react-native-community/react-native-camera/blob/master/docs/RNCamera.md)
- [ ] Take a look in the [QA](https://github.com/react-native-community/react-native-camera/blob/master/docs/QA.md)

**Ask your Question**
There is a check `timePassedSinceSubmittingForBarcode > 0.5` in the code which basically means that barcode detection will be attempted once every 0.5 seconds. This limits the camera performance and causes delays in scanning. I wanted to know if there is any historical reason for this. If not, I'd like to change it. The same check is applicable for text and face detection.

**Tags**

Performance

> Love react-native-camera? Please consider supporting our collective: 👉  https://opencollective.com/react-native-camera/donate
> Want this issue to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

sibelius:
Let’s reduce this, or make this configurable

fabriziobertoglio1987:
@sibelius this could happen. This could be solved in my barcode scanner refactoring in https://github.com/react-native-community/react-native-camera/pull/2766 

My logic would decode the Image after turning it in the correct orientation (I still need to add some logic to handle all camera orientations).

https://github.com/react-native-community/react-native-camera/blob/5abbacaac4448311775c5544ea5feb3c3b5b94f9/android/src/main/java/org/reactnative/camera/tasks/BarCodeScannerAsyncTask.java#L39-L63

while the previous logic would 

1) Try to decode the Barcode in the original camera orientation

https://github.com/react-native-community/react-native-camera/blob/7706ccbea7cb53d4573ff93e7308cccf85d77fb9/android/src/main/java/org/reactnative/camera/tasks/BarCodeScannerAsyncTask.java#L76-L87

2) If the Barcode could not be decoded (because the image was in the wrong orientation), catch `NotFoundException` and try again decoding it

https://github.com/react-native-community/react-native-camera/blob/7706ccbea7cb53d4573ff93e7308cccf85d77fb9/android/src/main/java/org/reactnative/camera/tasks/BarCodeScannerAsyncTask.java#L88-L100

3) Repeat 2)

[...]

PR #2884, opened 2020-06-30
[iOS] Remove artificial throttling of barcode scanning

manishkumar03:
# Summary
The barcode scanning using ML was being artificially throttled to scan barcodes once every 500ms. This is not only highly inefficient but not even required because the following callback is invoked at regular intervals by the system itself:

`- (void)captureOutput:(AVCaptureOutput *)captureOutput
    didOutputSampleBuffer:(CMSampleBufferRef)sampleBuffer
           fromConnection:(AVCaptureConnection *)connection`

With this change the barcode scanning using ML is now much faster and can actually make use of all the hardware power of the device.

Related: [Feature Request](https://github.com/react-native-community/react-native-camera/issues/2872)
## Test Plan
Tested by scanning a number of barcodes. The scanning was as accurate as before but much faster.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅   |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -37,10 +37,8 @@
 @property (nonatomic, copy) RCTDirectEventBlock onRecordingEnd;
 @property (nonatomic, assign) BOOL finishedReadingText;
 @property (nonatomic, assign) BOOL finishedDetectingFace;
-@property (nonatomic, assign) BOOL finishedDetectingBarcodes;
 @property (nonatomic, copy) NSDate *startText;
 @property (nonatomic, copy) NSDate *startFace;
-@property (nonatomic, copy) NSDate *startBarcode;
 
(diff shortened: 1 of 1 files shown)


ios/RN/RNCamera.m
```diff
@@ -92,7 +88,7 @@ - (id)initWithBridge:(RCTBridge *)bridge
         self.cameraId = nil;
         self.isFocusedOnPoint = NO;
         self.isExposedOnPoint = NO;
-        self.invertImageData = false;
+        self.invertImageData = true;
```
sibelius, line 91: why this change?

manishkumar03: If the scanning mode is set to `ALTERNATE`, this flag gets inverted. This change will make sure that the first time the image is scanned, this flag is set to false and thus we try to scan the image first as a normal image.

PR #2911, opened 2020-07-21
Added pause and resume recording

pdeep95:
# Summary
Added pause recording and resume recording functionality for Android (API Version 24 onwards). It uses MediaRecorder's pause and resume methods.

## Usage

To pause recording use : 
`this.camera.pauseRecording();`
To resume recording use : 
`this.camera.resumeRecording();`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [X] I have tested this on a device (Android API 29)

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -853,6 +853,16 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
     }
 
     @Override
+    void pauseRecording() {
+        pauseMediaRecorder();
+    }
+
+    @Override
+    void resumeRecording() {
+        resumeMediaRecorder();
+    }
+
+    @Override
     int getCameraOrientation() {
         return mCameraInfo.orientation;
     }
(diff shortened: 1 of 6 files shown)


android/src/main/java/com/google/android/cameraview/Camera1.java
```diff
@@ -1569,6 +1579,18 @@ private void stopMediaRecorder() {
         }
     }
 
+    private void pauseMediaRecorder() {
+        if (Build.VERSION.SDK_INT >= 24) {
+            mMediaRecorder.pause();
+        }
```
sibelius, line 1585: what happen in old versions?

android/src/main/java/com/google/android/cameraview/Camera1.java
```diff
@@ -1569,6 +1579,18 @@ private void stopMediaRecorder() {
...
+    }
+
+    private void resumeMediaRecorder() {
+        if (Build.VERSION.SDK_INT >= 24) {
+            mMediaRecorder.resume();
```
sibelius, line 1590: same

android/src/main/java/com/google/android/cameraview/Camera1.java
pdeep95: Haven't tried in older version. But as per documentation pause and resume is available from API version 24 ( https://developer.android.com/reference/android/media/MediaRecorder#pause() ). 
It may throw an exception.

yg-yash: pause and resume is still not working in API 29 above

adrienfloor:
I apologise for what might be a stupid question but why not implementing it on iOS as well ? Is it more complicated ?
I found this in the documentation : https://developer.apple.com/documentation/avfoundation/avcapturefileoutput/1389849-resumerecording , so I assumed that it would be possible.
I'd be glad to work on it myself be to be honest I checked out the code and I don't even know where to start, I'm sorry, I've quite never worked with native code. I would need some help for sure.

Thanks a lot for the great work on this lib 🚀

Issue #2950, opened 2020-08-18
repeat scan qrcode, the app gets more and more stuck, and finally stuck


liufeiteng:
# Bug Report
When i repeat scan qrcode more than 10 times, my app gets more and more stuck, and completely stuck finally.

**Versions**
- Android: 9
- react-native-camera: 3.23.1
- react-native: 0.62.2
- react: 16.11.0
- react-navigation: 4.2.2

Issue #2904, opened 2020-07-18
3.30.0 leaks memory when scanning barcode


bozou:
# Bug Report

**To Do First**

- [No ] Did you try latest release?
- [Yes, but could neither run the sample app mlkit on device nor in simulator ] Did you try master?
- [Yes ] Did you look for existing matching issues?

**Platforms**

iOS

**Versions**

- Android: didn't try
- iOS:13.5.1
- react-native-camera:3.30.0
- react-native:0.62.2
- react:16.11.0
  
- react-navigation:5.5.0

**Description/Current Behaviour**
          &lt;RNCamera
            style={styles.preview}
            type={RNCamera.Constants.Type.back}
            onGoogleVisionBarcodesDetected={onGoogleBarCodeRead}
            googleVisionBarcodeMode={RNCamera.Constants.GoogleVisionBarcodeDetection.BarcodeMode.ALTERNATE}
            zoom={0.007}
            captureAudio={false}
            >
Leave the barcode scanner camera on for several minutes, the app would crash (grow over 1GB).  I turned on react native Perf Monitor and can see memory increase quickly.

Reverted back to 3.27.1.  Memory no longer leaks.

**Expected Behaviour**
no memory leak.

**Steps to Reproduce**
see description.

**Additionals**

anhtuank7c:
I have the same issue.
My iPhone will run out of memory after 3-5 minutes then crash the app

mav10:
+1

scottmickey:
+1

marvintrajano:
Running into this issue as well, did anyone find a solution to this?

PR #2955, opened 2020-08-22
[iOS] Optimize invertColors / fix memory leak

stanleykerr:
# Summary

Releases the CGImage that was created in order to populate the UIImage's CGImage property. This probably relates to #2904 and possibly #2950.

## Test Plan

Tested on iPhone X (14.0b5), iPhone 11 Simulator (13.5, 14.0), iPhone 11 Pro Simulator (13.5, 14.0)

I used Xcode's Leak Instruments to narrow down the reason for the increasing memory usage over time and found that the CGImage that was being created wasn't being released in a timely manner, as the images were being processed.

Device (using example code works)

Compare memory usage before/after implementing these changes via a debugger/profiler

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/RN/RNImageUtils.m
+++ b/ios/RN/RNImageUtils.m
@@ -126,10 +126,13 @@
 
     // A UIImage initialized directly from CIImage has its CGImage property set to NULL. So it has
     // to be converted to a CGImage first.
-    CIContext *context = [CIContext context];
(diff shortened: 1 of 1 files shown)


manishkumar03:
Thank you for fixing this bug (I am the original author of that block of code).

PR #2964, opened 2020-09-04
Relase camera on Android when pausing preview

santerioksanen:
# Summary
I am working on an application that requires both camera and nfc-access on the same screen, not simultaneously though. Certain Android phones however disable NFC-polling when camera is open due to NFC interfering with the camera. Background on this issue: https://stackoverflow.com/questions/34607168/nfc-unavailable-when-camera-is-open

This PR releases the camera so that NFC is free for use while the camera preview is paused.

## Test Plan
Tested on simulator and Huawei P20 Pro.
While camera preview is on, NFC is unavailable. Calling pausePreview() releases camera, making NFC available.
Calling continuePreview() works as expected.

Android phone with NFC

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [x ] I have tested this on a device and a simulator

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -352,6 +352,9 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
             @Override
             public void run() {
                 synchronized(this){
+                    if(mCamera == null){
+                        start();
+                    }
                     mShowingPreview = true;
                     startCameraPreview();
                 }
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -367,6 +367,7 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
 
             if(mCamera != null){
                 mCamera.stopPreview();
+                releaseCamera();
             }
         }
     }
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -352,6 +352,9 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
             @Override
             public void run() {
                 synchronized(this){
+                    //if(mCamera == null){
+                    //    start();
+                    //}
                     mShowingPreview = true;
                     startCameraPreview();
                 }
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -352,9 +352,6 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
             @Override
             public void run() {
                 synchronized(this){
-                    //if(mCamera == null){
-                    //    start();
-                    //}
                     mShowingPreview = true;
                     startCameraPreview();
                 }
--- a/android/src/main/java/org/reactnative/camera/CameraModule.java
+++ b/android/src/main/java/org/reactnative/camera/CameraModule.java
@@ -245,12 +245,7 @@ public class CameraModule extends ReactContextBaseJavaModule {
 
                 try {
                     cameraView = (RNCameraView) nativeViewHierarchyManager.resolveView(viewTag);
-                    if (cameraView.isCameraOpened()) {
-                        cameraView.resumePreview();
-                    } else {
-                        cameraView.start();
-                        cameraView.resumePreview();
-                    }
+                    cameraView.resumePreview();
                 } catch (Exception e) {
                     e.printStackTrace();
                 }
--- a/android/src/main/java/org/reactnative/camera/CameraModule.java
+++ b/android/src/main/java/org/reactnative/camera/CameraModule.java
@@ -247,6 +247,9 @@ public class CameraModule extends ReactContextBaseJavaModule {
                     cameraView = (RNCameraView) nativeViewHierarchyManager.resolveView(viewTag);
                     if (cameraView.isCameraOpened()) {
                         cameraView.resumePreview();
+                    } else {
+                        cameraView.start();
+                        cameraView.resumePreview();
                     }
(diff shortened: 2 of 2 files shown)


android/src/main/java/com/google/android/cameraview/Camera1.java
```diff
@@ -352,6 +352,9 @@ public void resumePreview() {
             @Override
             public void run() {
                 synchronized(this){
+                    //if(mCamera == null){
+                    //    start();
+                    //}
```
MateusAndrade, line 357: This can be removed?

MateusAndrade (changes requested):

android/src/main/java/com/google/android/cameraview/Camera1.java
santerioksanen: Sure, sorry, forgot it there after testing. Pushed a updated removing the comments.

sibelius (approved):

santerioksanen:
I found a bug caused by my first implementation: Concurrent calls to pausePreview() and resumePreview() caused the app to crash. However, moving the starting logic to Camera1.java fixed this behavior.

MateusAndrade:
> I found a bug caused by my first implementation: Concurrent calls to pausePreview() and resumePreview() caused the app to crash. However, moving the starting logic to Camera1.java fixed this behavior.

It's working without now?

santerioksanen:
> > I found a bug caused by my first implementation: Concurrent calls to pausePreview() and resumePreview() caused the app to crash. However, moving the starting logic to Camera1.java fixed this behavior.
> 
> It's working without any problem now?

Yeah, have not found any new bugs in my implementation, and I am planning to utilize this in production rather soon.
However, the roadmap of this project intends to replace Camera1 API with Camera2 API. I assume these are corresponding to the .java files on Android, so should probably implement these changes to Camera2.java as well?

Furthemore, I am not acknowledged of the expected behavior when calling pausePreview under special circumstances (f.ex zoom applied when pausing preview, video recording on, etc), so I haven't considered these while testing.

br. Santeri

sibelius:
You also need to implement this on Camera 2

sibelius:
You also need to implement this on Camera 2

santerioksanen:
> You also need to implement this on Camera 2

The changes between the APIs seem to be rather big, but I'll take a look at it. Do you have any help for testing using the Camera2 API?

sibelius:
We need more help for us

@cristianoccazinsp should I test these changes as they can break other use cases

cristianoccazinsp:
Won't this make the camera start/stop process much slower, or does it only impact `startPreview` / `stopPreview` calls, which are not normally used unless `pauseAfterCapture` is used? Rather than that, the changes seem harmless, but it would still be good to test it in a few more devices.

For the use case this is trying to fix, wouldn't it be much easier to just unmount the camera when you need NFC?

santerioksanen:
> Won't this make the camera start/stop process much slower, or does it only impact `startPreview` / `stopPreview` calls, which are not normally used unless `pauseAfterCapture` is used? Rather than that, the changes seem harmless, but it would still be good to test it in a few more devices.
> 
> For the use case this is trying to fix, wouldn't it be much easier to just unmount the camera when you need NFC?

It only impacts startPreview/stopPreview calls, so should not affect start/stop process.

I'm using a NativeBase Tabbar in my application, with one tab for the camera, the other for NFC. As changing between tabs did not make NFC available I added a if/else condition for mounting the camera component only if the camera tab is selected. This way the camera view did not render, and the UI became sluggish when changing back to the camera tab - without making NFC available. Might be issues with the cache etc of the application, but considering the sluggishness in the UI caused by mounting/unmounting the cameraview, I thought a better approach would be to release the camera when required.

cristianoccazinsp:
> > Won't this make the camera start/stop process much slower, or does it only impact `startPreview` / `stopPreview` calls, which are not normally used unless `pauseAfterCapture` is used? Rather than that, the changes seem harmless, but it would still be good to test it in a few more devices.
> > For the use case this is trying to fix, wouldn't it be much easier to just unmount the camera when you need NFC?
> 
> It only impacts startPreview/stopPreview calls, so should not affect start/stop process.
> 
> I'm using a NativeBase Tabbar in my application, with one tab for the camera, the other for NFC. As changing between tabs did not make NFC available I added a if/else condition for mounting the camera component only if the camera tab is selected. This way the camera view did not render, and the UI became sluggish when changing back to the camera tab - without making NFC available. Might be issues with the cache etc of the application, but considering the sluggishness in the UI caused by mounting/unmounting the cameraview, I thought a better approach would be to release the camera when required.


You may need to wrap the camera mounting / unmounting process in a `runAfterInteractions` or even `setTimeout` call to prevent it from breaking your animations / tab changing.  If I recall right, the camera now also starts/destroys in a non UI thread, so it starting it up should not lag the UI. Did you try with the most up to date version?

santerioksanen:
> > > Won't this make the camera start/stop process much slower, or does it only impact `startPreview` / `stopPreview` calls, which are not normally used unless `pauseAfterCapture` is used? Rather than that, the changes seem harmless, but it would still be good to test it in a few more devices.
> > > For the use case this is trying to fix, wouldn't it be much easier to just unmount the camera when you need NFC?
> > 
> > 
> > It only impacts startPreview/stopPreview calls, so should not affect start/stop process.
> > I'm using a NativeBase Tabbar in my application, with one tab for the camera, the other for NFC. As changing between tabs did not make NFC available I added a if/else condition for mounting the camera component only if the camera tab is selected. This way the camera view did not render, and the UI became sluggish when changing back to the camera tab - without making NFC available. Might be issues with the cache etc of the application, but considering the sluggishness in the UI caused by mounting/unmounting the cameraview, I thought a better approach would be to release the camera when required.
> 
> You may need to wrap the camera mounting / unmounting process in a `runAfterInteractions` or even `setTimeout` call to prevent it from breaking your animations / tab changing. If I recall right, the camera now also starts/destroys in a non UI thread, so it starting it up should not lag the UI. Did you try with the most up to date version?

Thanks for the details. I tried with version 3.31, did not try to optimize the UI at all, as the NFC was still blocked from reading.

PR #3065, opened 2020-12-28
[Android] add playSoundOnRecord

### LINKED ISSUE #3022: Android has no sound when recording video
**Version**

- "react-native": "0.61.5",
- "react-native-camera": "^3.40.0",

**Description**

Android records video without sound, which is normal on iOS devices

r0b0t3d:
Hi, 
Currently, we only have `playSoundOnCapture`. So I added `playSoundOnRecord` to make sound when start/stop record

Fix #3022


MateusAndrade (approved):
Thanks!

MateusAndrade (changes requested):
Nice! Btw I forgot to point the missing docs. 😢 

Would you mind adding this @r0b0t3d ? Please, let me know if you need any help with that.

r0b0t3d:
Hi @MateusAndrade , I did add doc in `docs/RNCamera.md`. Could you let me know what else that I need to add?

MateusAndrade (approved):

MateusAndrade:
> Hi @MateusAndrade , I did add doc in `docs/RNCamera.md`. Could you let me know what else that I need to add?

Nice! My bad, I didn't see that. Thanks!!! 🚀

cristianoccazinsp:
Hi guys, @r0b0t3d .  This change is good; however, it has a bug on some devices.  Testing on a Pixel 5 (and Pixel 2) with `playSoundOnRecord={true}`, triggering a record call will result in 2 sounds being played.  One from the default camera, and one from the `sound.play(MediaActionSound.START_VIDEO_RECORDING);` introduced here.  Therefore, this PR is actually incomplete: you also need to add some code to mute the default camera record sound (which I don't know how to do right now).  Please consider taking a second look at this change.

MateusAndrade:
> Hi guys, @r0b0t3d . This change is good; however, it has a bug on some devices. Testing on a Pixel 5 (and Pixel 2) with `playSoundOnRecord={true}`, triggering a record call will result in 2 sounds being played. One from the default camera, and one from the `sound.play(MediaActionSound.START_VIDEO_RECORDING);` introduced here. Therefore, this PR is actually incomplete: you also need to add some code to mute the default camera record sound (which I don't know how to do right now). Please consider taking a second look at this change.

Do you think we can add an issue for that or reverting this change?

cristianoccazinsp:
I guess adding an issue, or just a "caveat" into the docs should be enough.  When I looked into it, it seems rather difficult to disable Android's default sound for video, so perhaps it's not even possible to fix this for all devices.

Additionally, the sound maybe can be improved by preloading the record (and also the photo capture) media object so there's no delay when actually playing it, and then it would play at the same time as the default one.  These are some options, but I don't really know if they would work.

MateusAndrade:
> I guess adding an issue, or just a "caveat" into the docs should be enough. When I looked into it, it seems rather difficult to disable Android's default sound for video, so perhaps it's not even possible to fix this for all devices.
> 
> Additionally, the sound maybe can be improved by preloading the record (and also the photo capture) media object so there's no delay when actually playing it, and then it would play at the same time as the default one. These are some options, but I don't really know if they would work.

I think we can create an issue regarding that and adding this example to the docs. Btw, I'm not familiar with this API but I would be glad to help. =D

PR #3087, opened 2021-01-18
- Android method to check for corrupted videos

cristianoccazinsp:
**Summary**
- Implement android checkIfVideoIsValid to handle corrupted recordings.
- Add missing try/catch
- Add missing types

**Motivation**
We have been for a long time seeing some devices being unable to record videos, or recording corrupted videos from time to time.  This PR adds a module-level method that can be used to check whether a video is valid or corrupted (Android only).  For iOS, the method will always return true (valid).

This is still a draft/test update. We will be testing this PR on various devices for the following weeks in order to confirm there are no "false positives".

Bonus: added a missing type method.

P.S.: What happened to the PR template?

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -14,7 +14,7 @@ buildscript {
 
     dependencies {
       //noinspection GradleDependency
-      classpath("com.android.tools.build:gradle:3.6.3")
+      classpath("com.android.tools.build:gradle:4.1.0")
     }
   }
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -22,13 +22,13 @@ buildscript {
 apply plugin: 'com.android.library'
 
 android {
-  compileSdkVersion safeExtGet('compileSdkVersion', 28)
-  buildToolsVersion safeExtGet('buildToolsVersion', '28.0.3')
+  compileSdkVersion safeExtGet('compileSdkVersion', 29)
+  buildToolsVersion safeExtGet('buildToolsVersion', '29.0.2')
 
   defaultConfig {
     multiDexEnabled true
     minSdkVersion safeExtGet('minSdkVersion', 16)
-    targetSdkVersion safeExtGet('targetSdkVersion', 28)
+    targetSdkVersion safeExtGet('targetSdkVersion', 29)
   }
 
   flavorDimensions "react-native-camera"
--- a/android/gradle/wrapper/gradle-wrapper.properties
+++ b/android/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -943,8 +943,8 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
             }
             mDeviceOrientation = deviceOrientation;
             if (isCameraOpened() && mOrientation == Constants.ORIENTATION_AUTO && !mIsRecording.get() && !isPictureCaptureInProgress.get()) {
-                mCameraParameters.setRotation(calcCameraRotation(deviceOrientation));
                 try{
+                    mCameraParameters.setRotation(calcCameraRotation(deviceOrientation));
                     mCamera.setParameters(mCameraParameters);
                 }
                 catch(RuntimeException e ) {
@@ -1081,6 +1081,16 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
             mCallback.onCameraOpened();
             return true;
         } catch (RuntimeException e) {
+            // if camera failed to fully open
+            // try to release it before returning an error
+            // in order to avoid erratic behaviour
+            // Both getParameters and open may return null
+            try{
+                mCamera.release();
+                mCamera = null;
+            }
+            catch(RuntimeException e2){}
+
             return false;
         }
     }
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -792,13 +792,25 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
                         synchronized(Camera1.this){
                             if(mCamera != null){
                                 if (options.hasKey("pauseAfterCapture") && !options.getBoolean("pauseAfterCapture")) {
-                                    mCamera.startPreview();
-                                    mIsPreviewActive = true;
-                                    if (mIsScanning) {
-                                        mCamera.setPreviewCallback(Camera1.this);
+                                    try{
+                                        mCamera.startPreview();
+                                        mIsPreviewActive = true;
+                                        if (mIsScanning) {
+                                            mCamera.setPreviewCallback(Camera1.this);
+                                        }
+                                    }
+                                    catch(Exception e){
+                                        mIsPreviewActive = false;
+                                        mCamera.setPreviewCallback(null);
+                                        Log.e("CAMERA_1::", "camera startPreview failed", e);
                                     }
                                 } else {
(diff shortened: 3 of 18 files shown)


MarcoScabbiolo:
Thanks for the PR, do you have any metrics for the failing devices/conditions?

cristianoccazinsp:
No actual metrics values, but we have seen in various LG and huawei
devices. I don't know yet the exact conditions, but it's probably related
to those device having wrong video presets when different qualities are
selected.

MarcoScabbiolo:
If you have access to the corrupted videos try (ffprobing)[https://ffmpeg.org/ffprobe.html] them to see if any pattern arises.

cristianoccazinsp:
Since we mostly only get errors, we only have 1 actual video so far from a user that shared it and was constantly getting corrupted videos.  I will test ffmpeg to see the output and report back.

cristianoccazinsp:
Corrupted videos report the following:
```
[mov,mp4,m4a,3gp,3g2,mj2 @ 0x7fc55201b000] moov atom not found
corrupted_orig.mp4: Invalid data found when processing input
```

@MarcoScabbiolo any ideas what that means? I hace also tested this PR's code with the same video, and it does in fact return that there's no video in the file, so I'm guessing the logic is good.

I still want to test on a few other devices just to be sure, and let that sit in the store for a while.

MarcoScabbiolo:
The moov atom is a map of the video, like an index. Most likely means the data stream (recording/encoding) was stopped early. To make the atom you need to have all the data so it's the last part of the stream for a video that is being recorded and/or encoded.

cristianoccazinsp:
I wonder if may be caused by the Java code stopping the recorders when `maxFileSize` or `maxDuration` are set.  I do not have a device that suffers from this problem, but perhaps some inconsistencies in their API implementations is causing this. For instance, maybe the stop call is happening in the wrong thread, or perhaps it is not waiting enough time for the stream to end be written to disk. Maybe the user installed the app in an SD card that has faulty sectors... so many possibilities.

MarcoScabbiolo:
I've had issues like this one for implementations that are made to save to disk and "optimize" the final output by appending the moov at the beginning of the file so when you start reading/downloading it you can seek to a specific time. This makes it impossible to treat the video as a stream while it's being created. But, as you said, there are many possibilities and its imposible to know the specific issue without reproducing and debugging it.

cristianoccazinsp:
Update: Looks like `MediaMetadataRetriever` is not that reliable. Xiaomi devices reporting "no video" even on valid videos... So still gonna work on this a bit more.

cristianoccazinsp:
Update: added some notes about the expected path format for MediaMetadataRetriever (to not use URIs) since not all devices support the file:// format returned by the record call.

Also added some extra checks to fallback if the main video detection check doesn't work.  Should be more reliable now, but still testing.

MarcoScabbiolo:
> Update: added some notes about the expected path format for MediaMetadataRetriever (to not use URIs) since not all devices support the file:// format returned by the record call.

You can handle the uri validation/transformation into a regular fs path in the native method.

cristianoccazinsp:
Since the recordAsync call takes a `path` argument without (file://), I think it makes sense for this method to do the same.  But either way, it can be handled in the JS side as you said.

cristianoccazinsp:
21 days afterwards, and the code updates seem to be working fine in production.  No issues observed so far, but still testing.

MarcoScabbiolo (approved):

MarcoScabbiolo:
Changes look OK, thanks!

MateusAndrade:
@cristianoccazinsp do you know what could break the build for android? I wasn't able to find out, looking at the logs at circleci

cristianoccazinsp:
I couldn't find it neither.  Perhaps the tester uses a lower SDK version?

MateusAndrade:
> I couldn't find it neither. Perhaps the tester uses a lower SDK version?

Looking at the logs, it's related to a `close` method: 

```
/root/project/android/src/main/java/org/reactnative/camera/CameraModule.java:536: error: cannot find symbol
            retriever.close();
```

Maybe, we would need to increase the min version of the SDK for the project? Or at least for the example app?

cristianoccazinsp:
That explains it. I've been using API 29 for a long time since that's what RN >= 0.62 is using.

```
// this many fail or may not be available in API < 29
          try{
            retriever.close();
          }
          catch(Throwable e){}
```

MarcoScabbiolo:
KitKat (API level 19) still holds a considerable 4% user base
https://www.xda-developers.com/android-version-distribution-statistics-android-studio/

cristianoccazinsp:
Note that API 29 is only the "compile" sdk version, not the min SDK.  The min SDK should remain the same (hence the try/catch).  But for the compile error to go away (and the close method even being available), the target SDK has to be 29.  I think it would make sense to bump it from 28 to 29 since that's what RN is currently using.

MateusAndrade:
> Note that API 29 is only the "compile" sdk version, not the min SDK. The min SDK should remain the same (hence the try/catch). But for the compile error to go away (and the close method even being available), the target SDK has to be 29. I think it would make sense to bump it from 28 to 29 since that's what RN is currently using.

So, we need to bump the compile version for the example or the project? Sorry, I'm confused 😅

cristianoccazinsp:
I think so.  W/e code the test suite is using, the gradle build file should be bumped from 28 to 29. Perhaps the example projects should be bumped too if they are using RN >= 0.62

MateusAndrade:
> I think so. W/e code the test suite is using, the gradle build file should be bumped from 28 to 29. Perhaps the example projects should be bumped too if they are using RN >= 0.62



> I think so. W/e code the test suite is using, the gradle build file should be bumped from 28 to 29. Perhaps the example projects should be bumped too if they are using RN >= 0.62

Do you think you can bump these at your PR? Otherwise, we would end our master branch broken. (I didn't check if I gout the rights to push to your branch if so, I can try fixing that)

cristianoccazinsp:
Just pushed a commit to bump the compile version. 

Note that the basic example already had SDK 29, but the other projects didn't. So this may mean a breaking change for those using SDK 28 instead of 29.

cristianoccazinsp:
It's still not building! Hmm..

cristianoccazinsp:
@MateusAndrade  with the bumped compile SDK version, the previous `close` error shouldn't happen anymore. Can you see what's failing in circleci now? I can't seem to get the logs without signing up and giving access to every private and public repo :D

MateusAndrade:
> @MateusAndrade with the bumped compile SDK version, the previous `close` error shouldn't happen anymore. Can you see what's failing in circleci now? I can't seem to get the logs without signing up and giving access to every private and public repo :D

Sure! Strange, it's still throwing an error on the `close` method. Did you try running this `gradlew check` locally?

```
> Task :processGeneralReleaseManifest
> Task :generateGeneralDebugRFile

> Task :compileGeneralDebugJavaWithJavac
/root/project/android/src/main/java/org/reactnative/camera/CameraModule.java:536: error: cannot find symbol
            retriever.close();
                     ^
  symbol:   method close()
  location: variable retriever of type MediaMetadataRetriever
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: /root/project/android/src/main/java/com/google/android/cameraview/Camera2.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error

> Task :compileGeneralDebugJavaWithJavac FAILED
> Task :generateGeneralReleaseRFile

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileGeneralDebugJavaWithJavac'.
> Compilation failed; see the compiler error output for details.
```

cristianoccazinsp:
So the tests don't build against the `test` folder, but rather against the very own build.gradle file from the library itself.  When run this way, it uses the default SDK versions (28). I'm guessing that's the problem.  Should be solved now.

Note that this probably needs to be marked as a breaking change since people will need to upgrade from compile SDK 28 to 29 (which should already be the case when using RN >= 0.62)

cristianoccazinsp:
Well, still not working! But `gradlew check` worked fine locally.

MateusAndrade:
> Well, still not working! But `gradlew check` worked fine locally.
> 
> Edit: There we go, looks like there were also NDK/Gradle issues. Upgrading gradle fixed it.

NOW IT'S WORKING! 🚀 🚀 🚀 🚀 🚀 🚀 

Thanks a lot @cristianoccazinsp! 😄

adam-bizgo:
Im still getting  error: cannot find symbol ; retriever.close();
 (minSdkVersion = 20;  compileSdkVersion = 28)

cristianoccazinsp:
Need to bump compile sdk from 28 to 29 for it to compile.

PR #3102, opened 2021-02-09
update exif dependency

cristianoccazinsp:
Upgrading Exifinterface from 1.1.0 to 1.3.2. No breaking changes expected, but just some overflow and memleaks fixes according to the logs (https://developer.android.com/jetpack/androidx/releases/exifinterface).

This is an attempt to fix some android devices corrupting images due to exif bugs, and also to bring it up to the same version as react-native-image-resizer, which works very well when combined with this library.

Related (https://github.com/bamlab/react-native-image-resizer/pull/273)


friederbluemle (approved):
Thanks @cristianoccazinsp :+1:

PR #3103, opened 2021-02-09
Remove exif thumbnail generation that may affect some devices.

cristianoccazinsp:
Following the exif library upgrade, and after further debugging corrupted images, I've found out that some devices (especially samsung and high res camera ones) will include a thumbnail within the EXIF metadata.

This is somehow an erratic behaviour, as only some devices do this. However, if `writeExif` is used, it may cause either an out of memory error, or end up returning a massive base64 string to the JS side.

This update tries to standardize the behaviour, by not returning any thumbnail within the exif data.  Note that this may be a breaking change if anyone was relying on this, which is unlikely since the behaviour varied from device to device.


MateusAndrade (approved):

MateusAndrade:
Thanks @cristianoccazinsp!

PR #3106, opened 2021-02-10
Update react-navigation.md

ThaJay:
Update docs to reflect react-navigation 5x with hooks

### DIFF
--- a/docs/react-navigation.md
+++ b/docs/react-navigation.md
@@ -4,25 +4,22 @@ title: React Navigation
 sidebar_label: React Navigation
 ---
 
-React-navigation does not unmount components when switching between tabs. So when you leave and return back to the screen with the camera component it will just be black view. So a good solution is to use `withNavigationFocus`, which is a higher order component, and wrap it around your component. Then, you can have access to `isFocused` from `props`.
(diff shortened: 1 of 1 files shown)


MarcoScabbiolo:
While updating the example to use react-navigation 5 is ok, the example is wrong because you cannot use a hook inside a class.

docs/react-navigation.md
```diff
@@ -8,10 +8,10 @@ React-navigation does not unmount components when switching between tabs. So whe
 
 ```jsx
 
-import { withNavigationFocus } from 'react-navigation' 
+import { useIsFocused } from '@react-navigation/core';
 
  render() {
-    const { isFocused } = this.props
+    const isFocused = useIsFocused();
```
MarcoScabbiolo, line 14: Hooks cannot be used in a class component. Making the example a function component would be ideal.

MarcoScabbiolo (changes requested):

docs/react-navigation.md
ThaJay: Ah yes I missed that. Will do.

ThaJay: I think I have changed it properly now.

MateusAndrade (approved):

MarcoScabbiolo (approved):

PR #3107, opened 2021-02-10
Fix onDoubleTap() not getting triggered in iOS

tam-nguyen-senseye:
`onDoubleTap()` is not getting triggered in iOS. This turns out to be due to a misspelling in the field name to which the boolean `isDoubleTap` is passed in.

Fixing this enables `onDoubleTap()` to be triggered as expected in iOS.


MarcoScabbiolo (approved):

MarcoScabbiolo:
Nice catch, thanks!

MateusAndrade (approved):

PR #3109, opened 2021-02-11
hasTorch method

theasc:


ios/RN/RNCameraManager.m
```diff
@@ -550,6 +550,20 @@ + (NSDictionary *)barcodeDetectorConstants
         }];
 }
 
+RCT_EXPORT_METHOD(hasTorch: (RCTPromiseResolveBlock)resolve
+                  rejecter:(RCTPromiseRejectBlock)reject) {
+    NSArray *devices = [AVCaptureDevice devicesWithMediaType:AVMediaTypeVideo];
```
MarcoScabbiolo, line 555: I think `videoCaptureDeviceInput?.device` and `getDevice()` from `RNCamera` should be used to get the device

theasc: Using this approach we can't call this method when the camera is not mounted, I think it's better to have the information whether the camera is mounted or not.

MarcoScabbiolo: Yes, it would be nice to have a method to detect the feature outside the camera component, but right now it is a method of the camera component. For the iOS implementation the feature detection is done on a specific device and it is best to detect the torch on the device that is currently being used

cristianoccazinsp:
Looks like a nice addition.

MateusAndrade:
@theasc  do you know if this implementation is working for iOS? I can test the android code, but since I don't have a device for testing on iOS it wouldn't be good to follow with this without the proper tests. 😢

cristianoccazinsp:
The code looks ok to me, I haven't tested it at all, but I assume the author did test it.  It's a very simple implementation, and while I agree that it would be better to have the method work on the current device, it's probably not worth the effort since it is safe to assume that most devices will have a flash on the back cameras (and not on the front cameras).

MateusAndrade:
> The code looks ok to me, I haven't tested it at all, but I assume the author did test it. It's a very simple implementation, and while I agree that it would be better to have the method work on the current device, it's probably not worth the effort since it is safe to assume that most devices will have a flash on the back cameras (and not on the front cameras).

Do you think we can go on merge with this? We can easily roll back this if there is any problem regarding this feature. 🤔

cristianoccazinsp:
Yes, it's fine by me.

Perhaps the author can say if he tested it on both android and iOS just to
be sure.

ios/RN/RNCameraManager.m
mostasim: is this method available? I don't find the method in my project.
 version "react-native-camera": "3.43.8"

theasc:
@cristianoccazinsp Yes, I've tried on both android and IOS and works

hamzaouhssain1993:
@cristianoccazinsp and @MateusAndrade  do you think it is possible to be marged? thank you

cristianoccazinsp:
@hamzaouhssain1993 for me the changes look good, and should work for most cases where a simple check is needed.

I do still have the following suggestions if you want to do another PR or improve this one to make the method more useful:

For iOS, `[device hasTorch]` and `[device isTorchModeSupported:AVCaptureTorchModeOn]` should both be used, as torch is not the same as flash, I think? So for iOS, it would be great for the method to also return if it supports "torch mode".

Additionally, could the methods receive the cameraId returned by `getCameraIds` ? That way, if given, you could check the if flash is supported for that specific camera. 

Either way, it's probably good to merge as it is right now and do the above later.

Thanks!

MateusAndrade (approved):

MateusAndrade:
It looks like that the base repository has been deleted. Because of that, I wasn't able to merge this @theasc.

hamzaouhssain1993:
here the pull request  #3340

PR #3113, opened 2021-02-18
Improve hardware iOS orientation sensor usage.

cristianoccazinsp:
This PR improves how the physical device orientation and sensors are used in order to not rely on UI/Status bar orientation.

The rationale behind this change, is that for some reason accelerometer updates stops working when other native code also uses it. However, using `startDeviceMotionUpdatesToQueue` instead of accelerometer data, correctly coexists with other libraries.

Sensor data is now used while the camera is mounted, and orientation data is immediately available and doesn't need to be polled. Additionally, it does not use a custom queue to poll for sensor data, which fixes issues (for unknown reasons) when other native code uses motion sensors.

Also, revert photo preset change to Photo quality to restore high quality photos on iOS.

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -283,6 +283,7 @@ BOOL _sessionInterrupted = NO;
         // after mount to set the camera's default, and that will already
         // this method
         // [self initializeCaptureSessionInput];
+        [self.sensorOrientationChecker start];
         [self startSession];
     }
     else{
@@ -296,6 +297,7 @@ BOOL _sessionInterrupted = NO;
 
         [[NSNotificationCenter defaultCenter] removeObserver:self name:AVAudioSessionInterruptionNotification object:[AVAudioSession sharedInstance]];
 
+        [self.sensorOrientationChecker stop];
         [self stopSession];
     }
 
@@ -727,15 +729,17 @@ BOOL _sessionInterrupted = NO;
 }
 
 - (void)takePictureWithOrientation:(NSDictionary *)options resolve:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRejectBlock)reject{
-    [self.sensorOrientationChecker getDeviceOrientationWithBlock:^(UIInterfaceOrientation orientation) {
-        NSMutableDictionary *tmpOptions = [options mutableCopy];
-        if ([tmpOptions valueForKey:@"orientation"] == nil) {
-            tmpOptions[@"orientation"] = [NSNumber numberWithInteger:[self.sensorOrientationChecker convertToAVCaptureVideoOrientation:orientation]];
-        }
-        self.deviceOrientation = [NSNumber numberWithInteger:orientation];
-        self.orientation = [NSNumber numberWithInteger:[tmpOptions[@"orientation"] integerValue]];
-        [self takePicture:tmpOptions resolve:resolve reject:reject];
-    }];
+    
+    UIInterfaceOrientation orientation = [self.sensorOrientationChecker getDeviceOrientation];
+    
+    NSMutableDictionary *tmpOptions = [options mutableCopy];
+    
+    if ([tmpOptions valueForKey:@"orientation"] == nil) {
+        tmpOptions[@"orientation"] = [NSNumber numberWithInteger:[self.sensorOrientationChecker convertToAVCaptureVideoOrientation:orientation]];
+    }
+    self.deviceOrientation = [NSNumber numberWithInteger:orientation];
+    self.orientation = [NSNumber numberWithInteger:[tmpOptions[@"orientation"] integerValue]];
+    [self takePicture:tmpOptions resolve:resolve reject:reject];
 }
 
 - (void)takePicture:(NSDictionary *)options resolve:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRejectBlock)reject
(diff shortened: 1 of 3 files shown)


cristianoccazinsp:
Although using the sensor this way makes it bug less frequently, it is still happening. Sensor stops working / detecting randomly when other libraries use it, making the camera unable to capture (and never resolving any promise).

Still debugging.

MarcoScabbiolo:
Might have to do with the way the orientation is retrieved, which is confusing right now:
https://github.com/react-native-camera/react-native-camera/blob/6a81dea6574389add42e4d3f005dabda89115b8c/ios/RN/RNCamera.m#L730

v4 has a different approach, it sets the orientation checker and subscribes to the device orientation
https://github.com/react-native-camera/react-native-camera-v4/blob/swift/packages/core/ios/Services/SensorOrientationChecker.swift

I'm not really sure what will work best and if there are any improvements to be made in the motion/acceleration itself, I'm just pointing out what could be causing the underlying issue.

cristianoccazinsp:
Does it mean it will be listening to the sensors constantly? When I was testing this, and accidentally forgot to turn off the sensor listener, the issue didn't happen as well.

However, with the current RNCamera implementation, there's a very strange behaviour that makes the component stay alive in memory even after unmounting it (perhaps it is react-native keeping it around just in case), so it is a bit dangerous to keep the sensor listening for changes.  Did you see this issue when testing with the swift version?

MarcoScabbiolo:
Couldnt test the swift version yet, it's not ready.

About keeping the sensor listening as long as the component is mounted, we would have to measure the performance impact but I don't think it should be a problem.

About the component and its resources not being released when it unomunts, that's a serious bug if its happening.

Can your create an issue for the bug? Maybe addressing it in v4 would be the best, unless you manage to get it working in v3 of course.

cristianoccazinsp:
The component not being released (View) bug is something I saw a long time ago and wasn't able to fix at all.  I suspect is really a RN thing, since at most 2 instances of the view are kept in memory.  My "fix" at that time, was to have the camera and event listeners be released and acquired not in the constructor, but rather on the `willMoveTo` method, so we detect when the view is detached from the hierarchy.

cristianoccazinsp:
I was trying your code above in this custom component that uses ARKit and bugs out RNCamera sensor usage, and voila! The same happens... Now, this is interesting.

Screen Shot 2021-02-19 at 11 09 22

Definitely not an issue with RNCamera. However, looks like the the method from this PR is slightly better than the original one that falls back to using the statusBar orientation.  Do you think the same would work in your Swift version? Any reason to not use this?

I will keep debugging to see what happens with the sensors... Perhaps it's a hardware issue.

MarcoScabbiolo:
Wouldn't be better to use https://developer.apple.com/documentation/uikit/uidevice ?

cristianoccazinsp:
I did test that option, but that one doesn't return the physical device orientation if it is "locked", which in turn will give you rotated photos if your app is either locked or only uses portrait. But yes, ideally, no core motion data should be used!

cristianoccazinsp:
Looks like not using a queue, but rather just calling `startUpdates` and reading the values directly do not bug out.  This means your idea of turning on the sensor for the life of the camera is probably better to avoid this issue.

cristianoccazinsp:
@MarcoScabbiolo about the Camera views being created more than once and not released, this gives a hint as of why it could happen: https://github.com/facebook/react-native/issues/725

Because of that, it is not good to keep listeners/sensors active on the init method and rely on the dealloc to terminate them when using UIViews.  If you add simple logs either on the RNCamera view manager or initWithBridge method of the view, you will clearly see how the view is ALWAYS created twice, but only disposed once.  I dug some good time to understand why this happens, but found no answer.  Interestingly though, this doesn't seem to happen to other native views implemented in Swift.

MarcoScabbiolo:
Thanks for the insights @cristianoccazinsp , really helpful info. I think we should address all of this on v4.

About the view lifecycle, as a last resort we can decouple all the logic from the view and store camera sessions in the module instead of the view, using the view manager as an orchestrator between both. But we might find a less drastic fix, what is not acceptable is having the entire session in an object that has the potential to be leaked/duplicated.

cristianoccazinsp:
It would be great to know why the component is instantiated twice, but either way, using the view life cycle methods of the view (attached/detatched) should be good enough, and relatively easy to do in swift as well.

cristianoccazinsp:
Small update:

Looks like the change to the sensor usage works, I haven't been able to get it stuck anymore when using other libraries that use the sensors. Perhaps it is an Apple bug? 

For the duplicated view, I believe it happens when RCT_CUSTOM_VIEW_PROPERTY is used.  React native then creates a "fake" or shadow node to manage custom properties: https://github.com/facebook/react-native/blob/master/React/Views/RCTComponentData.m#L87

Probably can't do much about it, and just live with the fact that a View instance may live throughout the entire life of the app when custom properties are used.

MateusAndrade (approved):

MarcoScabbiolo (approved):
Changes look good, only thing I'm not sure about is how the change in the algorithm will affect end users, but since the iOS sdk is pretty stable in terms of behavior it shouldn't be a problem if testing is done. @cristianoccazinsp it's your call to merge or not, by this point I think all of us trust your judgement and testing. Thanks for your continuous support for the project, it's great to have you around.

cristianoccazinsp:
Thanks guys! If you want let's hold the merge for a week or so until we get
more testing in production from our app. We have a good user base so we
will be able to catch any additional edge case.

cristianoccazinsp:
Seems to be working fine, but we still haven't got enough testing from every iOS device.

cristianoccazinsp:
Has been working great without issues. I've just merged automatically from master since it was outdated, hopefully that didn't add any odd merge issue.

PR #3117, opened 2021-02-19
Use empty string for undefined cameraId

cristianoccazinsp:
Following up from here (https://github.com/react-native-camera/react-native-camera/pull/3113) .  

The reason the RNCamera component is always rendered twice (aka, a `defaultView` is created on the native side) is because of `RCT_CUSTOM_VIEW_PROPERTY` combined with a property that is set to `null`. When this happens, RN creates a "fake" default view to store default props.  Although this works fine most of the time, it creates some unnecessary classes and resources being kept in memory even after the camera unmounts.

After digging a bit, the prop `cameraId` being null is the sole reason RN creates this `defaultView`.  This PR uses an empty string as the "no value" value instead, removing the extra view creation.

Note: The change is not breaking, anyone using `cameraId=null` will continue to work as before. However, to prevent this extra view from being created, the prop must be set to `cameraId=''` instead.

Lastly, this only shows how we must be careful with `null` properties and resources acquired/released in the init/dealloc methods!


cristianoccazinsp:
@MarcoScabbiolo  sorry that I keep adding changes that just adds more stuff for the v4 migration, this should be the last one! But since we constantly use this library, these fixes are a must for us.

MarcoScabbiolo (approved):
OK, didnt test myself.

MarcoScabbiolo:
> @MarcoScabbiolo sorry that I keep adding changes that just adds more stuff for the v4 migration, this should be the last one! But since we constantly use this library, these fixes are a must for us.

No problem, I'm keeping track of all the changes to make sure they land in v4 as well. Thanks for contributing.
BTW in v4 only `RCT_EXPORT_VIEW_PROPERTY` is used, so this should be fixed out of the box.

MateusAndrade (approved):

PR #3172, opened 2021-04-02
Fix: iOS mirroring video always outputs low quality video

thessler27:
Currently, using the prop `mirrorVideo` the mirrored output only delivers in 480 resolution. 

This PR configures the output from mirroring the camera output to align with the session-dictated AV preset, rather than hard-coding it.


MateusAndrade (approved):
Thanks! 🚀 🚀 🚀 🚀

Issue #3050, opened 2020-12-15
Device doesn't support flash mode


Romick2005:
I got warning every time I use camera on my old iPhone 5s. 
Flash is not disabled and works properly with camera.

On earlier versions of react-native-camera everything was working as expected.

[RNCamera updateFlashMode]: device doesn't support flash mode

hamzaouhssain1993:
I do not understand. There is a method inside the project that is "hasFlassh()" but we are not able to use it. and for that we are not able to disable the flash if the camera does not support it at all.

caojianfeng:
Same issues on ipad mini4 (ios 13.5.1)

gyoung-panopto:
I'm hitting this issue on an iPad as well. Fortunately it just seems to warn and doesn't affect the behavior, but the log spam is annoying.

Romick2005:
> I do not understand. There is a method inside the project that is hasFlash() but we are not able to use it. and for that we are not able to disable the flash if the camera does not support it at all.

Device camera actually support flash mode, but the library give a warning that saying the opposite.
Tested iOS 12.5.1

tsoisauce:
Having the same issue, creating an internal business tool for iPad that reads barcodes and I am unable to exclude `flashMode`

Any help would be kindly appreciated.

Romick2005:
#3183 I removed this annoying unneeded warning in this pull request.
AS for me initially I assumed that device = iPhone and I was using iPhone 5s that has two camera (back camera has flash mode and front camera does not have flash mode). So I was thinking that this message is wrong as my device support flash mode. But after short investigation I discovered that device is not an iPhone itself it is only camera device so than when I was using front camera and in the app I use only front camera, I got this warning, but it doesn't tell anything useful or something that I can fix so to me it seems that a better solution for this is just to remove this warning as code logic does not brake and works as expected.

Romick2005:
Until those warning are removed I use:
`
LogBox.ignoreLogs(
  [
    "[RNCamera updateFlashMode]: device doesn't support flash mode"
  ]
);
`

PR #3183, opened 2021-04-14
#3050 Remove annoying  warning 

Romick2005:
This warning actually does not warn as there are no way to fix it and it has no impact on the updateFlash method success execution.


MateusAndrade:
What would be the benefits of removing this? Maybe, it would be harder to debug some stuff if we remove this. What about if we call this only when we're with `debug=true`? I don't know if this is possible, because I'm don't know a lot about Objective-C. 😢

Romick2005:
Yes there is a way for this like:

'''
#ifdef DEBUG
// debugging code
#endif
'''

But again this warning is showing only in debug mode as those warning do not popup in production release. I do not see how this warning could help. From my point of view if I use library correctly I should not see this warning that doesn't help me. This warning is only annoying and have no benefits as for me. 
But I am happy to hear when this warning can help, probably I am miss something.

Issue #2933, opened 2020-07-31
Video recording doesn't work for ChromeBook


Teal-labs-developer:
# Bug Report

**To Do First**

- [x] Did you try latest release?
- [x] Did you try master?
- [x] Did you look for existing matching issues?

**Platforms**

**Versions**

- Android: 9, ChromeOS
- react-native-camera: 3.31.1
- react-native: 0.62.2
- react: 16.13.1
  

**Description/Current Behaviour**

When I start to record the video on a Chrome Book by calling recordVideoAsync(), it doesn't start. Here is the log that comes on logcat
E/CAMERA_1::: Record start failed
    java.lang.NullPointerException: Attempt to read from field 'int android.media.CamcorderProfile.quality' on a null object reference
        at com.google.android.cameraview.Camera1.setUpMediaRecorder(Camera1.java:1521)
        at com.google.android.cameraview.Camera1.record(Camera1.java:812)
        at com.google.android.cameraview.CameraView.record(CameraView.java:631)
        at org.reactnative.camera.RNCameraView.access$2601(RNCameraView.java:37)
        at org.reactnative.camera.RNCameraView$3.run(RNCameraView.java:314)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
        at android.os.HandlerThread.run(HandlerThread.java:65)

**Additionals**

> Love react-native-camera? Please consider supporting our collective: 👉 https://opencollective.com/react-native-camera/donate
> Want this issue to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

Issue #2805, opened 2020-04-26
Attempt to read from field 'int android.media.CamcorderProfile.quality' on a null object reference


zhonghanwen:
# Bug Report

**To Do First**

- [x] Did you try latest release?
- [x] Did you try master?
- [x] Did you look for existing matching issues?

**Platforms**

Android

**Versions**

- Android:
- react-native-camera: 3.23.1
  

**Description/Current Behaviour**

**Expected Behaviour**

**Steps to Reproduce**

```java
2020-04-26 19:25:19.369 10952-18845/com.ainirobot: E/CAMERA_1::: Record start failed
    java.lang.NullPointerException: Attempt to read from field 'int android.media.CamcorderProfile.quality' on a null object reference
        at com.google.android.cameraview.Camera1.setUpMediaRecorder(Camera1.java:1516)
        at com.google.android.cameraview.Camera1.record(Camera1.java:812)
        at com.google.android.cameraview.CameraView.record(CameraView.java:632)
        at org.reactnative.camera.RNCameraView.access$2601(RNCameraView.java:31)
        at org.reactnative.camera.RNCameraView$3.run(RNCameraView.java:303)
        at android.os.Handler.handleCallback(Handler.java:751)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:154)
        at android.os.HandlerThread.run(HandlerThread.java:61)
2020-04-26 19:25:19.377 10952-18813/com.ainirobot E/ReactNativeJS: , 'Starting video recording failed. Another recording might be in progress.'
```

**Additionals**
**My android Device is no Black type camera.**
I think here should use `CamcorderProfile.get(cameraId, CamcorderProfile.QUALITY_HIGH);`
![图片](https://user-images.githubusercontent.com/12220537/80308349-152f7d00-8801-11ea-9b6e-3a1f09a36d6e.png)
![图片](https://user-images.githubusercontent.com/12220537/80308525-1f9e4680-8802-11ea-89b6-e9a5165b7e54.png)
![图片](https://user-images.githubusercontent.com/12220537/80308935-48274000-8804-11ea-88ea-b287b74ec91c.png)

kmcmanimen:
If anyone comes across this, the issue for us was related to a device that does not have a back facing camera.  When passing null the camera would throw an error and not proceed.  We solved the issue by patching the library to set the default if the profile is null:

```
[...]

PR #3259, opened 2021-06-17
Use specified cameraId on record

akadko:
Right now, on Android library ignores the specified camera ID set via prop `cameraId` on record start. It creates `CamcorderProfile` without specifying camera ID and that leads to exception on devices with only one front-facing camera (e.g. Chromebook). See: #2933, #2805. This PR takes into consideration the camera ID on creation of CamcorderProfile.


cristianoccazinsp:
Isn't `cameraId` a string? Although it is an int with Camera1, I believe it is a string with Camera2 (and on iOS), so you may want to handle that case.

PR #3269, opened 2021-07-02
iOS - Fix orientation value in fast consecutive calls

cristianoccazinsp:
This update makes sure that orientation value stays consistent from the beginning to the end of the capture process.

Right now, calling capture multiple times quickly may end up with invalid orientation values due to the fact that it is set at the beginning and unset at the end with a shared instance variable.

The change basically stores the value locally so it stays consistent throughout the entire capture process. Note that it is synchronized so multiple consecutive calls do not break the value in between.

Ultimately, the whole `takePictureWithOrientation:options` should be removed since orientation is now always readily available and an async call is no longer needed, but that can wait for another PR.

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -750,13 +750,24 @@ BOOL _sessionInterrupted = NO;
         return;
     }
 
-    if (!self.deviceOrientation) {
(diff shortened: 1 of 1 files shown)


MateusAndrade (approved):
Thanks! 🚀

PR #3292, opened 2021-07-22
fix the format finding logic when fps is specified

### LINKED ISSUE #3291: recordAsync(): If `fps` is specified, format selection is unstable.
# Bug Report

**To Do First**

- [x] Did you try master?

**Platforms**

iOS

**Versions**

- Android:
- iOS: 14.5.1
- react-native-camera: 4.0.0
- react-native: 0.64.2
- react: 17.0.2
  

If `fps` is specified in the option of `recordAsync()`. Format selection looks unstable.
Specifically, [this condition](https://github.com/react-native-camera/react-native-camera/blob/215b5643ba7f8c5e291958a51f92be310589a2f1/ios/RN/RNCamera.m#L116
) checks that the width of the new format fits the current one, but not for height.
Also, seems it doesn't need to store `maxWidth` for me. Actually, with my device, a 1440 x 1920 px video has been recorded with `1080p` setting. It should record a 1080 x 1920 px video. I'm assuming this is from the above line is checking only the format width.

Additionally, it shows a non-deterministic behavior. I see this issue from time to time, not always.

If it's intentional behavior, could I hear about the reason here a little bit?

Thank you!

**Expected Behaviour**
It should select a format that has the correct dimension.

**Steps to Reproduce**

Call `recordAsync({fps: 60}` with the 1080p `defaultVideoQuality` prop
(As this is non-deterministic, you might not see it occasionally.)

**Additionals**

> Love react-native-camera? Please consider supporting our collective: 👉 https://opencollective.com/react-native-camera/donate
> Want this issue to be resolved faster? Please consider adding a bounty to it https://issuehunt.io/repos/33218414

Sidebook:
Fix #3291

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -1160,20 +1160,20 @@ BOOL _sessionInterrupted = NO;
 
             AVCaptureDeviceFormat *selectedFormat = nil;
             int32_t activeWidth = activeDimensions.width;
-            int32_t maxWidth = 0;
+            int32_t activeHeight = activeDimensions.height;
 
             for (AVCaptureDeviceFormat *format in [device formats]) {
                 CMFormatDescriptionRef formatDescription = format.formatDescription;
                 CMVideoDimensions formatDimensions = CMVideoFormatDescriptionGetDimensions(formatDescription);
                 int32_t formatWidth = formatDimensions.width;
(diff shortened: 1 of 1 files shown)


cristianoccazinsp:
Looks good to me.

renatobenks (approved):
sounds good

PR #3293, opened 2021-07-22
Bugfix [iOS]: isRecordingInterrupted is not being set to `true` when it should

tam-nguyen-senseye:
I was encountering unexpected behavior in iOS where if I call `recordAsync()` and then minimize or navigate away from the app during the recording, `isRecordingInterrupted` will still return as `false`. Meanwhile, testing the same scenario in Android produces `true` as expected.

It would seem like this condition is already being handled via an event handler (`sessionWasInterrupted`). However, I found that whenever a session is interrupted, the recording automatically stops and propagates `isRecordingInterrupted` before the event handler has a chance to make the intended update.

To fix this, I make use of captureOutput's `error` param, which provides the following info whenever a recording is stopped due to an interrupted session:
```
Domain=AVFoundationErrorDomain
Code=-11818 "Recording Stopped"
UserInfo={
  AVErrorRecordingSuccessfullyFinishedKey=true,
  NSLocalizedDescription=Recording Stopped,
  NSLocalizedRecoverySuggestion=Stop any other actions using the recording device and try again.,
  AVErrorRecordingFailureDomainKey=1,
  NSUnderlyingError=0x2819d2850 {Error Domain=NSOSStatusErrorDomain Code=-16414 "(null)"}
}
```

### DIFF
--- a/ios/RN/RNCamera.m
+++ b/ios/RN/RNCamera.m
@@ -1893,6 +1893,10 @@ BOOL _sessionInterrupted = NO;
         if (value) {
             success = [value boolValue];
         }
+        // Check if the recording stopped due to the session being interrupted.
+        if ([error code] == AVErrorSessionWasInterrupted) {
+          self.isRecordingInterrupted = YES;
+        }
     }
     if (success && self.videoRecordedResolve != nil) {
         NSMutableDictionary *result = [[NSMutableDictionary alloc] init];


cristianoccazinsp:
Nice catch, looks good to me! Can you make sure your branch is up-to-date with master?

Do you know any specific scenario where `sessionWasInterrupted` is not called and this is in fact used?

tam-nguyen-senseye:
> Do you know any specific scenario where `sessionWasInterrupted` is not called and this is in fact used?

I've tested scenarios where I either press the power button (turn off screen) or swipe up (minimize the app) and haven't found an instance where `sessionWasInterrupted` is not called -- from my debugging via logs and tracking their execution order, `sessionWasInterrupted` is always called, but only after `AVCaptureMovieFileOutput.captureOutput()`, at which point, line 1703 (`if ([self isRecording])`) resolves to false anyway so it never updates `isRecordingInterrupted`. Also, I believe the error code that's propagated to `captureOutput` stems from the same event that triggers `sessionWasInterrupted`, so I don't foresee them covering disjointed scenarios. If anything, I think the error code that this solution checks is a clear indication for setting `isRecordingInterrupted`, so it may be redundant to set it inside `sessionWasInterrupted` at all.

Other scenarios I've tested, like pulling down to view the control center or swiping up + hold to view other running apps, don't seem to trigger an "interruption" in iOS's eyes, so neither `sessionWasInterrupted` nor this solution are hit in these instances, but that's pre-existing behavior and I assume is expected.

PR #3305, opened 2021-08-05
Workaround for hardware JPEG rotation bug on some Samsung devices

drash-course:
### Description

This PR implements a workaround to avoid this issue: https://github.com/react-native-camera/react-native-camera/issues/3060

### What is the root cause

Android's `Camera.PictureCallback::onPictureTaken(byte[] data, Camera camera)` is returning a corrupted file. The byte[] data array is full of 0x00 after a few kB of data. This happens when mCameraParameters.setRotation(...) is set to any rotation besides 0 on our Galaxy Tab S2 9.7 Wi-Fi SM-T813.

### What is the workaround / fix 

The facilities in `ResolveTakenPictureAsyncTask.java` can handle the rotation in software. The workaround is to set `mCameraParameters.setRotation(0)`, take the picture, and perform the rotation in the async task. To differentiate this "backup" rotation from the rotation performed when "fixOrientation" is set, a new parameter `softwareRotation` has been added to `ResolveTakenPictureAsyncTask`.

The `softwareRotation` is only used for the known affected devices, set in `BROKEN_ROTATION_DEVICE_MODELS`. Other devices will continue to use hardware rotation (which is probably faster).

### How to test

If you don't have an affected device on hand, you can still test this alternate implementation by changing `Camera1::fallbackToSoftwareRotation()` to return true or add your test device model to `BROKEN_ROTATION_DEVICE_MODELS`.

### What can be improved

I only added "SM-G532M" and "SM-T813" which we know have this problem. More devices are probably affected. Since it seems to always be Samsung devices, we could enable this alternate rotation mode for all of them.

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -52,7 +52,7 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
 
     private static final SparseArrayCompat<String> FLASH_MODES = new SparseArrayCompat<>();
 
-    private static final String[] BROKEN_ROTATION_DEVICE_MODELS = {"SM-G532M", "SM-T813"};
+    private static final String[] BROKEN_ROTATION_DEVICE_MODELS = {"SM-G532M", "SM-T813", "SM-T307U", "SM-T713"};
 
     static {
         FLASH_MODES.put(Constants.FLASH_OFF, Camera.Parameters.FLASH_MODE_OFF);
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -179,12 +179,11 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
                         // and we might need to re-create it and restart the camera
                         surfaceWasDestroyed = true;
 
-                        try{
+                        try {
                             mCamera.setPreviewCallback(null);
                             // note: this might give a debug message that can be ignored.
                             mCamera.setPreviewDisplay(null);
-                        }
-                        catch(Exception e){
+                        } catch (Exception e) {
                             Log.e("CAMERA_1::", "onSurfaceDestroyed preview cleanup failed", e);
                         }
                     }
@@ -266,18 +265,16 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
         // or host destroyed. Should avoid crashes with concurrent calls
         synchronized(this){
             if (mMediaRecorder != null) {
-                try{
+                try {
                     mMediaRecorder.stop();
-                }
-                catch(RuntimeException e){
+                } catch (RuntimeException e) {
                     Log.e("CAMERA_1::", "mMediaRecorder.stop() failed", e);
                 }
 
-                try{
+                try {
                     mMediaRecorder.reset();
                     mMediaRecorder.release();
-                }
-                catch(RuntimeException e){
+                } catch (RuntimeException e) {
                     Log.e("CAMERA_1::", "mMediaRecorder.release() failed", e);
                 }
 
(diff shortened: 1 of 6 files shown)


cristianoccazinsp:
This looks promising. We have been getting reports of gray images on SM-T813 devices too and just assumed the devices were too old.  

"Blacklisting" all samsung devices is probably a really bad idea, but using this kind of list of affected devices should be good.  I don't know how this list could be maintained easier than what you have done so far.

Any issues on unaffected devices? Slower captures or something?

cristianoccazinsp:
Also, do you know if the code could be improved in a way that `mCameraParameters.setRotation` is not needed?

Here's another device we saw the gray photos issue: SM-T307U

drash-course:
> Any issues on unaffected devices? Slower captures or something?

I tested this on SM-T813 with 1920x1080 photos, could not tell the difference in performance. Did not measure it precisely though.

> Isn't it dangerous to change the camera parameters right before requesting a capture?

The previous implementation changes parameters before capture as well. The Android documentation states that's how you do it (https://developer.android.com/reference/android/hardware/Camera). We do the calls on the same thread so no concurrency issues here.

> Also, do you know if the code could be improved in a way that mCameraParameters.setRotation is not needed?

I don't think so. `mCameraParameters.setRotation(...)` is also set before you take a picture, e.g. in `Camera1::setDeviceOrientation` and `Camera1::adjustCameraParameters`. It's set when you take a picture if you provide the `"orientation"` option when calling `takePictureAsync(options)` from JS.

My fix relies on the fact that `Camera1::takePictureInternal` will do `mOrientation = options.getInt("orientation");` if the user provides a specific orientation. For broken devices I do `int requestedRotation = calcCameraRotation(orientationEnumToRotation(mOrientation));` so it will respect the option if provided. Otherwise it uses the `mOrientation` value, just like the normal implementation.

cristianoccazinsp:
Thanks for the details, I believe it makes sense.  Impressive detective work finding this bug!

I need some time before being able to test this as I have to catch up with the MLKit changes from another PR, but I would love to see this getting merged. 

Thoughts @MateusAndrade @mikehardy ?

android/src/main/java/com/google/android/cameraview/Camera1.java
```diff
@@ -763,6 +769,21 @@ void takePictureInternal(final ReadableMap options) {
...
+                int requestedRotation = calcCameraRotation(orientationEnumToRotation(mOrientation));
+                if (requestedRotation != 0 && fallbackToSoftwareRotation()) {
+                    softwareRotation = requestedRotation;
+                    mCameraParameters.setRotation(0);
+                    try{
+                        mCamera.setParameters(mCameraParameters);
+                    }
+                    catch(RuntimeException e ) {
+                        Log.e("CAMERA_1::", "setParameters 0 rotation failed", e);
+                    }
```
mikehardy, line 782:
```suggestion
                    try {
                        mCamera.setParameters(mCameraParameters);
                    }
                    catch (RuntimeException e) {
                        Log.e("CAMERA_1::", "setParameters 0 rotation failed", e);
                    }
```

inconsistent use of blank space is not important for functionality but is an eyesore while reviewing :eyes:

mikehardy (approved):
seems fine to me. In my experience Samsung has some buggy hardware (they messed up webview hardware rendering for Android 8 too, that was fun to find) just like MIUI has a buggy AOSP implementation. There's nothing to do but trust those that are directly probing the affected devices and accept their workarounds I think.

This seems like a nice way to handle an irritating problem :+1:

android/src/main/java/com/google/android/cameraview/Camera1.java
drash-course: There are many of them. I pushed a lint version for Camera1.java

mikehardy: Yeah - I noticed the whole file itself was inconsistent, I didn't note the ones where your change was at least the same as other ones, just the one that was new code. But linting in general makes me happier than it should ;-)

MateusAndrade (approved):

drash-course:
@cristianoccazinsp Would it be possible to have this fix in the 3.44 line as well? That's what we use since we aren't ready to move to Google ML Kit for all projects.

cristianoccazinsp:
@drash-course that's really up to @MateusAndrade .  However, if you are in such situation, I recommend forking the 3.44 yourself and cherry picking specific commits you may be interested in (like this). You can then change your package file to point to your github fork/branch instead.  

This is what we also do since we stay behind the master branch a few commits and gradually merge from it in order to test all changes thoroughly first and be able to use important fixes right away.

PR #3320, opened 2021-08-18
Missing TS type definition

theasc:
We still have the option to provide a custom path when taking pictures, this also should be added to the type definition.


theasc:
Also added return type for method `getCameraIdsAsync`

PR #3324, opened 2021-08-24
implement onCameraReady event on Windows

burtonrodman:
when repeatedly clicking to flip front/back cameras, the app would crash.  noticed that some docs/examples relied on the onCameraReady event to prevent this.  however, this was not implemented for Windows.


MateusAndrade (approved):

burtonrodman:
you're awesome!  thanks!

PR #3331, opened 2021-08-26
Fix Android camera unmount ANR or Freezes

cristianoccazinsp:
Remove the `mCamera.setPreviewDisplay` call on surface destroy as it may cause ANRs or slow down the UI when unmounting.  

The side effect of this is a nasty warning `BufferQueue has been abandoned` that can be ignored. 

I will be testing this on various other devices to ensure the buffer warning is still a warning and does not cause surprise crash on different devices.

### DIFF
--- a/android/src/main/java/com/google/android/cameraview/Camera1.java
+++ b/android/src/main/java/com/google/android/cameraview/Camera1.java
@@ -177,12 +177,17 @@ class Camera1 extends CameraViewImpl implements MediaRecorder.OnInfoListener,
                 // it may have been destroyed already.
                 // Do not lock as it would freeze the UI if the camera is being released
                 if (mCamera != null) {
-                    try {
-                        // mCamera.setPreviewCallback(null); // Not needed as stop() already clears it
-                        mCamera.setPreviewDisplay(null); // needed to prevent buffer abandoned logs
-                    } catch (Exception e) {
-                        Log.e("CAMERA_1::", "onSurfaceDestroyed preview cleanup failed", e);
-                    }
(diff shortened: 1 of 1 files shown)


fedeerbes:
Hi @cristianoccazinsp we're experiencing some issues on low-mid end devices related to unmount of RNCamera. #3346 
I tried your solution on one of my Samsung SM-J600G devices (where the problem is reproducible) and it works. Do you think we can make this PR open to review/discussion?

@MarcoScabbiolo @fabriziobertoglio1987 sorry to tag you folks, but what do you think about this PR?

cristianoccazinsp:
I haven’t seen issues with these changes neither, but we still haven’t got enough users to test the latest update with this change, reason I’m waiting to see if another ANR pops up. 

How were you able to reproduce the issue? I haven’t been able to consistently reproduce it, all I know is that it randomly happens.

fedeerbes:
@cristianoccazinsp take a look at #3346 there I left an explanation of what we're doing. But in short we have consecutive screens, each one with an RNCamera component. If the user goes back or forth in the navigation even if we unmount the RNCamera, we get this error on some devices.

MateusAndrade (approved):

MateusAndrade:
Do you think we can follow with this @cristianoccazinsp ? We can easily revert this if with start noticing any problem.

cristianoccazinsp:
@MateusAndrade yes, I think it is safe to merge and release.  With Android nothing is ever 100% sure, so I can't guarantee that it won't fail on some devices.

Idris634 (approved):
[]()

PR #3360, opened 2021-10-01
Fix/camera api2 flash

SimonErm:
Currently the flash isn't triggered on android when using camera2api enabled.
This PR fixes this bug by triggering the precapture.


SimonErm:
This seems to be the wrong way, i will do another PR.

PR #3372, opened 2021-10-15
Fix race condition on start/stop

cristianoccazinsp:
Under some circumstances (e.g., system alert/permission prompt), the pause/resume/destroy events may be called one after another. Without this delay and check, we would end up having a failed to start event (because the camera is started when it is being stopped). The delay is minimal and should have no visual impacts.

Will be testing this for a bit.


cristianoccazinsp:
Seems to be working fine.

PR #3384, opened 2021-11-19
Account for screen orientation when passing rotation values to detector tasks

### LINKED ISSUE #2670: Barcode/Text feature bounds incorrect on rotation without change in orientation
# Bug Report

- Did you try latest release? Yes - 3.15.1
- Did you try master? No
- Did you look for existing matching issues? Yes

**Platforms**

Tested on Android Only

- Android: 10, Samsung S10+, One UI 2.0
- react-native-camera: 13.5.1
- react-native: 0.61.5
- react: 16.9.0
- react-navigation: 4.0.10

On rotating a phone with One UI, without auto-rotate enabled, the user can press a button to change the phones orientation otherwise nothing happens.

When rotating the phone and NOT pressing the button, the bounds for barcodes and text scanned by RNCamera (haven't tested facial features) are rotated anyway, causing them to be in incorrect positions on the screen.

**Expected Behaviour**

When the camera is rotated but orientation has not changed, feature positions do not get rotated.

**Steps to Reproduce**

- Enable google barcode detection, and record the position
- Check how the position changes normally (pitching up and down changes y dimension, etc)
- Rotate the device WITHOUT changing orientation, then check position again, this time
pitching up and down changes the x dimension.

**Additionals**

My explanation may be a bit confusing sorry. I have attached a video that shows more clearly what is happening.

In the video I first move the camera whilst the device is vertical. Then I rotate the device without changing orientation, and move the camera. Then I change the orientation. The yellow dot is calculated to be at the centre of the bounds given by onGoogleVisionBarcodesDetected.

[Link to video](https://thumbs.gfycat.com/UnawareAcrobaticBedbug-mobile.mp4)

I've tested this with onGoogleVisionBarcodesDetected, and onTextRecognized

chiubaka:
Resolves #2670.

**NOTE**: Makes the (potentially bad) assumption that natural device orientation is always either `PORTRAIT` or `LANDSCAPE_LEFT`. I'm not sure if this is actually true, so it would be awesome if people with tablet devices could try this out. This definitely works for devices with a `PORTRAIT` natural orientation, however (most mobile phones, and some tablets), and in the worst-case is no more broken than functionality is currently.

Essentially, this PR accounts for screen orientation when calculating rotation values to pass to detector tasks, which naturally results in detector bounds coming back correct for current view coordinates.

With these changes, if the screen orientation is locked the bounds will not rotate as if the screen orientation were not locked. When the screen orientation isn't locked, this still works as it did before.

Tested on my Pixel 3 and Samsung Galaxy Tab A 10.

### DIFF
--- a/android/src/main/java/org/reactnative/camera/RNCameraView.java
+++ b/android/src/main/java/org/reactnative/camera/RNCameraView.java
@@ -2,6 +2,7 @@ package org.reactnative.camera;
 
 import android.Manifest;
 import android.annotation.SuppressLint;
+import android.content.Context;
 import android.content.pm.PackageManager;
 import android.content.res.Configuration;
 import android.content.res.Resources;
@@ -14,11 +15,14 @@ import android.os.Build;
 import androidx.core.content.ContextCompat;
 
 import android.util.DisplayMetrics;
+import android.util.Log;
 import android.view.GestureDetector;
 import android.view.MotionEvent;
 import android.view.ScaleGestureDetector;
 import android.view.View;
 import android.os.AsyncTask;
+import android.view.WindowManager;
+
 import com.facebook.react.bridge.*;
 import com.facebook.react.uimanager.ThemedReactContext;
 import com.google.android.cameraview.CameraView;
@@ -40,6 +44,8 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 
 public class RNCameraView extends CameraView implements LifecycleEventListener, BarCodeScannerAsyncTaskDelegate, FaceDetectorAsyncTaskDelegate,
     BarcodeDetectorAsyncTaskDelegate, TextRecognizerAsyncTaskDelegate, PictureSavedDelegate {
+  private static final String TAG = "RNCameraView";
+
   private ThemedReactContext mThemedReactContext;
   private Queue<Promise> mPictureTakenPromises = new ConcurrentLinkedQueue<>();
   private Map<Promise, ReadableMap> mPictureTakenOptions = new ConcurrentHashMap<>();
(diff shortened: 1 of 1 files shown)


android/src/main/java/org/reactnative/camera/RNCameraView.java
```diff
@@ -223,6 +260,10 @@ protected void onLayout(boolean changed, int left, int top, int right, int botto
     }
     float width = right - left;
     float height = bottom - top;
+
+    WindowManager windowManager = (WindowManager) mThemedReactContext.getSystemService(Context.WINDOW_SERVICE);
+    int rotation = windowManager.getDefaultDisplay().getRotation();
```
cristianoccazinsp, line 265: May be missing a null check here for `windowManager`

cristianoccazinsp:
@chiubaka do you really need all the orientation detection code? The camera views/implementation already use an existing class/helper to do so (https://github.com/react-native-camera/react-native-camera/blob/master/android/src/main/java/com/google/android/cameraview/DisplayOrientationDetector.java). 

Have you checked if there's a performance impact after the added detection on `onFramePreview` ?

Lastly, what about iOS, does it have the same issue?

chiubaka:
Re performance: I have this running in production on Android right now. No performance problems that I can tell so far. Also, the extra work is really only done in the constructor and `onLayout` or `onOrientationChange`, so these orientation calculations aren't computed on every frame. Any added work during the actual frame processing should be relatively negligible--just some simple arithmetic.

chiubaka:
Re iOS: I haven't looked at the iOS code closely, but the iOS implementations for this library don't seem to have any of the orientation / bounding box issues that the Android implementations do, so no need for a separate fix there.

chiubaka:
Re the existing `DisplayOrientationDetector`: yeah, I saw that. I think because `DisplayOrientationDetector` has no concept of "natural screen orientation", it isn't that useful to us here. We also don't need a live callback to get real-time orientation/rotation changes as much as we need to know what the values were the last time a screen orientation change was detected, which I'm doing in `onConfigurationChanged` (not the same as what `DisplayOrientationDetector` is detecting, unfortunately). Missing this logical distinction is likely what caused this error in the first place.

It might be possible to do a slight refactor to rely on a `DisplayOrientationDetector` instance for its `getLastKnownDisplayOrientation` method rather than calling `context.getResources().getConfiguration().orientation` in some places, but IMO that would actually make this more complicated, not less, and probably isn't necessary.

cristianoccazinsp:
Hey, great answers, I think it makes sense!

About DisplayOrientationDetector, it is already being used for as long as the camera is mounted (camera1 and camera2), so I thought it could be re used without adding further callbacks. But if you think there’s no point in trying to re use it, I guess it’s ok too.

chiubaka:
Ah, I don't have access to the existing `DisplayOrientationDetector` instance created by the Camera because I think it's private. So I'd actually have to create a new instance, which would be an extra physical callback even though it's not an extra logical one.

Like I mentioned though, I think we actually just need a snapshot of what the orientation values _were_ the last time the screen (i.e. UX) orientation changed, since that's what really matters to both the client JS code and the end user. That's not the same callback as the one provided by `DisplayOrientationDetector` as far as I can tell.

PR #3442, opened 2022-10-16
change imports for view props

azizmobarak:
since ViewPropTypes is deleted from react-native I have changed the import from `react-native` to `deprecated-react-native-prop-types`

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "logo": "https://opencollective.com/opencollective/logo.txt"
   },
   "dependencies": {
-    "prop-types": "^15.6.2"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -9,7 +9,8 @@
     "logo": "https://opencollective.com/opencollective/logo.txt"
   },
   "dependencies": {
-    "prop-types": "^15.6.2"
+    "deprecated-react-native-prop-types": "^2.3.0",
```
maxcwolf, line 12: The `yarn.lock` change needs to be committed with this since CI uses `yarn --frozen-lockfile`. It won't build otherwise.

PR #3443, opened 2022-10-17
Switch to deprecated-react-native-prop-types package

pbbadenhorst:
Switched to the specified package so that newer react-native projects can continue using this package

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -12,7 +12,8 @@
  * If you are seeing this from the future, please, send us your cutting-edge technology :) (if it exists)
  */
 import { Component, ReactNode } from 'react';
-import { NativeMethods, ViewProperties, findNodeHandle } from 'react-native';
+import { NativeMethods, findNodeHandle } from 'react-native';
+import { ViewProperties } from 'deprecated-react-native-prop-types';
 
 type Orientation = Readonly<{
   auto: any;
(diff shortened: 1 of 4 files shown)


types/index.d.ts
```diff
@@ -12,7 +12,8 @@
  * If you are seeing this from the future, please, send us your cutting-edge technology :) (if it exists)
  */
 import { Component, ReactNode } from 'react';
-import { NativeMethods, ViewProperties, findNodeHandle } from 'react-native';
+import { NativeMethods, findNodeHandle } from 'react-native';
+import { ViewProperties } from 'deprecated-react-native-prop-types';
```
azizmobarak: are you sure `ViewProperties` exist in `deprecated-react-native-prop-types`?

azizmobarak (commented):
also I see that is crashing on android after doing the update , I m not sure if it's related to the change, please check it from your side to confirm

tomruijgrok:
This PR would be very helpful for me

types/index.d.ts
pbbadenhorst: Sorry that was a mistake

shubhamdhage99 (approved):

danielpadu (approved):

azizmobarak (approved):

